PR #4, opened 2017-05-29
Scanning for transitive dependencies.

isoos:
Scanning transitive libs is slow at the moment, because parsing a dart file takes about 1 second on my macbook, and a package with large amount of files (e.g. angular) takes a lot of time.


isoos:
@kevmoo: I haven't checked it yet, but maybe using regex would be faster than the analyzer library, although certainly less precise (e.g. how to filter commented imports). Any preference on that?

kevmoo:
I'm not too worried about this being a bit slow. I can run through 140+ packages in a few minutes on my linux box. Given the weird stuff folks do in source, I'd rather avoid regexp.

We can through CPU at this in production...

isoos:
I've updated this PR with the clever parts of `graph_lib`, it is much faster indeed. However, end2end analysis of angular2 is still around 1.5 minutes.

I'm not sure if it is always worth to keep the transient lib references (it can get really long), added a flag for it.

I've deferred the platform classification to a subsequent PR.

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,185 @@
...
+  final String _package;
+  final String _projectPath;
+
+  AnalysisContext _context;
+  UriResolver _packageResolver;
+  Map<String, List<String>> _cachedLibs = {};
+
+  LibraryScanner(this._package, this._projectPath);
+
+  Future<Map<String, List<String>>> scanDirectLibs() {
```
kevmoo: could be `=>` method

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,185 @@
...
+
+import 'utils.dart';
+
+class LibraryScanner {
+  final String _package;
+  final String _projectPath;
+
+  AnalysisContext _context;
+  UriResolver _packageResolver;
+  Map<String, List<String>> _cachedLibs = {};
```
kevmoo: Should be final

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,185 @@
...
+          todo.addAll(await _scanUri(lib));
+        }
+      }
+
+      results[key] = processed.toList()..sort();
+    }
+    return toSortedMap(results);
+  }
+
+  Future _init() async {
```
kevmoo: Why return a `Future` here?

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  throw "not supported - $uri";
+}
+
+String _toUri(String package, String relativePath) {
+  if (relativePath.startsWith('lib/')) {
+    return 'package:$package/${relativePath.substring(4)}';
+  } else {
+    return 'path:$package/$relativePath';
```
kevmoo: Usually we just go w/ relative paths here: `bin/markdown.dart`

Since we know the context of non-package paths...right?

kevmoo: Then (maybe) put those in a third case.

If the import is just path relative to the lib dir, it should just be a path IMHO

isoos: I have one argument for the current way:
https://github.com/kevmoo/pana.dart/pull/4/files#diff-5037fb91fae4e421431eae8fd216c9cbR161

With the current impl, both the keys and the library values follow the same pattern, regardless of their context (e.g. whether this is the source package or a dependent one). This allows allows simple caching, and if we remove it, the caching logic needs to become more complex (e.g. need to consider/transform the values in a different context).

kevmoo: okay okay 😜

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,184 @@
...
+          todo.addAll(await _scanUri(lib));
+        }
+      }
+
+      results[key] = processed.toList()..sort();
+    }
+    return toSortedMap(results);
+  }
+
+  void _init() {
```
kevmoo: just roll this into the ctor, no?

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,184 @@
...
+    // TODO: fail more clearly if this...fails
+    var sdkPath = cli.getSdkDir().path;
+
+    JavaSystemIO.setProperty("com.google.dart.sdk", sdkPath);
+    var resourceProvider = PhysicalResourceProvider.INSTANCE;
+    DartSdk sdk = new FolderBasedDartSdk(
+        resourceProvider, resourceProvider.getFolder(sdkPath));
+
+    _packageResolver = _getPackageResolver(sdk);
```
kevmoo: points for making a private ctor + public factory and having all fields private - but not a huge deal

isoos:
Rebased, updated, PTAL.

lib/pana.dart
```diff
@@ -64,6 +66,18 @@ class PackageAnalyzer {
         upgrade.exitCode, upgrade.stdout, upgrade.stderr, pkgDir);
     log.info("Package version: ${summary.pkgVersion}");
 
+    LibraryScanner libraryScanner = new LibraryScanner(package, pkgDir);
+    Map<String, List<String>> directLibs;
+    Map<String, List<String>> transitiveLibs;
+    try {
+      directLibs = await libraryScanner.scanDirectLibs();
+      transitiveLibs = await libraryScanner.scanTransitiveLibs();
```
kevmoo: Is one more likely to fail than the other? Any reason why these are not ran separately?

lib/src/library_analyzer.dart
```diff
@@ -0,0 +1,177 @@
...
+    }
+    return results;
+  }
+
+  List<String> _parseLibs(
+      String package, String packageDir, String relativePath) {
+    String fullPath = p.join(packageDir, relativePath);
+    LibraryElement lib = _getLibraryElement(fullPath);
+    if (lib == null) return [];
+    Set<String> refs = new Set();
```
kevmoo: If you use a SplayTreeSet here, you don't have to sort after toList

isoos:
fixed

kevmoo:
Once travis is green, land it! 👍

PR #6, opened 2017-05-31
Add scoring logic inspired by npms.io work

kevmoo:
CC @isoos

### DIFF
--- a/lib/src/scoring.dart
+++ b/lib/src/scoring.dart
@@ -1,0 +1,190 @@
+// Copyright (c) 2017, Kevin Moore. All rights reserved. Use of this source code
+// is governed by a BSD-style license that can be found in the LICENSE file.
+
+import 'dart:math' as math;
+
+class Summary {
+  final num max;
+  final num min;
+  final num mean;
+
+  Summary._(this.max, this.min, this.mean) {
+    assert(max >= mean);
+    assert(mean >= min);
+  }
+
+  factory Summary(Iterable<num> numbers) {
+    var total = 0.0;
+    num max, min;
+
+    var count = 0;
+    for (var i in numbers) {
+      //TODO(kevmoo): validate i for Nan/Infinity
+      count++;
+      total += i;
+
+      if (max == null || i > max) {
+        max = i;
+      }
+
+      if (min == null || i < min) {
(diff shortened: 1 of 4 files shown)


lib/src/scoring.dart
```diff
@@ -0,0 +1,192 @@
...
+    var total = 0.0;
+    num max, min;
+
+    var count = 0;
+    for (var i in numbers) {
+      //TODO(kevmoo): validate i for Nan/Infinity
+      count++;
+      total += i;
+
+      if (max == null || i > max) {
```
isoos: import 'dart:math' show min, max;
num maxValue = numbers.fold(numbers.first, max);
num minValue = numbers.fold(numbers.first, min);
(or use the prefixed lib if you keep the min/max field names)

lib/src/scoring.dart
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+
+  // Convert back from depressed cubic
+  for (var i = 0; i < roots.length; i++) roots[i] -= b / (3 * a);
+
+  return roots;
+}
+
+num abs(a) => a < 0 ? -a : a;
```
isoos: why not a.abs() ?
https://api.dartlang.org/stable/1.23.0/dart-core/num/abs.html

lib/src/scoring.dart
kevmoo: then we iterate twice – and then a third time to count the items. I designed it to allow a streamed input.

lib/src/scoring.dart
kevmoo: Oh...yeah... 😊

lib/src/scoring.dart
isoos: right...
the if can be spared though with:
max = math.max(i, max ?? i);

but not that important

kevmoo: nice!

kevmoo:
@isoos PTAL

isoos:
This got mixed commits with the "Use dart files to store the expected output". I'm actually fine with that too, so merging...

kevmoo:
In the future, let me do the merge after you approve. I like clean history.
:-)

PR #7, opened 2017-06-01
Include file sizes

kevmoo:


isoos:
At one point it'll be worth to merge all the per-file analysis into a single LibInfo (?) object, including:
- file size
- dartfmt result
- analyzer issues
- lib dependencies
- platform classification

shall we defer this PR for that?

kevmoo:
> shall we defer this PR for that?

Uh...well, I guess...

PR #8, opened 2017-06-01
Platform detection basics.

isoos:
I've deferred to make it part of the summary, in case you decide that we shall merge the library-level analysis.

### DIFF
--- a/lib/src/platform.dart
+++ b/lib/src/platform.dart
@@ -1,0 +1,115 @@
+import 'pubspec.dart';
+
+abstract class KnownPlatforms {
+  static const String browser = 'browser';
+  static const String console = 'console';
+  static const String flutter = 'flutter';
+  static const String mirrors = 'mirrors';
+
+  static const String angular = 'angular';
+}
+
+class PlatformSummary {
+  final Platform package;
+  final Map<String, Platform> libraries;
+
+  PlatformSummary(this.package, this.libraries);
+
+  bool get hasConflict =>
+      package.hasConflict ||
+      libraries.values.any((p) => p.hasConflict) ||
+      _conflictsFlutter;
+
+  bool get _conflictsFlutter =>
+      package.uses.contains(KnownPlatforms.flutter) &&
+      libraries.values.any((p) => !p.worksInFlutter);
+}
+
+class Platform {
+  final List<String> uses;
+
+  Platform._(this.uses);
+
+  factory Platform(Iterable<String> uses) =>
+      new Platform._((uses ?? []).toList()..sort());
+
+  factory Platform.fromMap(Map<String, dynamic> map) =>
+      new Platform(map['uses']);
+
+  bool get hasConflict =>
+      (!worksAnywhere) ||
+      (uses.contains(KnownPlatforms.flutter) && !worksInFlutter);
+
+  bool get worksEverywhere =>
+      worksInBrowser && worksInConsole && worksInFlutter;
+
+  bool get worksAnywhere => worksInBrowser || worksInConsole || worksInFlutter;
+
+  bool get worksInBrowser =>
+      _hasNoUseOf([KnownPlatforms.flutter]) &&
+      (uses.contains(KnownPlatforms.browser) ||
+          _hasNoUseOf([KnownPlatforms.console]));
+
+  bool get worksInConsole =>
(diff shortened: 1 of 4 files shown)


test/platform_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      expect(p.excludes, [KnownPlatforms.browser]);
+    });
+
+    test('dart:html', () {
+      Platform p = classifyPlatform(['dart:html']);
+      expect(p.isEmpty, isFalse);
+      expect(p.hasConflict, isFalse);
+      expect(p.only, [KnownPlatforms.browser]);
+      expect(p.uses, []);
+      expect(p.excludes, [KnownPlatforms.console]);
```
kevmoo: and flutter?

test/platform_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+      p = classifyPlatform(['dart:svg']);
+      expect(p.isEmpty, isFalse);
+      expect(p.hasConflict, isFalse);
+      expect(p.only, [KnownPlatforms.browser]);
+      expect(p.uses, []);
+      expect(p.excludes, [KnownPlatforms.console]);
```
kevmoo: and flutter

lib/src/platform.dart
```diff
@@ -0,0 +1,129 @@
...
+
+  bool get hasConflict =>
+      package.hasConflict ||
+      libraries.values.any((p) => p.hasConflict) ||
+      libraries.values.any((p) => p.excludes
+          .any((s) => package.only.contains(s) || package.uses.contains(s)));
+}
+
+class Platform {
+  final List<String> only;
```
kevmoo: Wondering if these should be `Set<String>`

lib/src/platform.dart
```diff
@@ -0,0 +1,129 @@
...
+  // browser
+  if (usesWeb && !usesIo) {
+    only.add(KnownPlatforms.browser);
+    excludes.add(KnownPlatforms.console);
+  }
+
+  // http and other cross-platform library
+  if (usesWeb && usesIo) {
+    uses.add(KnownPlatforms.browser);
+    uses.add(KnownPlatforms.console);
```
kevmoo: should exclude `console`. web code can import `dart:io`, but not the other way around

lib/src/platform.dart
kevmoo: I'd love to understand your logic here.

using `dart:html` should exclude flutter (like using `dart:mirrors` excludes flutter`).

I guess I'd expect just a `uses` field – and then a function that can answer "will this work on platform X" with the logic centralized.

isoos: during the analysis it is a `Set`, but after the results are done, it is more convenient to use `List` for storing in JSON, validating it in a test or displaying it in predictable order.

isoos: well, there was no such rule in the platform signals doc, and I was leaning on a more optimistic rule :)

lib/src/platform.dart
isoos: but then `http` package excludes console, which is not true.

lib/src/platform.dart
kevmoo: You could use `SplayTreeSet` from `dart:collection` for getting a stable order.

Not a biggy...

kevmoo: I could have been more clear in the doc. I have "web only" – which excludes flutter in the end.

My thought: just populate "uses" in the serialized data model. Then the effect of a set of uses becomes purely procedural – can evolve over time, etc.

lib/src/platform.dart
kevmoo: the http package is both.

It has one library that is web-only and one library that is io-only.

I thought we were rolling this data up at the library level – then the package can have a mix.

lib/src/platform.dart
isoos: Yeah, but then `Platform()` gets slightly different. As the number of items in these sets < 2, I doubt it will matter anyway.

isoos: Updated. PTAL.
I don't really get the part with the "uses" and it being purely procedural, please explain.

kevmoo: true true

isoos:
Rebased, and updated as discussed on HO. PTAL.

kevmoo:
Are you going to rebase this on Master? It shares code, no?

PR #13, opened 2017-06-02
Move pubspec-parsing utils to pubspec.dart.

isoos:


kevmoo:
Nice! – need a reformat to make travis happy...

lib/src/pubspec.dart
```diff
@@ -1,21 +1,36 @@
...
       hasFlutterKey &&
       _data['flutter'] is Map &&
       _data['flutter']['plugin'] != null;
 
-  bool get dependsOnFlutterSdk {
-    if (dependencies == null) return false;
-    return dependencies.values
-        .any((value) => value is Map && value['sdk'] == 'flutter');
-  }
+  bool get dependsOnFlutterSdk =>
```
kevmoo: Couldn't this just be `dependentSdks.contains('flutter')`?

This we can get rid of the `_dependsOnFlutterSdk` helper? Seems a bit redundant...

isoos: good catch, I meant to do that.

PR #14, opened 2017-06-02
Refactor summary: per-file reports.

isoos:


isoos:
> We don't need the platform info for lib/src things – just makes the output files HUGE.

I think that is how it works right now: we only report the `bin/` and `lib/`, without the `lib/src/` directories. Do you have an example where it doesn't work like that?

kevmoo:
Ah...sorry. Just double checked. Looks good! Land it!

PR #22, opened 2017-06-09
Collect tooling/config/runtime issues, instead of throwing on the first one that breaks.

### LINKED ISSUE #19: fail gracefully when dartanalyzer's output is not in the expected format
for example run `dart bin/main.dart glob`

isoos:
Fixes #19.


kevmoo:
Love it!

I'm going to pull out the crazy from `PubSummary` and use this instead...

PR #24, opened 2017-06-09
Move Flutter SDK-related calls to FlutterSdk.

isoos:


lib/src/sdk_env.dart
```diff
@@ -69,13 +62,49 @@ class DartSdk {
   }
 }
 
+class FlutterSdk {
+  String _flutterBin;
```
kevmoo: make this final.

lib/src/sdk_env.dart
```diff
@@ -69,13 +62,49 @@ class DartSdk {
...
+
+  FlutterSdk({String sdkDir}) {
+    String path = _prepareSdkPath(sdkDir);
+    _flutterBin = '${path}flutter';
```
kevmoo: Why not p.join w/ the path package?

I avoid path concat w/ strings like the plague

lib/src/sdk_env.dart
```diff
@@ -214,3 +235,14 @@ final _versionDownloadRexexp =
     new RegExp(r"^MSG : (?:Downloading |Already cached )([\w-]+) (.+)$");
 
 const _pubEnvironmentKey = 'PUB_ENVIRONMENT';
+
+String _prepareSdkPath(String sdkDir) {
+  String path = '';
+  if (sdkDir != null) {
+    path = sdkDir;
+    if (!sdkDir.endsWith(Platform.pathSeparator)) {
+      path += Platform.pathSeparator;
```
kevmoo: See comment above WRT pkg/path – avoid string manipulation of paths...

lib/src/sdk_env.dart
```diff
@@ -114,29 +143,14 @@ class PubEnvironment {
...
-      result = await runProc(
-        cmd,
-        args,
-        workingDirectory: packageDir,
-        environment: _environment,
-      );
+      if (isFlutter) {
+        result = await flutterSdk._execUpgrade(packageDir, _environment);
+      } else {
+        result = await _execUpgrade(packageDir);
```
kevmoo: wondering if this _execUpgrade should be moved to the `DartSdk` class?

isoos: good catch!

kevmoo: And then should we have an abstract base class that ...blah blah blah...

isoos: I don't think that would make it cleaner.

kevmoo: Just being an astronaut – no worries.

kevmoo:
Ship it!

PR #26, opened 2017-06-12
Create PlatformSummary from Summary. Fixes #25.

### LINKED ISSUE #25: Rationalize Platform + PlatformSummary w/ Summary
I'd like to take the output from bin/main.dart and/or bin/runner.dart and throw that into a Platform/PlatformSummary

isoos:


lib/src/summary.dart
```diff
@@ -126,6 +127,15 @@ class Summary {
 
     return map;
   }
+
+  PlatformSummary getPlatformSummary() {
+    Map<String, Platform> libraries = {};
```
kevmoo: new Map.fromInterable would be greate here, instead.

lib/src/summary.dart
```diff
@@ -126,6 +127,15 @@ class Summary {
...
+    dartFiles.values.forEach((dfs) {
+      libraries[dfs.uri] = dfs.platform;
```
kevmoo: Should we filter to be only non-src files?

I'm crashing because I'm hitting src files from analysis that have a `null` `platform` value.

isoos: good point. I've updated to filter on `platform != null`, which atm. is the same as checking for only non-src files. However it is only a temp measure, instead, we shall introduce a flag on DartFileSummary, which should also track non-trivial `src` paths like `lib/something/src/internal.dart`.

kevmoo:
Land it!

PR #28, opened 2017-06-12
Remove redundant PubSummary.lockedVersion

kevmoo:
But keep the logic around to validate the pub output parsing


kevmoo:
> Can you confirm that if there is an assert issue triggered, it will end up in AnalyzerIssue?

Done. PTAL at the new code...

lib/pana.dart
```diff
@@ -79,8 +79,18 @@ class PackageAnalyzer {
 
     PubSummary summary;
     if (upgrade.exitCode == 0) {
-      summary = PubSummary.create(upgrade.stdout, path: pkgDir);
-      log.info("Package version: ${summary.pkgVersion}");
+      try {
+        summary = PubSummary.create(upgrade.stdout, path: pkgDir);
+      } catch (e, stack) {
+        log.severe("Problem summarizing package", e, stack);
+        //(TODO)kevmoo - should add a helper that handles logging exceptions
```
kevmoo: @isoos sound good to you? Just add an inline function?

lib/src/pub_summary.dart
```diff
@@ -13,15 +13,13 @@ import 'utils.dart';
 
 class PubSummary {
   static final _solvePkgLine = new RegExp(
-      r"(?:[><\+ ]) (\w+) (\S+)(?: \((\S+) available\))?(?: from .+)?");
+      r"(?:[><\+\! ]) (\w+) (\S+)(?: \((\S+) available\))?(?: from .+)?");
```
kevmoo: Oops! Didn't handle the override case. Doing the validation actually helped me fix a bug

lib/pana.dart
isoos: I don't mind writing two lines of code in such cases, but I don't oppose the helper either.

isoos:
> Done. PTAL at the new code...

LGTM

PR #29, opened 2017-06-13
Cleanup

kevmoo:
The runner changes are a bit nuts. Don't worry about them too much.

Changes to the core libraries, though, I want you to feel good about them.

### DIFF
--- a/bin/main.dart
+++ b/bin/main.dart
@@ -14,26 +14,52 @@ final _none = '\u001b[0m';
 String gray(text) => "$_gray$text$_none";
 
 main(List<String> arguments) async {
+  // as provided, `arguments` is fixed length – turn it into a mutable `List`
+  arguments = arguments.toList();
+
+  var json = false;
+  if (arguments.contains('--json')) {
+    json = true;
+    arguments.removeWhere((a) => a == '--json');
+  }
+
   log.Logger.root.level = log.Level.ALL;
-  log.Logger.root.onRecord.listen((record) {
-    var wroteHeader = false;
 
-    var msg = LineSplitter
-        .split([record.message, record.error, record.stackTrace]
-            .where((e) => e != null)
-            .join('\n'))
-        .map((l) {
-      String prefix;
-      if (wroteHeader) {
-        prefix = '';
-      } else {
-        wroteHeader = true;
-        prefix = record.level.toString();
+  if (json) {
+    log.Logger.root.onRecord.listen((log) {
+      var map = <String, Object>{};
+
+      if (log.loggerName.isNotEmpty) {
+        map['logName'] = log.loggerName;
       }
-      return "${prefix.padRight(10)} ${l}";
-    }).join('\n');
 
-    stderr.writeln(gray(msg));
+      map.addAll({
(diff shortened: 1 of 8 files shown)


bin/main.dart
```diff
@@ -14,26 +14,52 @@ final _none = '\u001b[0m';
 String gray(text) => "$_gray$text$_none";
 
 main(List<String> arguments) async {
+  // as provided, `arguments` is fixed length – turn it into a mutable `List`
```
isoos: why not use the package:args?

bin/main.dart
```diff
@@ -14,26 +14,52 @@ final _none = '\u001b[0m';
...
+
+  // Docker is WEIRD
+  // The SIGTERM signal sent to `docker run...` DOES propagate a signal to the
+  // running process. But...
+  //   * It is received as SIGINT
+  //   * It won't terminal the Dart process either – *BUT* we can listen for it
+  // So this is how we do "clean" shutdown when running in Docker.
+  var subscription = getSignals().listen((sig) async {
+    log.Logger.root.severe("Received signal `$sig` – terminating.");
+    exit(130);
```
isoos: out of curiosity: why 130? :)

bin/main.dart
```diff
@@ -56,12 +82,33 @@ main(List<String> arguments) async {
...
       exitCode = 1;
     }
   } finally {
     tempDir.deleteSync(recursive: true);
   }
+
+  await subscription.cancel();
+}
+
+void _logWriter(record) {
```
isoos: nit: always wondered if there is a package for this? let's keep it for now, but I'd really like to know what others are using (and replace my similar hacks with that).

lib/src/pub_summary.dart
```diff
@@ -134,27 +134,45 @@ class PubSummary {
...
       if (versionConstraint == null) {
-        stderr.writeln('No constraint provided for $package');
+        logWeird('BAD! No constraint provided for $package');
         return;
       }
 
       if (versionConstraint is Map) {
-        stderr.writeln(
-            'Non-versioned constraint for $package\n  $versionConstraint');
+        if (versionConstraint['sdk'] == 'flutter') {
```
isoos: is this a duplicate with `Pubspec.dependsOnFlutterSdk` (in `pubspec.dart`)?

bin/main.dart
kevmoo: 128 + 2 = ~convention for user killed.

http://tldp.org/LDP/abs/html/exitcodes.html

bin/main.dart
kevmoo: Agreed!

bin/main.dart
kevmoo: I'm lazy! 😄 

I have it in there as a TODO - somewhere

lib/src/pub_summary.dart
kevmoo: I'll look and see if I can fix

kevmoo: Ah! No. This is detecting if a package depends on flutter – not the whole Pubpsec/package.

PR #30, opened 2017-06-14
More updates

kevmoo:


lib/src/library_analyzer.dart
```diff
@@ -115,8 +115,14 @@ class LibraryScanner {
 
   Future<List<String>> _scanUri(String libUri) async {
     Uri uri = Uri.parse(libUri);
-    String package = uri.path.split('/').first;
-    String fullPath = _packageResolver.resolveAbsolute(uri).fullName;
+    String package = uri.pathSegments.first;
+
+    var source = _packageResolver.resolveAbsolute(uri);
```
kevmoo: Was throwing w/ an NPE on `firebase_database` – at least we get a more helpful log now

Dockerfile
```diff
@@ -10,7 +10,11 @@ ENV PATH="/flutter/bin:${PATH}"
 
 # Running `flutter config --no-analytics` downloads the Dart SDK and
 # disables analytics tracking – which we always want
+# `1a1bbacf0d` maps to alpha release 0.0.8 @ June 13, 2017
+#    We hard wire the SHA to ensure Docker rebuilds
```
isoos: what is the difference if we were to use tags instead?

kevmoo: There is a change the tag could be moved – but Docker wouldn't rebuild because the `RUN` content didn't change.

By hardwiring the SHA, we are ensured that the underlying state is not changing – unless we change it.

isoos: so... I guess we should apply the same for pub as well

isoos:
otherwise LGTM

Dockerfile
kevmoo: I think so.

The case I'm worried about: we pin to some  tag and then someone decides to update it – we'll have to fall back to this logic.

Having it explicit makes it very clear where we're pointed – there's no notion of "0.0.8" as of June 25

PR #31, opened 2017-06-14
Use source_gen for JSON serialization

kevmoo:
...for a few classes

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -18,7 +18,6 @@ linter:
      - valid_regexps
 
      # Style
-     - annotate_overrides
      - avoid_init_to_null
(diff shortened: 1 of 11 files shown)


kevmoo:
We'll need https://github.com/dart-lang/source_gen/pull/150 to support `Version`

I'll also want to look at support for excluding the key for `null` values.

isoos:
> We'll need dart-lang/source_gen#150 to support Version

Do we need to support versions? I'm fine using String for versions, and lazily parse when needed.

kevmoo:
> Do we need to support versions? I'm fine using String for versions, and lazily parse when needed.

Strictly, no. But it's so much SEXIER to store the value we want to keep. 😁

PR #33, opened 2017-06-19
handle errors while checking dartfmt

kevmoo:


lib/pana.dart
```diff
@@ -173,7 +188,7 @@ class PackageAnalyzer {
       files[dartFile] = new DartFileSummary(
         uri,
         size,
-        !unformattedFiles.contains(dartFile),
+        unformattedFiles == null ? null : !unformattedFiles.contains(dartFile),
```
isoos: I'm a bit unconvinced on the use of null here. Not that I know any better, but I wish there would be a better way of indicating the error.

Some options:
- if `dartfmt` fails miserably, let's assume every file is unformatted ("strong mode" of formatting)
- if `dartfmt` fails for the entire package, let's call it separately for each file

isoos: ... or just make a comment for `isFormatted` field that it can be null.

kevmoo: Done...

PR #38, opened 2017-07-04
Export internals that we want to use in caller libraries.

isoos:


kevmoo:
@isoos – check your travis – missing a couple of things.

Well, actually things are hidden

isoos:
Renamed `Platform` -> `PlatformInfo`, because I think its easier to handle the rename now than always taking care of the conflict with `dart:io`.

PR #39, opened 2017-07-04
Detecting some licenses.

isoos:
I think the [licensee](https://github.com/benbalter/licensee) or [detect-license](https://github.com/tantalor/detect-license/) should be used only as a fallback when we fail to detect it. We shall try to get away with trivial patterns, because:
- it is simple to do so for most use cases
- faster to execute than an external process with is own runtime baggage
- doesn't require additional setup for local development.

This CL is for the very simple patterns, we can add more as we go ahead with the package analysis, and eventually some third-party if we have many files that are not classified properly.

### DIFF
--- a/lib/pana.dart
+++ b/lib/pana.dart
@@ -11,6 +11,7 @@ import 'package:pub_semver/pub_semver.dart';
 
 import 'src/analyzer_output.dart';
 import 'src/library_analyzer.dart';
+import 'src/license.dart';
 import 'src/logging.dart';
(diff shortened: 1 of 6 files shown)


kevmoo:
Can has resolve?

isoos:
> Can has resolve?

Ah, missed the merge conflict, rebased.

kevmoo:
@isoos you might want to rebase again – just fixed the tests

PR #41, opened 2017-07-06
Filter not only lib/src, but all of the /src/ directories inside lib/

isoos:


kevmoo:
Explain?

https://www.dartlang.org/tools/pub/package-layout

In general, we say we only support `lib/src` as non-public sources.

isoos:
I vaguely remembered that some project has a non-lib-level `/src/` dir, I had `angular_components` in mind, but haven't checked. It actually does: https://github.com/dart-lang/angular_components/tree/master/lib/src/laminate/popup/src but it is inside a lib-level `lib/src`, so I guess it is good.

PR #42, opened 2017-07-12
Migrate `uses` values to store dart:* references directly

kevmoo:
Makes the values stored in JSON output much less opaque

### DIFF
--- a/lib/src/platform.dart
+++ b/lib/src/platform.dart
@@ -14,7 +14,7 @@ abstract class PlatformFlags {
   /// Denotes a library that depends on a native extensions via `dart-ext:`
   static const String dartExtension = 'dart-ext';
 
-  /// Denotes a library that depends on pkg/angular or pkg/angular2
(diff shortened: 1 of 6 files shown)


lib/src/platform.dart
```diff
@@ -1,20 +1,20 @@
...
   static const String flutter = 'flutter';
-  static const String mirrors = 'mirrors';
 
-  /// Native extensions (Dart VM with C/C++ code).
-  static const String native = 'native';
+  /// Denotes a library that depends on a native extensions via `dart-ext:`
+  static const String dartExtension = 'dart-ext';
 
+  /// Denotes a library that depends on `pkg/angular` or `pkg/angular2`.
   static const String angular = 'angular';
```
isoos: Nit: as we are storing `dart:*` in the list, maybe we shall use `package:[blah]` here too?

kevmoo: Just going to remove the concept of `angular` entirely. No plan to use it.

PR #43, opened 2017-07-13
Graph and refactor

kevmoo:
Builds off https://github.com/dart-lang/pana/pull/42

### DIFF
--- a/lib/src/library_analyzer.dart
+++ b/lib/src/library_analyzer.dart
@@ -128,6 +128,30 @@ class LibraryScanner {
     AnalysisEngine.instance.clearCaches();
   }
 
+  Future<Map<String, List<String>>> scanDependencyGraph() async {
+    var items = await scanTransitiveLibs();
+
+    var graph = new SplayTreeMap<String, List<String>>();
+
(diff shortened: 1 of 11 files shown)


lib/src/library_analyzer.dart
```diff
@@ -54,8 +54,24 @@ class LibraryScanner {
     var pubPackageMapProvider = new PubPackageMapProvider(
         PhysicalResourceProvider.INSTANCE, sdk, runPubList);
     var packageMapInfo = pubPackageMapProvider.computePackageMap(
-        PhysicalResourceProvider.INSTANCE.getResource(projectPath) as Folder);
+        PhysicalResourceProvider.INSTANCE.getResource(packagePath) as Folder);
     var packageMap = packageMapInfo.packageMap;
+
+    var packageNames = <String>[];
+    packageMap.forEach((k, v) {
+      if (v.any((f) => p.isWithin(packagePath, f.path))) {
```
isoos: This to prevent a wild import of '../../../../../etc/passwd'? Maybe we should flag it in the package analysis / general health score?

kevmoo: Uh...sure. Wanna open an issue for that?

PR #50, opened 2017-07-25
Overrides

kevmoo:


kevmoo:
@isoos – I landed this, but would still like your comments...

lib/pana.dart
```diff
@@ -142,7 +142,14 @@ class PackageAnalyzer {
 
     if (summary != null) {
       try {
-        libraryScanner = new LibraryScanner(pkgDir, isFlutter);
+        var overrides = {
```
isoos: why are these overrides?

kevmoo: They import dart:io – but in a way that's safe for the web.

isoos: We can already detect that in a generic way, not specific to any given package. Do we need something over that?

PR #51, opened 2017-07-26
Add version info to summary, etc

kevmoo:


dart_test.yaml
```diff
@@ -0,0 +1,3 @@
+tags:
```
isoos: Is this a config file that some Dart tooling reads?

kevmoo: pkg/test

kevmoo:
> My only question is the config file, otherwise lgtm

What's your question?

isoos:
>> My only question is the config file, otherwise lgtm
> What's your question?

dart_test.yaml, but you have answered that.

PR #52, opened 2017-07-27
Disable gray log on Mac, because it is invisible in terminal.

isoos:


bin/main.dart
```diff
@@ -11,7 +11,7 @@ import 'package:pana/pana.dart';
 final _gray = '\u001b[1;30m';
 final _none = '\u001b[0m';
 
-String gray(text) => "$_gray$text$_none";
```
kevmoo: I'm on mac and think it's fine!

bin/main.dart
```diff
@@ -11,7 +11,7 @@ import 'package:pana/pana.dart';
 final _gray = '\u001b[1;30m';
```
kevmoo: Change `30m` to `37m` – see how it looks?

kevmoo: Per http://tldp.org/HOWTO/Bash-Prompt-HOWTO/x329.html

bin/main.dart
isoos: I'm using `Terminal.app` in `Pro` mode with semi-transparent background, in which `30m` is not visible, `37m` is better. I guess you are using it in either totally black or white background?

kevmoo: Totally black

kevmoo: Or just add a flag for 'no color' or something?

isoos: At that point could just use `--json`, but it is harder to read :-)
I'm fine overriding this locally, although a bit inconvenient.

kevmoo: Just add a flag, right?

Or read an environment var?

I like having another color – it lets me distinguish log output from "other" print/stdout/stderr

PR #53, opened 2017-07-27
Skip dartanalyzer if there are no dart files in the package. Fixes #48.

### LINKED ISSUE #48: Handle packages without dart files better.
`dartfmt` log duplicates the chatter about files not being found. For example:
- `bootstrap_sass 3.3.6`

isoos:


lib/pana.dart
```diff
@@ -179,7 +179,9 @@ class PackageAnalyzer {
       }
 
       try {
-        analyzerItems = await _pkgAnalyze(pkgDir);
+        if (dartFiles.isNotEmpty) {
```
kevmoo: Put the try block in the if statement?

PR #56, opened 2017-08-01
Move flutter's list-package-dirs to the proper environment.

isoos:


lib/src/library_scanner.dart
```diff
@@ -48,9 +48,11 @@ class LibraryScanner {
           'expectetd location.\n$dotPackagesPath');
     }
 
+    // TODO: figure out why non-flutter pub list doesn't work the same way as the default
```
kevmoo: The pub client that's part of the flutter SDK "understands" flutter – it does the right thing when it sees a flutter dependency, for example

isoos: The issue at hand I have was with `PubPackageMapProvider._runPubListDefault`. It looks like it does the same thing I was doing in the non-flutter path of the new list code, but it behaved differently, and I'm not sure why.

PR #65, opened 2017-08-23
Override analysis_options.yaml to make results consistent across packages.

isoos:


isoos:
Any idea how to fox travis tests where dart 1.25 fails and 1.24.2 passes? (1.25 reports one more linter error).

kevmoo:
This is actually a very helpful error – clearly the analyzer is running w/ pana's analysis_options and not the target package.

pkg/http doesn't have the 'prefer single quotes' lint enabled!

lib/src/analysis_options.dart
```diff
@@ -0,0 +1,110 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+const String analysisOptions = '''
+analyzer:
+  strong-mode: true
+
+# Source of linter options:
+# http://dart-lang.github.io/linter/lints/options/options.html
+
+linter:
```
kevmoo: Nuke all of these lints. We shouldn't enforce lints on packages – unless we're sure they should be universal

kevmoo:
> This is actually a very helpful error – clearly the analyzer is running w/ pana's analysis_options and not the target package.

Oopps...saw that you actually have that lint in your file. See my note. We shouldn't have lints in our analysis_options

lib/src/analysis_options.dart
isoos: I think some of it should be universal, like:
- `annotate_overrides`: being able detect when the implemented interface changes in an unexpected way
- `hash_and_equals`
- `valid_regexps`

The others could be still part of our health score, and therefore present here. Which ones exactly: that is up for debate, but I'd definitely include these:
-  `implementation_imports`: because the package has a known risk factor
- `unawaited_futures`: may be an oversight or an error
- naming and style conventions (like `camel_case_types`), because these help users to build projects with the same universal style

isoos:
> Oopps...saw that you actually have that lint in your file. See my note. We shouldn't have lints in our analysis_options

Yeah, it is in my lint file, but that doesn't explain the error with the new version. Any idea there?

kevmoo:
> Yeah, it is in my lint file, but that doesn't explain the error with the new version. Any idea there?

The single quote lint doesn't exist before some version of 1.25.0-dev – hence inconsistency

isoos:
I've reduced the lint rules, but had not nuked it completely. I believe we shall keep track of some class of linter rules that we consider more important and being part (although it may be a small part) of the health score.

lib/src/analysis_options.dart
```diff
@@ -0,0 +1,25 @@
...
+  rules:
+    - annotate_overrides
+    - hash_and_equals
+    - valid_regexps
```
kevmoo: Annotate overrides is still not something we do consistently in our own code. I'd just keep these two...

isoos: ack

PR #67, opened 2017-08-25
Re-run codegen

isoos:


kevmoo:
You need to pub upgrade – you don't have latest json_serializable

PR #71, opened 2017-08-25
Fixes for library scanner and description

kevmoo:


lib/src/library_scanner.dart
```diff
@@ -66,19 +67,35 @@ class LibraryScanner {
           'for the file at `$dotPackagesPath`.');
     }
 
+    String package;
```
isoos: @kevmoo: what is this doing? I'm not sure I understand what your goal is with it...

kevmoo: This code figures out the name of the package by looking at the `.packages` file to see which directory exists under the current directory.

The logic before messed up if there were multiple `pubspec.yaml` files somewhere in the root.

PR #72, opened 2017-08-28
Mask only the dart:io library in overrides. Fixes #58.

### LINKED ISSUE #58: Let's remove LibraryScanner overrides
(1) The override in its current form is misleading, because it creates a `dart:web_safe_io` package which doesn't exists, but somebody may think it does. A `dart-trusted:http` would be a more adept name.

(2) I think having these overrides is a bad precedent, for the following reasons:
- We don't want to keep track all of the libraries that use `dart:io` in a web-safe way. Getting onto that list, changing it would require efforts on all sides.
- We may track a few, but forget about it. As the libraries, change, some of these information gets outdated, even in a breaking way, and provides misleading results.
- Tracking `package_resolver`'s `http` use is an ever worse precedent: are we going to list everything there that has `http` as its dependency?

The correct solution: analyze the Dart sources and detect, whether the client is going to use any of the non-web-related `dart:io` code, and mark it as such. This a lot of effort, and probably not worth it at the moment.

The optimistic solution: let's assume that when both `dart:html` and `dart:io` is present, it is importing it in a web-safe-io way. We already have tests for this: https://github.com/dart-lang/pana/blob/master/test/platform_test.dart#L83

I'd go with optimistic solution, and remove the overrides altogether.

isoos:
I've also used `dart-pana:` prefix to prevent confusion and make it clear that the package does not exists.


lib/src/library_scanner.dart
```diff
@@ -239,10 +244,26 @@ class LibraryScanner {
...
     });
+
+    var pkgUri = toPackageUri(package, relativePath);
+    _applyOverrides(pkgUri, refs);
+
     refs.remove('dart:core');
     return new List<String>.unmodifiable(refs);
   }
 
+  _applyOverrides(String pkgUri, Set<String> set) {
```
kevmoo: void?

lib/src/library_scanner.dart
```diff
@@ -22,20 +22,31 @@ import 'logging.dart';
 import 'sdk_env.dart';
 import 'utils.dart';
 
+class LibraryOverride {
```
kevmoo: Nice!

lib/pana.dart
```diff
@@ -143,11 +143,12 @@ class PackageAnalyzer {
 
     if (summary != null) {
       try {
-        var overrides = {
```
kevmoo: Yup

kevmoo:
@isoos I reverted this in 33944c87 – we need to propagate down the `pana-dart:` libraries to platform so we can determine if they still work on the web.

isoos:
> we need to propagate down the pana-dart: libraries to platform so we can determine if they still work on the web.

What is the error-case?

kevmoo:
`dart bin/main.dart test`

```json
"lib/test.dart": {
   "uri": "package:test/test.dart",
   "size": 12384,
   "isFormatted": true,
   "analyzerItems": [],
   "directLibs": [
    "dart:async",
    "package:matcher/matcher.dart",
    "package:path/path.dart",
    "package:test/src/backend/declarer.dart",
    "package:test/src/backend/invoker.dart",
    "package:test/src/backend/test_platform.dart",
    "package:test/src/frontend/expect.dart",
    "package:test/src/frontend/expect_async.dart",
    "package:test/src/frontend/future_matchers.dart",
    "package:test/src/frontend/on_platform.dart",
    "package:test/src/frontend/prints_matcher.dart",
    "package:test/src/frontend/skip.dart",
    "package:test/src/frontend/spawn_hybrid.dart",
    "package:test/src/frontend/stream_matcher.dart",
    "package:test/src/frontend/stream_matchers.dart",
    "package:test/src/frontend/tags.dart",
    "package:test/src/frontend/test_on.dart",
    "package:test/src/frontend/throws_matcher.dart",
    "package:test/src/frontend/throws_matchers.dart",
    "package:test/src/frontend/timeout.dart",
    "package:test/src/runner/configuration/suite.dart",
    "package:test/src/runner/engine.dart",
    "package:test/src/runner/plugin/environment.dart",
    "package:test/src/runner/reporter/expanded.dart",
    "package:test/src/runner/runner_suite.dart",
    "package:test/src/utils.dart"
   ],
   "platform": {
    "uses": [
     "dart:async",
     "dart:collection",
     "dart:convert",
     "dart:io",
     "dart:isolate",
     "dart:math",
     "dart:typed_data"
    ]
   }
  }
```

We get `dart:io` here. No way to tell that this is safe for web to import

isoos:
Its direct dependency is [src/executable.dart](https://github.com/dart-lang/test/blob/master/lib/src/executable.dart), which uses `dart:io` in a non-web-safe way. I believe the analysis is correct in this case.

kevmoo:
Nope. src/executable is only imported from bin/test.dart

isoos:
Oh, yeah, my mistake there. However, I've found the culprit:

`package:package_resolver/src/package_root_resolver.dart` pulls the `dart:io` dependency. I think this could be handled with the additional override, and we don't need much else to change. How shall we proceed with it? Rollback the rollback, or re-submit as a new PR?

kevmoo:
Rollback the rollback w/ this update?

kevmoo:
...make sure you test pkg/test and pkg/googleapis – as two good examples

PR #73, opened 2017-08-28
Calculate file and package fitness

isoos:
As usual, naming is hard. `Fitness` is a lightweight version of `health`, with the major difference that `health` would get a deeper analysis, for example:
- running code (e.g. tests)
- code coverage
- readability

It is proportional to the library size (see https://github.com/dart-lang/pub-dartlang-dart/issues/305).

### DIFF
--- a/lib/src/fitness.dart
+++ b/lib/src/fitness.dart
@@ -1,0 +1,114 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library pana.health;
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:math';
+
+import 'package:json_serializable/annotations.dart';
+import 'package:path/path.dart' as p;
+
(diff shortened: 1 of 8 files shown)


lib/src/fitness.dart
```diff
@@ -0,0 +1,105 @@
...
+
+import 'analyzer_output.dart';
+import 'platform.dart';
+import 'summary.dart';
+
+part 'fitness.g.dart';
+
+@JsonSerializable()
+class Fitness extends Object with _$FitnessSerializerMixin {
+  final double value;
```
kevmoo: doc comments on these would be nice – what's the difference between them?

kevmoo:
@isoos - will need to rebase w/ the removal of your last commit

isoos:
Rebased/updated.

Just noticed: the complexity calculation have one weak spot: it doesn't traverse the part 'whatever.dart' statements (yet).

isoos:
> Just noticed: the complexity calculation have one weak spot: it doesn't traverse the part 'whatever.dart' statements (yet).

Never mind, this is a non-issue: the parts are treated as a separate dependency anyway.

PR #74, opened 2017-08-29
Re-submit: Mask only the dart:io library in overrides. Fixes #58.

### LINKED ISSUE #58: Let's remove LibraryScanner overrides

isoos:
Manually checked `test`, `googleapis` and a few other packages. The algorithm for the "transitive deps with override" could use some formal test and verification, but the end-to-end tests should cover us for now.


kevmoo:
I'll throw this through my test set before I merge this time 😄

kevmoo:
Awesome!

kevmoo:
FYI – this makes things a lot better...

PR #75, opened 2017-08-30
Detect broad dependencies

isoos:


lib/src/pubspec.dart
```diff
@@ -30,13 +43,34 @@ class Pubspec {
...
+        if (version == null || version == 'any') {
+          set.add(pkg);
+          return;
+        }
+      }
+
+      final constraint = v is Map ? v['version'] : v;
+      if (constraint is String) {
+        final vc = new VersionConstraint.parse(constraint);
+        if (vc.allowsAny(_tooLargeVersion)) {
```
kevmoo: `vc` is either `VersionRange` or `Version` – I'd just special-case these and avoid the `_tooLargeVersion` variable.

isoos:
PTAL, renamed the field and added it to the fitness calculation too.

PR #76, opened 2017-08-31
Enable custom package directory (for direct downloads or local tests).

isoos:
Slight change of the summary.packageVersion field (non-nullable -> nullable), because there is a slight chance that a local directory is not able to identify the version, and it is better to not fail on the serialization in such cases.

### DIFF
--- a/lib/pana.dart
+++ b/lib/pana.dart
@@ -43,8 +43,12 @@ class PackageAnalyzer {
         dartSdk: _dartSdk, flutterSdk: _flutterSdk, pubCacheDir: pubCacheDir);
   }
 
-  Future<Summary> inspectPackage(String package,
-      {String version, bool keepTransitiveLibs: false}) async {
+  Future<Summary> inspectPackage(
+    String package, {
+    String version,
+    String packageDir,
+    bool keepTransitiveLibs: false,
+  }) async {
     var issues = <AnalyzerIssue>[];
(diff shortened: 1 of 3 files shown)


lib/pana.dart
```diff
@@ -40,8 +40,12 @@ class PackageAnalyzer {
         dartSdk: _dartSdk, flutterSdk: _flutterSdk, pubCacheDir: pubCacheDir);
   }
 
-  Future<Summary> inspectPackage(String package,
-      {String version, bool keepTransitiveLibs: false}) async {
```
kevmoo: Wondering if this should be refactored – so there is a method that just operates off a package in a directory – then wrap it in method that does the download to cache?

isoos: For that I'd slightly modify the PubSummary as it is now: the `pubspec.yaml` parsing should not be blocked on the pub output parsing and conflict detection (there are a few throws there if there is a conflict somewhere). But other than that, I think the two method should be good to go.

isoos: Hm, it is a bit more than just having a new method there, I'd rather do it in a separate PR.

kevmoo:
Now we have merge conflicts 🤷‍♂️

isoos:
rebased

kevmoo:
Travis? 😁

PR #81, opened 2017-09-05
Storing CodeProblem.type values separately, also renaming error field to description

isoos:


lib/src/fitness.dart
```diff
@@ -77,9 +77,9 @@ Future<Fitness> calcFitness(
 
   if (fileAnalyzerItems != null) {
     for (var item in fileAnalyzerItems) {
-      if (item.type.startsWith('INFO')) {
+      if (item.severity.startsWith('INFO')) {
```
kevmoo: can now be `==` now

lib/src/fitness.dart
```diff
@@ -77,9 +77,9 @@ Future<Fitness> calcFitness(
...
         penalties += hintPoints;
-      } else if (item.type.startsWith('WARN')) {
+      } else if (item.severity.startsWith('WARN')) {
```
kevmoo: Ditto - can now be `==` now

isoos: is it 'WARN' or 'WARNING'? I wasn't sure, hence the startsWith

kevmoo:
WARNING – good call!

PR #85, opened 2017-09-07
Refactor PubSummary to PkgResolution

isoos:
This PR has a lot of functional change, I've tried my best to match the old behavior, but there are obvious differences (e.g. to store the dependency type and the constraint type).


kevmoo:
check you formatting?

PR #88, opened 2017-09-11
Better separation for inspect package by directory

isoos:


lib/pana.dart
```diff
@@ -46,38 +44,32 @@ class PackageAnalyzer {
...
+  }
 
-    Version ver;
-    if (version != null) {
-      ver = new Version.parse(version);
-      log.info("Version: $ver");
-    }
+  Future<Summary> inspectDir(String packageDir,
+      {bool keepTransitiveLibs: false}) async {
+    _logEnv();
```
kevmoo: The next call to `_inspect` also calls `_logEnv` – redundant?

lib/pana.dart
```diff
@@ -46,38 +44,32 @@ class PackageAnalyzer {
...
     String version,
-    String packageDir,
     bool keepTransitiveLibs: false,
   }) async {
-    var toolProblems = <ToolProblem>[];
-    var sdkVersion = _dartSdk.version;
-    log.info("SDK: $sdkVersion");
-
-    log.info("Package: $package");
+    _logEnv();
```
kevmoo: Also ends up calling _inspect – perhaps _logEnv should just be within _inspect...

lib/pana.dart
isoos: Good catch! The intent here is to log env information before starting to download the package, and that's reason it is duplicated in both method. Shall I just move both `_inspect`?

kevmoo: Well, it's already there 😄

isoos: done

PR #90, opened 2017-09-13
Detect more license: MIT, Apache, LGPL, MPL, Unlicense.

isoos:


kevmoo:
...I could imagine something like this - https://github.com/dart-lang/dart-samples/blob/master/LICENSE

isoos:
Good point, opened https://github.com/dart-lang/pana/issues/92

PR #97, opened 2017-10-11
Return multiple licenses.

isoos:
(Detection change will be part of a subsequent PR.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 * Greatly expanded and improved license detection.
 
+## 0.5.0
+
+* **Breaking changes**
+
+  * `Summary.license` -> `licenses`: we'll return multiple licenses
+  * Removed `LicenseNames.missing`: empty List will indicate no license file
+
 ## 0.4.0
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 * Greatly expanded and improved license detection.
 
+## [next release]
```
kevmoo: update pubspec to be 0.5.0-dev – then make this `0.5.0`

test/license_test.dart
```diff
@@ -35,7 +35,7 @@ main() {
 
   group('BSD', () {
```
kevmoo: Do we have a test for package w/ multiple licenses?

isoos: I thought I have, but I can't recall it. I remember a license file that has two licenses copied into it.

kevmoo: I don't see a test result that exercises it. Would be nice...

isoos: I'll add this with the next change that will also include the file name in the license object.

Issue #94, opened 2017-09-13
Extend the scope of analyzer options with stagehand defaults?


isoos:
If these are good defaults for stagehand, maybe they are good for pana too?

https://github.com/dart-lang/stagehand/blob/master/templates/package-simple/analysis_options.yaml

kevmoo:
Maybe?... 🤷‍♂️ 

All of these have issues – which is why they are lints and not baked into the platform. Not a high priority IMHO

isoos:
I don't mind being opinionated when the linter rule can expose a high-risk error. It would be nice if we could have a statistics on all of the linter rules and the most popular packages... 📊

PR #98, opened 2017-10-11
A few more linter rules for code health score.

isoos:
Reasons:
- `camel_case_types`: I think it is better to respect the style guide in this instance.
- `iterable_contains_unrelated_type`: high probability for errors.
- `list_remove_unrelated_type`: high probability for errors.
- `unrelated_type_equality_checks`: high probability for errors.

Most of it is coming from is coming from `stagehand` defaults (#94), and I think the rest can be ignored for now.

### DIFF
--- a/lib/src/analysis_options.dart
+++ b/lib/src/analysis_options.dart
@@ -11,15 +11,10 @@ analyzer:
 
 linter:
   rules:
+    - camel_case_types
     - hash_and_equals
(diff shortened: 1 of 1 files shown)


kevmoo:
Seems like this ... disappeared.

Can you see this PR?

isoos:
wow, no idea. I'm not able to create a new one on the same branch, because
pull request already exists...

PR #101, opened 2017-10-13
Detect flutter in dependencies block, also fix releated test. 

isoos:


lib/src/platform.dart
```diff
@@ -127,7 +127,10 @@ DartPlatform classifyPkgPlatform(
   final primaryLibrary =
       _selectPrimaryLibrary(pubspec, transitiveLibs.keys.toSet());
 
-  if (pubspec.hasFlutterKey || pubspec.hasFlutterPluginKey) {
```
kevmoo: Wondering if all of these should just be rolled up into one property on pubspec?

isoos: yeah, I would do that later, after reviewing the very similar Pubspec class in the dart-site project, and unifying all.

PR #102, opened 2017-10-13
License file location in the summary.

isoos:


kevmoo:
Format!

lib/src/license.dart
```diff
@@ -53,22 +55,31 @@ abstract class LicenseNames {
...
-    return [];
-  }
-  return [await detectLicenseInFile(licenseFile)];
+Future<List<LicenseFile>> detectLicensesInDir(String baseDir) async {
+  final rootFiles = await new Directory(baseDir).list().toList();
+  final licenseCandidates =
+      rootFiles.where((fse) => fse is File).where(_isLicenseFile).toList();
+  if (licenseCandidates.isEmpty) return [];
+
+  final licenses = await Future.wait(licenseCandidates.map(
```
kevmoo: Assuming this list is not dozens long, I guess it's fine...

isoos: shall we limit it to 10?

kevmoo: Something like that...

PR #105, opened 2017-10-18
Reduce scope to bin/ and lib/ directories

isoos:
If we are going to ignore the `dev_dependencies`, it doesn't make sense to analyze the `.dart` files outside of the `bin/` or the `lib/` directories. The excluded ones wouldn't contribute to package health or platform analysis anyway, and we aren't losing information with this PR.

If we change our mind later, we can still do a second round of `pub upgrade` with the original `dev_dependencies` and do the required analysis on them. If this and the follow-up dev_dependencies change gets merged, I'm planning to do a follow-up API cleanup to make it configurable which directories we are interested in.


kevmoo:
LGTM. I have some nits, but I'll just layer them on...

PR #107, opened 2017-10-18
Remove dev_dependencies from pubspec.yaml

isoos:


lib/src/sdk_env.dart
```diff
@@ -267,6 +267,20 @@ class PubEnvironment {
       );
     }
   }
+
+  /// Removes the dev_dependencies from the pubspec.yaml
+  Future removeDevDependencies(String packageDir) async {
+    final file = new File(p.join(packageDir, 'pubspec.yaml'));
+    final original = await file.readAsString();
```
kevmoo: you can 'readLines' too, I think

lib/src/sdk_env.dart
```diff
@@ -267,6 +267,20 @@ class PubEnvironment {
...
+    final replaced = original
+        .split('\n')
+        .map((line) => line.trim().startsWith('dev_dependencies')
+            ? line.replaceFirst(
+                'dev_dependencies', '__pub_replaced__dev_dependencies')
```
kevmoo: This really scares me. JSON is valid yaml. You could just read in the Yaml as a map, remove `dev_dependencies` and then write it back as JSON. Should work fine – and be less buggy

isoos: right, fixed it

PR #112, opened 2017-10-26
Tool or code errors should make platform classification 'nowhere'.

isoos:


kevmoo:
Okay...

isoos:
[Removed WIP description]. PTAL. I've checked with a few packages, and this seems to be solid. I've selected a small package for the end-to-end test.

PR #114, opened 2017-10-27
Make all returned scores doubles

kevmoo:


test/scoring_test.dart
```diff
@@ -59,13 +59,13 @@ void main() {
         expect(() => summary.simpleScore(49), throwsRangeError);
         expect(() => summary.simpleScore(151), throwsRangeError);
 
-        expect(summary.simpleScore(150), 1);
+        expect(summary.simpleScore(150) as double, 1);
```
isoos: why is the `as double` needed here?

kevmoo: This ensures that the runtime type of `simpleScore` is a double – and throws otherwise.

It acts like a regression test for the changed behavior

isoos: I think `expect(summary.simpleScore(150), 1.0);` should have the same result. I'm fine either way.

kevmoo: It would – because the comparison between of `1.0` or `1` with `1` is the same.

This is just checking that everything is a `double`

Issue #78, opened 2017-09-01
Use direct downloads of packages instead of using `pub cache`


mkustermann:
We should directly download the packages from the normal `https://pub.dartlang.org/packages/<name>/versions/<version>.tar.gz` urls. Doing so will avoid running into the memcache issue. Furthermore it seems not particularly nice using the internal caching of pub for other purposes.

kevmoo:
Hopefully mitigated by https://github.com/dart-lang/pub-dartlang-dart/pull/338

isoos:
This is now enabled via https://github.com/dart-lang/pana/pull/88, pub site can download the package directly and call analyzer on the directory.

mkustermann:
Is this issue in fact fixed - i.e. does pana now request the url **directly** (as mentioned above)?

Somehow it's not clear from https://github.com/dart-lang/pana/pull/88 . We might go back on the memcaching decision.

isoos:
Pana does not download the URL directly. Pub's `analyzer` could do that though...

mkustermann:
So the issue wasn't fixed after all :-/

isoos:
question is: do we want to handle this in pana or on the pub site?

mkustermann:
If there is a tool which wants to download a package tarball for a specific version then it can and should use the API to do so. Why would it not?

isoos:
At the point of `pub get`, we'll hit the pub site to fetch dependencies. The difference I see is about memcaching: direct download gets you the given version if memcache is failing for whatever reason. And as far as I can recall, we may have skipped the implementation of the direct download just around the time we had consistent memcache, it may be related. But I agree: to minimize the risk, we can download the package directly as .tar.gz

mkustermann:
Of course generally speaking, all instances of the service should be consistent and an upload to one of them should affect subsequent requests to all other instances as well.

Though it doesn't mean that if one knows (e.g. due to polling of pub.dartlang.org for new versions) that a particular (package, version) is now available, it has to check again later: It can and probably should use the direct download instead.

It just happens to be the case that we don't test for existence (therefore also not use memcache) on the server and just redirect to the GCS tarball (so far).

PR #121, opened 2017-11-07
Direct download of package+version

isoos:


lib/src/download_utils.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// Returns a non-null Directory instance only if it is able to download and
+/// extract the direct package dependency. On any failure it clears the temp
+/// directory, otherwise it is the caller's responsibility to delete it.
+Future<Directory> downloadPackage(String package, String version) async {
+  final temp = await Directory.systemTemp.createTemp('pana-');
+  final dir = new Directory(await temp.resolveSymbolicLinks());
+  try {
+    final url =
+        'https://storage.googleapis.com/pub-packages/packages/$package-$version.tar.gz';
```
kevmoo: Sure this handles weird URL characters that can be in versions?

You might consider using `new Uri.https`...

isoos: ah, good point...

lib/src/download_utils.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// Returns a non-null Directory instance only if it is able to download and
+/// extract the direct package dependency. On any failure it clears the temp
+/// directory, otherwise it is the caller's responsibility to delete it.
+Future<Directory> downloadPackage(String package, String version) async {
+  final temp = await Directory.systemTemp.createTemp('pana-');
+  final dir = new Directory(await temp.resolveSymbolicLinks());
+  try {
+    final uri = new Uri.https('storage.googleapis.com',
+        '/pub-packages/packages/$package-$version.tar.gz');
```
mkustermann: This is the wrong URL - the bucket name is an implementation detail! Please see the issue I originally filed at https://github.com/dart-lang/pana/issues/78 !

isoos: hm, we expose this on the versions pages, under the archive icon: https://pub.dartlang.org/packages/pana#pub-pkg-tab-versions

I guess we shall do `https://pub.dartlang.org/packages/<name>/versions/<version>.tar.gz` in both cases.

mkustermann: That's less important (but we could do that).

What's important is that all clients talk to `pub.dartlang.org` and it can decide where to redirect the client to - i.e. that the server is in control.

PR #124, opened 2017-11-13
Maintenance score.

isoos:


lib/src/maintenance.dart
```diff
@@ -0,0 +1,155 @@
...
+];
+
+final List<String> readmeFileNames = const [
+  'readme.md',
+  'readme',
+];
+
+final List<String> analysisOptionsFiles = const [
+  'analysis_options.yaml',
+  '.analysis_options',
```
kevmoo: We should give half credit for `.analysis_options`. Folks should use `analysis_options.yaml`

isoos: good point, fixing.

PR #125, opened 2017-11-13
Detect and check license URL.

isoos:


lib/src/download_utils.dart
```diff
@@ -6,7 +6,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:http/http.dart' as http;
-import 'package:path/path.dart' as path;
+import 'package:path/path.dart' as p;
```
kevmoo: thank you!

PR #127, opened 2017-11-15
Produce human-readable suggestions.

isoos:


test/end2end/pub_server_data.dart
```diff
@@ -304,7 +304,17 @@ final _data = {
...
+    'warningCount': 1,
+    'hintCount': 0,
   },
   'fitness': {'magnitude': 455.0, 'shortcoming': 1.0},
+  'suggestions': [
+    {
+      'level': 'warning',
+      'title': 'Use `analysis_options.yaml`.',
+      'description':
+          'Static analysis will help you to detect bugs and potential issues earlier.',
```
kevmoo: We're analyzing (in pana) all the time. The important bit is that the user has strong-mode enabled.

That really should be the check here...

isoos: PTAL, added strong-mode check.

lib/src/maintenance.dart
```diff
@@ -154,14 +173,64 @@ Future<Maintenance> detectMaintenance(
...
+  if (analysisOptionsExists) {
+    for (var name in analysisOptionsFiles) {
+      final file = new File(p.join(pkgDir, name));
+      if (await file.exists()) {
+        final content = await file.readAsString();
+        try {
+          final Map map = yaml.loadYaml(content);
+          final analyzer = map['analyzer'];
+          if (analyzer != null) {
+            strongModeEnabled = analyzer['strong-mode'] == true;
```
kevmoo: Or! `analyzer['strong-mode']` is a Map – see https://www.dartlang.org/guides/language/analysis-options#specifying-strong-mode

isoos: Good catch! Fixed...

PR #133, opened 2017-11-17
Don't block platform classification on all dartanalyzer error.

isoos:


lib/pana.dart
```diff
@@ -234,16 +235,23 @@ class PackageAnalyzer {
           : !unformattedFiles.contains(dartFile);
       final fileAnalyzerItems =
           analyzerItems?.where((item) => item.file == dartFile)?.toList();
+      final codeErrors =
+          fileAnalyzerItems?.where((cp) => cp.isError)?.toList() ?? const [];
+      final platformBlockers = codeErrors
```
kevmoo: Put this in a const array so it's easier to reason about?

const _ignoreForPlatformsomething something?

PR #136, opened 2017-11-17
`Summary.sdkVersion` is now a `Version` instead of `String`

kevmoo:
Fixes https://github.com/dart-lang/pana/issues/135

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -31,14 +31,15 @@ class DartSdk implements DartSdkInfo {
         _dartfmtCmd = _join(sdkDir, 'bin', 'dartfmt'),
         _pubCmd = _join(sdkDir, 'bin', 'pub');
 
-  factory DartSdk({String sdkDir, Map<String, String> environment}) {
+  static Future<DartSdk> create(
+      {String sdkDir, Map<String, String> environment}) async {
     sdkDir ??= cli.getSdkPath();
     environment ??= {};
 
     var dartCmd = _join(sdkDir, 'bin', 'dart');
 
     var r = handleProcessErrors(
-        Process.runSync(dartCmd, ['--version'], environment: environment));
+        await runProc(dartCmd, ['--version'], environment: environment));
     var versionString = r.stderr.toString().trim();
 
     var info = new DartSdkInfo.parse(versionString);
@@ -135,9 +136,8 @@ class PubEnvironment {
   final String pubCacheDir;
   final Map<String, String> _environment = {};
 
-  PubEnvironment({DartSdk dartSdk, FlutterSdk flutterSdk, this.pubCacheDir})
-      : this.dartSdk = dartSdk ?? new DartSdk(),
-        this.flutterSdk = flutterSdk ?? new FlutterSdk() {
+  PubEnvironment(this.dartSdk, {FlutterSdk flutterSdk, this.pubCacheDir})
+      : this.flutterSdk = flutterSdk ?? new FlutterSdk() {
     _environment.addAll(this.dartSdk._environment);
     if (!_environment.containsKey(_pubEnvironmentKey)) {
       // Then do the standard behavior. Extract the current value, if any,
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -8,43 +8,46 @@ import 'dart:io';
 
 import 'package:cli_util/cli_util.dart' as cli;
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 
 import 'analysis_options.dart';
 import 'logging.dart';
 import 'utils.dart';
 
+final _sdkRegexp =
+    new RegExp('Dart VM version:\\s([^\\s]+)\\s\\(([^\\)]+)\\) on "(\\w+)"');
+
 class DartSdk {
-  final String _sdkDir;
-  final Map<String, String> _environment = {};
-  final String _dartCmd;
+  final String sdkDir;
+  final Map<String, String> _environment;
   final String _dartAnalyzerCmd;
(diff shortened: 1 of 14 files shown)


kevmoo:
This is a *super* break – because it'll no longer parse 0.6.x summaries.

We could modify the version parsing logic to handle it.

Thoughts?

isoos:
> This is a super break – because it'll no longer parse 0.6.x summaries.

ugh, yeah. options: (1) rename it to `dartVersion`, (2) don't care, give `analyzer` service 2+ days headstart on next deploy.

lib/src/sdk_env.dart
```diff
@@ -8,43 +8,46 @@ import 'dart:io';
 
 import 'package:cli_util/cli_util.dart' as cli;
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 
 import 'analysis_options.dart';
 import 'logging.dart';
 import 'utils.dart';
 
+final _sdkRegexp =
```
isoos: can we get await with a simpler if (version.startsWith('Dart VM version:)) {
   // get first string after colon
}?
the date and the rest is not really important. Maybe store the original value and extract only the version?

kevmoo:
Just pushed some more fun. PTAL

lib/src/summary.dart
```diff
@@ -120,8 +122,16 @@ class Summary extends Object with _$SummarySerializerMixin {
       this.suggestions,
       {this.flutterVersion});
 
-  factory Summary.fromJson(Map<String, dynamic> json) =>
-      _$SummaryFromJson(json);
+  factory Summary.fromJson(Map<String, dynamic> json) {
+    var panaVersion = new Version.parse(json['panaVersion']);
+    if (panaVersion.major == 0 && panaVersion.minor == 6) {
```
isoos: panaVersion.minor <=6

lib/src/summary.dart
```diff
@@ -120,8 +122,16 @@ class Summary extends Object with _$SummarySerializerMixin {
...
+      // Update the json in-place to have the expected values
+      var info = new DartSdkInfo.parse(json['sdkVersion']);
+      json['sdkVersion'] = info.version.toString();
```
isoos: info.toJson() ?

isoos: ah, it is not a @JsonSerializable... it could be though... if we are breaking things, we could do a full break as well :)

lib/src/summary.dart
```diff
@@ -107,6 +105,9 @@ class Summary extends Object with _$SummarySerializerMixin {
   @JsonKey(includeIfNull: false)
   final List<Suggestion> suggestions;
 
+  final PkgResolution pkgResolution;
```
isoos: can we do this in a separate change? this is going to be big with all the test data changes...

lib/src/summary.dart
kevmoo: `Version` is not in our package. We add support in w/ a custom TypeHelper in the tool dir.

I'd rather not churn thing...

lib/src/summary.dart
kevmoo: ah poo...forgot about tests. Yup

kevmoo: PTAL

lib/src/summary.dart
isoos: What I meant was that DartSdkInfo could be an object that we store in `Summary`.

lib/src/summary.dart
kevmoo: done

lib/src/summary.dart
kevmoo: Ah...er...um...sure?

kevmoo: We just want the SDK version in summary. The date/platform is just noise at the moment.

While I was parsing, I wanted to get everything.

Let's hold off until we want other things...at least that's my thought

lib/src/sdk_env.dart
```diff
@@ -8,43 +8,43 @@ import 'dart:io';
...
-  factory DartSdk({String sdkDir, Map<String, String> environment}) =>
-      new DartSdk._(sdkDir ?? cli.getSdkPath(), environment);
+  factory DartSdk({String sdkDir, Map<String, String> environment}) {
+    sdkDir ??= cli.getSdkPath();
+    environment ??= {};
 
-  String get sdkDir => _sdkDir;
+    var dartCmd = _join(sdkDir, 'bin', 'dart');
 
-  String get version {
```
isoos: I'd rather have something along the lines of:

```
DartSdkInfo _version;
Future<DartSdkInfo> get version async {
  _version ??= await _getVersion();
  return _version;
}

Future<String> _getVersion() async {
  // Process.run(_dartCmd, ['--version'), environment: _environment);
}
```

That way the initialization is kept simple.

kevmoo: hrm...okay. Will ponder...

kevmoo: I'd almost rather the ctor of the class fail than have an async error later...we can discuss next week

isoos: One drawback is: we have no control over`Process.runSync`'s timeout. If it were in an async block (either by doing the creation in an `static Future<DartSdk> newDartSdk()`  method or in a separate `Future init()`, we could set a timeout. Maybe the static method would be a better approach here?

kevmoo: Hrm. Sure. I'm not too worried about running this command timing out. If it fails, it'll fail immediately...unless the target system does some crazy RPC when a command is not found.

I love having the instance locked down right away.

Maybe put in an async factor method (not ctor)

kevmoo:
@isoos PTAL

PR #137, opened 2017-11-21
Summary: move pkgResolution and dartFiles to end of serialized output

kevmoo:
Makes scanning the raw file easier


isoos:
strange that tests doesn't need change... I guess order of the keys doesn't matter.

isoos:
btw. it is not clear whether we are optimizing to read the beginning or the end of the json ;)

kevmoo:
I'm going for the beginning 😄

PR #140, opened 2017-11-22
Logging tweaks, code layout, prepare for 0.7.0

kevmoo:


kevmoo:
> Looks good. I'll use the url rewrite stuff from download_utils.dart, but there is no need to export it yet, I'd rather check first how it will get used.

Uh...okay!

Issue #143, opened 2017-11-27
Classify crashes of tools differently


kevmoo:
See https://github.com/dart-lang/pana/issues/138

There should be a `Suggestion.bug` or similar that tells users to file an issue.

We need to be able to find these types of issues quickly and (hopefully) fix them.

PR #144, opened 2017-11-27
Categorize suggestions as bugs if they represent tool failures

kevmoo:
Closes https://github.com/dart-lang/pana/issues/143


kevmoo:
@isoos happy to run at this another way. Or you're welcome to run at it.

Just want to make sure we can find these types of issues quickly.

PR #145, opened 2017-11-27
Release

kevmoo:


lib/src/summary.dart
```diff
@@ -156,21 +156,11 @@ class Suggestion extends Object
...
-          stack
-        ].join('\n'))
-        .take(100)
-        .join('\n');
-
-    return new Suggestion(
-        SuggestionLevel.bug,
-        'There is a bug in the `pana` package or one of its dependencies.',
-        lines);
+    var lines =
```
isoos: Why remote the "File an issue..." part?

kevmoo: Because context matters. If the package site hits a bug, it should print its own `report an issue` link, ideally with a pre-filled body w/ the pana version, etc.

Don't want to sprinkle links all over

Issue #138, opened 2017-11-21
Stack overflow w/ angular_ast at 0.4.0-alpha+2


kevmoo:
```
dart --checked bin/main.dart angular_ast 0.4.0-alpha+2
```

```
SEVERE     Error scanning direct librariers
           Stack Overflow
           #0      _HashMap.putIfAbsent (dart:collection-patch/dart:collection/collection_patch.dart:130)
           #1      CacheEntry.getResultData (package:analyzer/src/context/cache.dart:423:23)
           #2      CacheEntry._setErrorState (package:analyzer/src/context/cache.dart:800:27)
           #3      CacheEntry._setErrorState.<anonymous closure> (package:analyzer/src/context/cache.dart:813:17)
           #4      _HashFieldBase&_HashBase&_OperatorEqualsAndHashCode&SetMixin.forEach (dart:collection/set.dart:145)
           #5      CacheEntry._setErrorState (package:analyzer/src/context/cache.dart:808:31)
           #6      CacheEntry._setErrorState.<anonymous closure> (package:analyzer/src/context/cache.dart:813:17)
           #7      _HashFieldBase&_HashBase&_OperatorEqualsAndHashCode&SetMixin.forEach (dart:collection/set.dart:145)
           #8      CacheEntry._setErrorState (package:analyzer/src/context/cache.dart:808:31)
           #9      CacheEntry._setErrorState.<anonymous closure> (package:analyzer/src/context/cache.dart:813:17)
           #10     _HashFieldBase&_HashBase&_OperatorEqualsAndHashCode&SetMixin.forEach (dart:collection/set.dart:145)
           ...
           ...
           #13660  _RootZone.runUnary (dart:async/zone.dart:1379)
           #13661  _FutureListener.handleValue (dart:async/future_impl.dart:129)
           #13662  _Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:635)
           #13663  _Future._propagateToListeners (dart:async/future_impl.dart:664)
           #13664  _Future._completeWithValue (dart:async/future_impl.dart:477)
           #13665  _Future._asyncComplete.<anonymous closure> (dart:async/future_impl.dart:509)
           #13666  _microtaskLoop (dart:async/schedule_microtask.dart:41)
           #13667  _startMicrotaskLoop (dart:async/schedule_microtask.dart:50)
           #13668  _runPendingImmediateCallback (dart:isolate-patch/dart:isolate/isolate_patch.dart:111)
[...]

PR #148, opened 2017-11-29
Use analyzer 0.31.0-alpha.2

isoos:
#138


isoos:
`dartanalyzer` on SDK 1.24.2 fails with this one. Shall we bump the SDK constraint too?

kevmoo:
Here's the annoying thing. We cannot publish the package w/ a dependency override.

We'll have to "fix" this on the pub site. And leave this open until we have a new analyzer version.

PR #153, opened 2017-12-02
Removed `hash_and_equals` from the default set of lints.

kevmoo:
There are valid reasons to not follow this rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.7.3
+
+* Removed `hash_and_equals` from the default set of lints.
+  There are valid reasons to not follow this rule.
+
 ## 0.7.2
 
 * Handle more critical exceptions and report them with more details.
--- a/lib/src/analysis_options.dart
+++ b/lib/src/analysis_options.dart
@@ -12,7 +12,6 @@ analyzer:
 linter:
   rules:
     - camel_case_types
-    - hash_and_equals
     - iterable_contains_unrelated_type
     - list_remove_unrelated_type
     - unrelated_type_equality_checks
--- a/lib/src/version.g.dart
+++ b/lib/src/version.g.dart
@@ -10,4 +10,4 @@ part of pana.version;
 // Generator: PackageVersionGenerator
 // **************************************************************************
 
-final _$panaPkgVersionPubSemverVersion = new Version.parse("0.7.2");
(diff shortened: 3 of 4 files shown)


matanlurey:
I honestly wouldn't use any lints to determine quality score.

We don't have any other tooling that tells you to use them, and internally we've banned quite a bit.

kevmoo:
> I honestly wouldn't use any lints to determine quality score.

Hrm. The set we have here are pretty safe, I think. What do you think, @isoos ?

matanlurey:
I'm not sure it's fair to punish people who are 100% strong-mode clean, and don't know about another package's rules to determine package health. But I'll let you decide if we should do it or not.

isoos:
1) lint-related code checks may decide whether you get 97 or 96, but their impact is roughly that. More precisely: if you get roughly ~1000 lines, and you have one INFO lint failure, your score will be 999/1000, which on the UI will be 100.

2) Lints are useful signals that can detect problems. If you know better, you can always do in-line comment exclude, and make the lint message go away, and `pana` won't see it (or penalize it) either. It is about awareness: we see a potentially harmful pattern, and the code author may or may not indicate whether they are aware of it (either by fixing it or by the comment).

3) You can get equals and hashCode generated for free in IntelliJ. They are linked so much that there isn't a separate menu item to generate equals only or the other. Unless you have a very complex and elaborate case, I cannot imagine why you'd have one, and not the other. In that case, add the comment and you'll be fine.

4) I'm pretty sure that sooner or later everybody will raise an objection for everything that makes their code look a bit less than the rest. At that point we can just remove the code health metric, because it becomes meaningless.

isoos:
TL;DR: no, I wouldn't remove this lint rule.

isoos:
Btw. stagehand has it in their templates too: https://github.com/dart-lang/stagehand/blob/master/templates/package-simple/analysis_options.yaml

kevmoo:
Hrm...okay

matanlurey:
I'm still totally against including any lints in the health score, but I also don't care to argue too much about this. If I see packages having lower scores for not adhering to lints they don't exist, I'll raise this issue again.

kevmoo:
> I'm still totally against including any lints in the health score, but I also don't care to argue too much about this. If I see packages having lower scores for not adhering to lints they don't exist, I'll raise this issue again.

Understood. We need to do some coordination here w/ stagehand, etc.

PR #156, opened 2017-12-05
Global activate

kevmoo:
CC @natebosch


isoos:
the build test needs to get fixed

kevmoo:
> We could advise pub global run pana:main, but rename is better.

Agreed!

PR #162, opened 2017-12-07
Detect the new format of native extensions

isoos:


kevmoo:
@isoos – need to run tool/build.dart

PR #172, opened 2017-12-08
Changed the meaning of version fields in `Maintenance`:

kevmoo:
`isExperimentalVersion` now means pre-V1.
`isPreReleaseVersion` now means there is a pre-release flag
  like `-beta`, `-alpha`, etc.


kevmoo:
This aligns better w/ the names we have in the pubspec. I'm not super worried about `0.0.*` vs `0.*`

kevmoo:
When there is pana version mismatch, we'll get some weirdness – but it should be short lived

PR #174, opened 2017-12-09
Maintenance: add optional `now` positional param to getMaintenanceScore

kevmoo:
Allows testing
Also added tests for Maintenance

### DIFF
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -101,8 +101,9 @@ class Maintenance extends Object with _$MaintenanceSerializerMixin {
   factory Maintenance.fromJson(Map<String, dynamic> json) =>
       _$MaintenanceFromJson(json);
 
-  double getMaintenanceScore(DateTime publishDate) {
-    final now = new DateTime.now().toUtc();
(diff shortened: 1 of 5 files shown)


lib/src/maintenance.dart
```diff
@@ -100,8 +100,8 @@ class Maintenance extends Object with _$MaintenanceSerializerMixin {
   factory Maintenance.fromJson(Map<String, dynamic> json) =>
       _$MaintenanceFromJson(json);
 
-  double getMaintenanceScore(DateTime publishDate) {
-    final now = new DateTime.now().toUtc();
+  double getMaintenanceScore(DateTime publishDate, {DateTime now}) {
```
isoos: Would be to have `Duration age` as the only parameter?

kevmoo: Ha! that's next

kevmoo: Wanted to avoid breaks until we got good tests...

LGTY?

isoos: please add a TODO there

PR #176, opened 2017-12-09
Unblock platform classification on invalid method overrides.

isoos:
This typically happens when somebody writes an equals or compareTo with the specific type instead of the generic (or vice versa).


kevmoo:
@isoos – travis?

PR #177, opened 2017-12-10
age and explainScore

kevmoo:


lib/src/maintenance.dart
```diff
@@ -101,43 +101,135 @@ class Maintenance extends Object with _$MaintenanceSerializerMixin {
...
+final _explainTypeStrings = new Map<ExplainType, String>.fromIterable(
+    ExplainType.values,
+    key: (e) => e as ExplainType,
+    value: (e) => e.toString().split('.').last);
+
+class ExplainScore {
+  final ExplainType explanation;
+  final double simpleScale;
+  final int exponent;
+  final double scale;
```
isoos: why not just a single score and an explanation?

lib/src/maintenance.dart
```diff
@@ -101,43 +101,135 @@ class Maintenance extends Object with _$MaintenanceSerializerMixin {
...
+
+  static final _vintage = new ExplainScore._(ExplainType.vintage, 0.0);
+  static final _updated = new ExplainScore._(ExplainType.vintage, 1.0);
```
isoos: this is very specific to maintenance score, I would hope we could similar structure for code health too...

lib/src/maintenance.dart
kevmoo: I want the "math" behind these to be 100% transparent. See toString. I want to put the math on the analysis page so folks understand how we get the numbers.

lib/src/maintenance.dart
kevmoo: ack

kevmoo:
@isoos –PTAL

lib/src/maintenance.dart
isoos: I still think this is a bit more complicated than it needs to be. We could create a text log of what happens, alongside the result of the calculation, without the need of a new structure. It will be a human who will read it, no need to return a parsed structure or a pseudocode here.

We also have a `Suggestion` class which will be a duplicate for most of these: if there is a penalty, we need to explain it and suggest something to be fixed. Maybe this log could be part of the description in a suggestion?

PR #182, opened 2017-12-12
Handle severe errors

kevmoo:


lib/src/package_analyzer.dart
```diff
@@ -46,12 +46,12 @@ class PackageAnalyzer {
...
-    String package, {
-    String version,
-    bool keepTransitiveLibs: false,
-    Logger logger,
-  }) async {
+  Future<Summary> inspectPackage(String package,
+      {String version,
+      bool keepTransitiveLibs: false,
+      Logger logger,
+      bool deleteTemporaryDirectory: true}) async {
```
isoos: I have a slight preference towards false default values, and in this case the name would be `keepTemporaryDirectory`. But the current works just as good.

PR #186, opened 2017-12-14
Use PUB_HOSTED_URL for downloads.

isoos:


kevmoo:
@isoos – I think we need to exclude or `allow failures` for analyzer in stable...

isoos:
added type, let's see how analyzer likes it now

Issue #180, opened 2017-12-12
Allow running pana on a pub server that isn't pub.dartlang.org


robbecker-wf:
# Problem
Workiva runs our own internal pub server due to IP concerns. Pana looks to have hard-coded the main pub server here: 
https://github.com/dart-lang/pana/blob/master/lib/src/download_utils.dart#L22

This means that we can't currently run it to analyze packages hosted on our internal server. I'd like to see a command-line parameter to specify the pub server to use.

On a related note: It should probably also handle sub dependencies declared in pubspec.yaml that are hosted on a different pub server.

robbecker-wf:
@kevmoo ping

kevmoo:
Ack. Pull requests welcome, @robbecker-wf !

kevmoo:
@robbecker-wf – this landed with https://pub.dartlang.org/packages/pana/versions/0.8.0

Let us know how it works!

robbecker-wf:
@kevmoo  Worked pretty well. It runs to completion and I got JSON output, but somehow it seemed to stick to PUB_HOSTED_URL when generating suggestions and couldn't find a public package (async in this case) because it was using the wrong server and should have used pub.dartlang.org

Example snippet:
```
 "suggestions": [
  {
   "level": "error",
   "title": "Fix dependencies in `pubspec.yaml`.",
   "description": "Running `pub upgrade` failed with the following output:\n\n```\nERR: Could not find package async at https://pub.workiva.org.\n Depended on by:\n - pub itself \n```\n"
```

isoos:
How are you using `pub get` / what is in your `pubspec.yaml` for referencing private and public projects?

robbecker-wf:
Public packages are generally just given a version or range. Internal packages are referenced with the full server info. 

Example pubspec:
```
dependencies:
  some_other_internal_package:
    hosted:
      name: some_other_internal_package
      url: https://pub.workiva.org
    version: ^2.0.0
  args: '>=0.13.7 <2.0.0'
```

I ran pana like so: `PUB_HOSTED_URL=https://pub.workiva.org pana an_internal_package`

isoos:
Though. @kevmoo: any idea how to resolve this?

kevmoo:
> Though. @kevmoo: any idea how to resolve this?

Yup! We need to ditch the `PUB_HOSTED_URL` logic and add an option to specify the server

Then the first download uses the server – and pub uses the `hosted` args in the downloaded package to do the right thing.

[...]

PR #190, opened 2017-12-19
--hosted-url instead of PUB_HOSTED_URL

isoos:
#180.

While this is a breaking change, it is not breaking of the format, and the feature was never really working, hence the 0.8.0+1 version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.8.1-dev
+
+* **BREAKING BEHAVIOR**: Don't use `PUB_HOSTED_URL` for package downloads,
+  as it has not worked out in practice. Instead, we've added a `--hosted-url`
+  command line argument.
+
 ## 0.8.0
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.8.0+1
```
kevmoo: At least call this 0.8.1? Or does that bork up the package site?

bin/pana.dart
```diff
@@ -17,7 +17,8 @@ final _parser = new ArgParser()
       abbr: 'j',
       help: 'Output log items as JSON.',
       defaultsTo: false,
-      negatable: false);
+      negatable: false)
+  ..addOption('hosted-url');
```
kevmoo: A tiny bit of help content here would be nice...

kevmoo:
@isoos – also, make travis happy 😄

PR #191, opened 2017-12-19
Flutter-specific dartanalyzer options.

isoos:
We can fine-tune how to synchronize the linter rules between the two, I've just copied the flutter's options file as it was to unblock supermixin.

### DIFF
--- a/lib/src/analysis_options.dart
+++ b/lib/src/analysis_options.dart
@@ -18,3 +18,33 @@ linter:
     - unrelated_type_equality_checks
     - valid_regexps
 ''';
+
+// Keep it updated with
+// https://github.com/flutter/flutter/blob/master/packages/flutter/lib/analysis_options_user.yaml
+const String flutterAnalysisOptions = '''
+analyzer:
+  language:
+    enableStrictCallChecks: true
+    enableSuperMixins: true
+    enableAssertInitializer: true
+  strong-mode: true
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: warning
+    # treat missing returns as a warning (not a hint)
+    missing_return: warning
+    # allow having TODOs in the code
+    todo: ignore
+
+# Source of linter options:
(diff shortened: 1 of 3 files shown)


lib/src/analysis_options.dart
```diff
@@ -13,8 +16,80 @@ linter:
...
+  strong-mode: true
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: warning
+    # treat missing returns as a warning (not a hint)
+    missing_return: warning
+    # allow having TODOs in the code
+    todo: ignore
+
+linter:
```
kevmoo: Let's remove these. return types on setters, use isNotEmpty, etc have no affect on runtime.

Let's use one set of lints and just turn on the language, strong-mode, and error tweaks

isoos: Like this? or do you want to have the error tweaks in the default options too?

kevmoo: Yup!

We could even share the lints in a variable...but not a biggy.

isoos: yeah. the benefit of this version: easy to copy-paste. Can do the variable thing if you prefer that.

kevmoo:
What version should this go in? 0.8.1?

Should flag as -dev, etc...unless we want to roll this out with the Pub-host changes, too...

isoos:
> What version should this go in? 0.8.1?

I omitted that because I didn't want to have a conflict with the other PR. I'll change it to 0.8.1 there and we can append the changelog of this PR.

lib/src/analysis_options.dart
kevmoo: No biggy

kevmoo:
> I omitted that because I didn't want to have a conflict with the other PR. I'll change it to 0.8.1 there and we can append the changelog of this PR.

Okay. I really like using -dev for pushed that aren't designed to be published. Not a biggy, though.

PR #206, opened 2018-01-03
Enable local project analysis through command-line.

### LINKED ISSUE #204: command line option to execute pana against local directory

isoos:
- Creates backup of `pubspec.yaml` and restores it.
- Fixes #204

### DIFF
--- a/bin/pana.dart
+++ b/bin/pana.dart
@@ -3,6 +3,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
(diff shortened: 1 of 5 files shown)


bin/pana.dart
```diff
@@ -106,15 +99,24 @@ main(List<String> args) async {
   try {
     try {
       var analyzer = await PackageAnalyzer.create(pubCacheDir: tempPath);
-      var summary = await analyzer.inspectPackage(pkg,
-          version: version, pubHostedUrl: result['hosted-url']);
 
+      final target = result.rest.first;
+      final isLocalDir = await FileSystemEntity.isDirectory(target);
```
kevmoo: Uh...not ideal. there is a package `test` for example.

I'd vote for `--source` with `site` (default) and `directory` as options.

isoos: good point. I'm not sure why you would need `site`, I've added `--source <directory>`, PTAL

kevmoo: Ah! See below for pub global activate

<package> is usually the package name (from hosted).

--source let's you say "no, <package> is a directory", etc.

Would be nice to align w/ what pub does here anyway...

```
Usage: pub global activate <package...>
-h, --help              Print this usage information.
-s, --source            The source used to find the package.
                        [git, hosted (default), path]

-f, --features          Feature(s) to enable.
-F, --omit-features     Feature(s) to disable.
    --no-executables    Do not put executables on PATH.
-x, --executable        Executable(s) to place on PATH.
    --overwrite         Overwrite executables from other packages with the same name.

Run "pub help" to see global options.
```

isoos: How about now?

isoos:
> Wondering if some sort of warning/FYI message would be good – telling folks that their pubspec.yaml file will be messed with?

In the log or upfront?

kevmoo:
> In the log or upfront?

🤷‍♂️

I'd almost argue for a forced "are you sure?" prompt that can be overridden w/ --force

isoos:
> I'd almost argue for a forced "are you sure?" prompt that can be overridden w/ --force

I think the `--source path` is a strong enough intent. I've updated the code using a `--no-warning` flag, and if you omit it, the analysis will be delayed by 15 seconds.

Issue #208, opened 2018-01-04
Add suggestion around examples


kevmoo:
See https://github.com/dart-lang/pub-dartlang-dart/blob/1ca93dcb1d1e1489bec4768de553b9091974cc58/app/lib/shared/utils.dart#L226

This helper should be moved to pana and used to suggest folks add an example to their package.

Then the pub site can use the same logic. Magic!

isoos:
Different kinds of suggestion would be helpful here:
- the package has no examples
- the package has files in the example directory, but not with recognized pattern
- the package has a recognized example, but seems to be too short or too long

kevmoo:
Yup!

PR #210, opened 2018-01-04
Add suggestions (as hints) about missing example.

isoos:


isoos:
It doesn't handle the 'too small / too large' example edge cases, but I'd wait for a real-world example before implementing that. Closes #208.

lib/src/maintenance.dart
```diff
@@ -206,6 +223,22 @@ Future<Maintenance> detectMaintenance(
         'Readme should inform others about your project, what it does, and how they can use it.',
         penalty: new Penalty(fraction: 500)));
   }
+  if (!exampleExists) {
+    final exampleDirExists = files.any((file) => file.startsWith('example/'));
+    if (exampleDirExists) {
+      maintenanceSuggestions.add(new Suggestion.hint(
+          'Maintain an example.',
+          'None of the files in your `example/` directory matches a known example patterns. '
```
kevmoo: We could print out a full list of the supported example names, right?

isoos: I wouldn't duplicate the list with the `lib/*` versions, and the `$package_example.dart` seemed a bit too long here. The others are listed here.

kevmoo:
SGTM

PR #211, opened 2018-01-05
Recommend descriptions between 60 and 180 characters.

### LINKED ISSUE #201: Recommend descriptions between 70 and 150 characters
There doesn't seem to be any official numbers from Google - https://support.google.com/webmasters/answer/35624?hl=en

But the webmaster tools does report when things are too short or too long.

It seems that >= 70 and <=150 is the conscious - https://www.quora.com/How-long-should-a-meta-description-be-I-have-warnings-on-Google-Webmaster-tools-stating-some-are-too-short

isoos:
I've expanded the range, because people tend to follow these numbers too closely, and I think a bit of buffer zone eases the pressure there. Fixes #201.

### DIFF
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -11,9 +11,9 @@ import 'dart:math';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
-import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart' as yaml;
 
+import 'pubspec.dart';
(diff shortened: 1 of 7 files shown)


lib/src/maintenance.dart
```diff
@@ -274,6 +276,29 @@ Future<Maintenance> detectMaintenance(String pkgDir, String pkgName,
         penalty: new Penalty(fraction: 200)));
   }
 
+  // Checking the length of description.
+  final description = pubspec.description?.trim();
+  if (description == null || description.isEmpty) {
+    maintenanceSuggestions.add(new Suggestion.warning(
```
kevmoo: Pretty sure `pub` won't let you upload a package without a description – but not bad to have, I guess.

PR #212, opened 2018-01-05
Require to specify version when using --hosted-url.

isoos:


isoos:
I think this will complete #180.

PR #215, opened 2018-01-09
Ignore uri_has_not_been_generated errors.

### LINKED ISSUE #213: Files that import .template.dart files that are generated by Angular produce failures in Strong Mode analysis.
Example:
https://pub.dartlang.org/packages/angular_components/versions/0.9.0-alpha%2B2#-analysis-tab-

The `uri_has_not_been_generated: ignore` option I have in the `analysis_options.yaml` file is not being used. https://github.com/dart-lang/angular_components/blob/master/analysis_options.yaml#L8

isoos:
Fixes #213

### DIFF
--- a/lib/src/analysis_options.dart
+++ b/lib/src/analysis_options.dart
@@ -2,7 +2,11 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-const String analysisOptions = '''
+import 'dart:convert';
+
+import 'package:yaml/yaml.dart' as yaml;
+
+const String _analysisOptions = '''
 analyzer:
   strong-mode: true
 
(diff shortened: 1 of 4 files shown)


isoos:
I still think that this is a bad precedent.

lib/src/analysis_options.dart
```diff
@@ -5,6 +5,8 @@
 const String analysisOptions = '''
 analyzer:
   strong-mode: true
+  errors:
+    uri_has_not_been_generated: ignore
```
kevmoo: No!

I think the idea is: if the source package has this setup – then copy it.

Just whitelist `uri_has_not_been_generated` for pass-through

lib/src/analysis_options.dart
```diff
@@ -48,3 +52,35 @@ linter:
     - unrelated_type_equality_checks
     - valid_regexps
 ''';
+
+const _analyzerErrorKeys = const <String>['uri_has_not_been_generated'];
+
+String customizeAnalysisOptions(String original, bool isFlutter) {
```
kevmoo: TODO: add an unit test for this – since it's more than a bit complicated

kevmoo: @isoos – did you sanity check it w/ one of the samples we care about?

isoos: Added tests, which caught an edge-case that was not caught with the sampling manual test.

kevmoo: 😄

PR #217, opened 2018-01-09
Better handling of repository-relative URLs.

isoos:
This will be reused in pub site too.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -45,13 +45,31 @@ Future<Directory> downloadPackage(String package, String version,
 }
 
 /// Returns an URL that is likely the downloadable URL of the given path.
-String joinDownloadUrl(String baseUrl, String relativePath) {
-  if (baseUrl == null || baseUrl.isEmpty) return null;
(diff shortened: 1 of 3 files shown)


lib/src/download_utils.dart
```diff
@@ -45,13 +45,31 @@ Future<Directory> downloadPackage(String package, String version,
...
+String getRepositoryUrl(String repository, String relativePath) {
+  if (repository == null || repository.isEmpty) return null;
+  try {
+    final uri = Uri.parse(repository);
+    final segments = new List<String>.from(uri.pathSegments);
+    while (segments.isNotEmpty && segments.last.isEmpty) {
+      segments.removeLast();
+    }
+
+    if (repository.startsWith('https://github.com/') ||
```
kevmoo: Identical behavior for gitlab, too?

isoos: yeah, gitlab copies the github behavior in many ways

isoos:
> if we want to link to images, we'll need to use the raw host, right?

Yeah, I was thinking of that too, but for non-images, the non-raw should be better. I'll add it in a follow-up.

PR #221, opened 2018-01-11
Use raw links for images in repository URLs.

isoos:
- github redirects it to raw.githubusercontent.com (or something)
- gitlab will keep serving the file with the /raw/ path segment

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -11,6 +11,8 @@ import 'package:path/path.dart' as p;
 import 'logging.dart';
 import 'utils.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/download_utils.dart
```diff
@@ -56,12 +56,19 @@ String getRepositoryUrl(String repository, String relativePath) {
 
     if (repository.startsWith('https://github.com/') ||
         repository.startsWith('https://gitlab.com/')) {
+      final relativePathLower = relativePath.toLowerCase();
+      final isRaw = relativePathLower.endsWith('.png') ||
```
kevmoo: Better to have a const list of all of the extensions and hit that list. IMHO

Issue #222, opened 2018-01-11
penalties should be reflected in suggestions


isoos:
- platform conflict (may be moved to maintenance)
- dependency without constraint (may be moved to maintenance)
- should note that files need formatting
- ~~should note that files have analysis errors, warning or hints~~

PR #223, opened 2018-01-12
Move unconstrained version penalty from health score to maintenance.

isoos:
#222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Use raw links for images in repository URLs.
 
+* Move unconstrained version penalty from health score to maintenance.
+
 ## 0.9.0+1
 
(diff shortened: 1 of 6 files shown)


mit-mit:
cc @isoos looks like some tests are failing in CI?

isoos:
Fix in progress.

I used to receive notifications when the CI failed, but not anymore. Do you happen to know where I need to set the appropriate checkbox to get notified?

isoos:
Oh, the current CI issue seems to be coming from the upgrade of the SDK to the dev 17. It is generating a bit different analyzer output, which is unrelated to my change, I'll address it separately. Not sure why we don't control the version in the CI config.

isoos:
CI cleared after we have pinned the Dart SDK version.

PR #224, opened 2018-01-12
Control Dart SDK version in CI.

isoos:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: dart
 
 dart:
-  - dev
+  - 2.0.0-dev.15.0
```
kevmoo: Uh...okay...so we sync this with what we have the pub site on?

isoos: We don't need to sync those. The version is important only for pana tests, to control the SDK version upgrades where `dartanalyzer` introduces a new hint or warning, or deprecates and older one.

PR #230, opened 2018-01-17
New platform classification

isoos:
Used separate commits for the different parts of the change.

The main changes are:
- Instead of a list of platforms (`worksEverywhere` + `restrictedTo`), we will report the components (e.g. `flutter`, `html`, `io`, `mirrors`...), and the actual use status of the platforms.
- `components` will be used to report more details on the analysis tab
- `uses` will be used to create the platform list tags for each package
- removed `server` platform, using a wider `other`.


isoos:
@jakobr-google: I've updated with another commit, which deals with the server -> other transition. PTAL.

PR #237, opened 2018-01-22
Add new platform component: build tools.

isoos:


lib/src/platform.dart
```diff
@@ -93,6 +96,15 @@ class ComponentDef {
     ],
   );
 
+  /// dart:nativewrappers and related libraries
+  static const ComponentDef build = const ComponentDef(
+    ComponentNames.build,
+    const <String>[
+      'package:barback',
```
kevmoo: I'd love to sanity check this before it goes out. A LOT of packages (angular, for instance) are web tools that depend on barback/build. We don't want them defined as build tools instead...

isoos: This is for the first phase of classification, identifying generic components. The current goal is to make it visible on the analysis page that the given package is using some kind of build tool component, but with the current rules it won't take away angular's `web` classification, nor would it classify as `other` or the not-yet-existent `build` platform.

If we were to add a `build tool` as a new platform, I think the correct rule would be to require `build` component, and forbid `flutter`, `html`, `js`, and `nativewrappers`. I agree, before we go there, there should be more checks, but we are not there yet.

matanlurey: Given barback is on its way out do we really want to spend time identifying these?

isoos: At the moment 113 packages depend directly on barback, 500+ as transitive. I think it is worth to identify it.

PR #239, opened 2018-01-22
isHosted flag for pub site

isoos:


lib/src/pkg_resolution.dart
```diff
@@ -308,6 +308,12 @@ class PkgDependency extends Object
   bool get isLatest => available == null;
   bool get isOutdated => !isLatest;
 
+  bool get isHosted =>
```
jakobr-google: Is an 'inherited' constraint always hosted?

isoos: Good point, I'm not sure about it.
I'd go with these for now, and run some checks on the pub site to see how they pan out.

PR #249, opened 2018-02-14
Expose all dartanalyzer and dartfmt problems as suggestions.

### LINKED ISSUE #222: penalties should be reflected in suggestions

isoos:
This ended up a bit longer than expected, but it closes #222 and also closes https://github.com/dart-lang/pub-dartlang-dart/issues/692. Increases the fine-grained data in the output, but the rendered parts on the pub site will be compact.

### DIFF
--- a/lib/src/fitness.dart
+++ b/lib/src/fitness.dart
@@ -26,7 +26,10 @@ class Fitness extends Object with _$FitnessSerializerMixin {
   /// The faults, penalties and failures to meet the standards.
   final double shortcoming;
 
-  Fitness(this.magnitude, this.shortcoming);
+  @JsonKey(includeIfNull: false)
+  final List<Suggestion> suggestions;
+
+  Fitness(this.magnitude, this.shortcoming, {this.suggestions});
 
   factory Fitness.fromJson(Map json) => _$FitnessFromJson(json);
 
@@ -37,6 +40,8 @@ class Fitness extends Object with _$FitnessSerializerMixin {
     final score = (magnitude - shortcoming) / magnitude;
     return max(0.0, min(1.0, score));
   }
+
+  bool get hasSuggestion => suggestions != null && suggestions.isNotEmpty;
 }
 
 Future<Fitness> calcFitness(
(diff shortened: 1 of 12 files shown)


lib/src/fitness.dart
```diff
@@ -71,29 +76,73 @@ Future<Fitness> calcFitness(
...
+    suggestions.add(new Suggestion.hint(
+      'Format `$dartFile`.',
+      'Run `dartfmt` to format `$dartFile`.',
+      file: dartFile,
+      penalty: new Penalty(amount: hintPoints.ceil()),
+    ));
+  }
+
+  String suggestionDescription(CodeProblem cp, String local) {
+    return 'Strong-mode analysis of `$dartFile` $local:\n\n'
```
jakobr-google: Is it always strong-mode, or does it depend on analyzer-options.yaml?

Since it'll eventually be the only mode, maybe we should just say 'Analysis of...', but we can leave that for a later change.

lib/src/fitness.g.dart
```diff
@@ -12,11 +12,30 @@ part of pana.health;
 
 Fitness _$FitnessFromJson(Map<String, dynamic> json) => new Fitness(
     (json['magnitude'] as num)?.toDouble(),
-    (json['shortcoming'] as num)?.toDouble());
+    (json['shortcoming'] as num)?.toDouble(),
+    suggestions: (json['suggestions'] as List)
+        ?.map((e) => e == null
```
jakobr-google: Do we want to keep `null` suggestions? Otherwise we could remove them first with a `where()`.

jakobr-google: Ah, this is generated code... Never mind!

lib/src/fitness.dart
isoos: I like to include it for now, because it raises awareness, but we may omit that after Dart 2 is finished. Added a TODO to change it later.

PR #256, opened 2018-02-27
Expose Flutter detection to clients, with better naming

isoos:


lib/src/sdk_env.dart
```diff
@@ -185,15 +186,20 @@ class PubEnvironment {
     }
   }
 
-  Future<ProcessResult> runUpgrade(String packageDir, bool isFlutter,
+  Future<bool> detectFlutterUse(String packageDir) async {
```
jakobr-google: Does this need to be async?

isoos: Doesn't need to, but I want to remove the non-async parts eventually, and it is better to start with the async API.

isoos: * non-async parts = parts where we use sync `dart:io`, while we could have used async instead.

jakobr-google: sgtm

Issue #269, opened 2018-03-16
Pana may wish to consider docs coverage in the score (or at least list as a warning)


eseidelGoogle:
Flutter for example tracks that all public members in `flutter/flutter` are documented.

`flutter analyze --flutter-repo --dartdocs` is the command we use.

Which just turns on:
http://dart-lang.github.io/linter/lints/package_api_docs.html

isoos:
I think this could be part of the code health (vs. the other option is maintenance).

Question: given a package with great documentation with 100/100 health score, if you'd remove all of the API docs, how much would you decrease its score? What would be a "fair" percentage for it?

kwalrath:
90/100? There's a tradeoff between encouraging doc comments and people gaming the system by writing unhelpful doc comments just to increase the score.

isoos:
For the first iteration, we could do the following
- Run `dartdoc`.
- After a successful `dartdoc` run, we'll get an `index.json`, that contains all the public entry points of the API. Counting them gives us the `apiCount`.
- When we run `dartanalyzer`, we can turn on the `public_member_api_docs` linter rule. The number of the reported issues gives us the `missingDocCount`.
- health (`Fitness`) needs to be adjusted with these, although there is already an "api size estimate" in the current `Fitness.magnitude`.

This does not handle empty documentation, nor does it take the method size into account, e.g. the following will be counted as one missing API entry point out of the many...
```
///
void method() {}
```

kevmoo:
FYI: `public_member_api_docs` seems to be (very) broken - https://github.com/dart-lang/linter/issues/213#issuecomment-373855503

kwalrath:
How hard should we be on packages that don't rely on API reference docs as their primary docs?

I'm thinking of angular_components, which relies on the gallery at https://dart-lang.github.io/angular_components_example rather than API docs, since you usually use the components via HTML rather than Dart.

I suppose we can detect these packages by whether they have a non-blank `documentation` field in their pubspec.

/cc @nshahan @tedsander

kwalrath:
Thinking about this some more, there's a hierarchy of doc importance:

1. README (used for API doc homepage)
2. library description (every library should have one, maybe even if the library isn't included in API docs by default)
3. class description (every important class should have one; ideally, all classes should give a little context about when you might need the class)
4. member descriptions (not always necessary; better to have no description than boilerplate)

nshahan:
[...]

PR #291, opened 2018-04-17
Check if dartdoc can run on the package.

isoos:
First iteration for #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * `ToolEnvironment.runAnalyzer` returns the text output of the process.
 
+Updates:
+
+* Check if `dartdoc` can run on the package.
+
 ## 0.10.6
 
 * Enable Dart 2 Preview in analyzer options (including non-Flutter packages).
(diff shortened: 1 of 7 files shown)


jakobr-google:
Doesn't this mean that we'll be generating the docs both in analyzer and dartdoc instances? If so, can we avoid doing the work twice? Maybe pull the dartdoc result in the analyzer client, and merge it into the analysis data?

isoos:
> Doesn't this mean that we'll be generating the docs both in analyzer and dartdoc instances? If so, can we avoid doing the work twice? Maybe pull the dartdoc result in the analyzer client, and merge it into the analysis data?

Yeah, this is likely the direction we are heading, although there are coordination issues, like: obsolete packages are not analyzed with `pana`, but are still running with `dartdoc`. Merging the two processing makes sense, but it requires more work on both pana (configurable/optional items in the analysis) and pub site (merging the two job processing).

For the immediate scope I think the `ToolEnvironment` parts should go in, and the actual running of `dartdoc` should be optional. I can split this PR or do a follow-up, whichever you prefer.

PR #306, opened 2018-05-04
Dartdoc options: retry and timeout

isoos:


lib/src/package_analyzer.dart
```diff
@@ -195,11 +199,23 @@ class PackageAnalyzer {
...
+      for (var i = 0; i <= options.dartdocRetry; i++) {
+        try {
+          final r = await _toolEnv.dartdoc(
+            pkgDir,
+            options.dartdocOutputDir,
+            validateLinks: i == 0,
+            timeout: options.dartdocTimeout,
+          );
+          dartdocSuccessful = r.wasSuccessful;
+          if (r.wasTimeout) {
```
jakobr-google: Since this is the last statement in the for block, you could also just do
```dart
if (!r.wasTimeout) {
  break;
}
```

PR #307, opened 2018-05-07
Fix deprecated imports.

isoos:


isoos:
Closing this, turns out I've upgraded dependency versions last week to 0.31.2-alpha, but the fresh upgrade does not do it yet.

PR #311, opened 2018-05-09
Bugfix: do not initialize dartdocSuccessful with a value.

isoos:
It will get the false as default, and we are reporting unsuccessful dartdoc generation in the analysis, when there was no actual run of it.


isoos:
The bug is actually a blocker (I don't have any sane way to have a workaround in pub site without changing some behaviour), therefore updating the `pana` version too, in order to have a new release ASAP.

isoos:
Re-classified as a regular bug (not changing version).

PR #314, opened 2018-05-10
Use a builder so folks can run `pub run build_runner build`

kevmoo:


kevmoo:
CC @jakemac53 @natebosch – did I do this correctly?

PR #317, opened 2018-05-16
Do not excluded libs in dartdoc processing.

isoos:
0.19 handles it:
https://github.com/dart-lang/pub-dartlang-dart/issues/1254

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -275,14 +275,16 @@ class ToolEnvironment {
     bool validateLinks: true,
     bool linkToRemote: false,
     Duration timeout,
+    List<String> excludedLibs,
   }) async {
     ProcessResult pr;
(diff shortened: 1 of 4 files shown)


lib/src/sdk_env.dart
```diff
@@ -275,14 +275,16 @@ class ToolEnvironment {
...
     bool linkToRemote: false,
     Duration timeout,
+    List<String> excludedLibs,
   }) async {
     ProcessResult pr;
     final args = [
       '--output',
       outputDir,
-      '--exclude',
-      dartdocExcludedLibraries.join(','),
```
jakobr-google: Is `dartdocExcludedLibraries` still in use? Otherwise we could remove it too.

isoos: good point, excluded.

PR #320, opened 2018-05-17
Use latest pkg:analyzer

kevmoo:


build.yaml
```diff
@@ -19,3 +20,10 @@ builders:
     builder_factories: ["buildPana"]
     build_extensions: {".dart": [".g.dart"]}
     build_to: source
+
+  version:
+    # See https://github.com/dart-lang/build/issues/1423
+    import: "../../../tool/builder.dart"
+    builder_factories: ["versionBuilder"]
+    build_extensions: {"$lib$": [".dart"]}
```
natebosch: This is sufficient because we only care (today) about whether this needs to run before other builders that require all Dart inputs... but it would be better to make it match the real build extensions

tool/src/version_builder.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: 2018

tool/src/version_builder.dart
```diff
@@ -0,0 +1,42 @@
...
+    var yaml = loadYaml(content) as Map;
+
+    var versionString = yaml['version'] as String;
+    versionString = new Version.parse(versionString).toString();
+
+    var versionFileId = new AssetId('pana', 'lib/src/version.dart');
+
+    // Write out the new asset.
+    await buildStep.writeAsString(versionFileId, '''
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: 2018

tool/src/version_builder.dart
```diff
@@ -0,0 +1,42 @@
...
+        await buildStep.readAsString(new AssetId.parse('pana|pubspec.yaml'));
+
+    var yaml = loadYaml(content) as Map;
+
+    var versionString = yaml['version'] as String;
+    versionString = new Version.parse(versionString).toString();
+
+    var versionFileId = new AssetId('pana', 'lib/src/version.dart');
+
+    // Write out the new asset.
```
natebosch: is this adding information?

tool/src/version_builder.dart
```diff
@@ -0,0 +1,42 @@
...
+
+import 'package:build/build.dart';
+import 'package:pub_semver/pub_semver.dart';
+import 'package:yaml/yaml.dart';
+
+class PackageVersionBulder extends Builder {
+  @override
+  Future build(BuildStep buildStep) async {
+    var content =
+        await buildStep.readAsString(new AssetId.parse('pana|pubspec.yaml'));
```
natebosch: I _guess_ this is fine since we don't expect it to be used outside of this package... but it seems weird to hardcode it when we have the information anyway.

tool/src/version_builder.dart
```diff
@@ -0,0 +1,42 @@
...
+
+    var versionFileId = new AssetId('pana', 'lib/src/version.dart');
+
+    // Write out the new asset.
+    await buildStep.writeAsString(versionFileId, '''
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library pana.version;
```
natebosch: why do you want a library statement?

build.yaml
kevmoo: Explain?

tool/src/version_builder.dart
kevmoo: copy-paste error.

tool/src/version_builder.dart
kevmoo: Explain?

build.yaml
natebosch: the extensions indicate you're going to writ the file `lib/.dart` because they don't match the extension on your actual builder which are `{"$lib$", ["src/version.dart"]}`. It doesn't actually hurt based on how we currently use `build_extensions` from this config, but it's still odd.

tool/src/version_builder.dart
natebosch: by hardcoding `pana` here you're writing a builder that will only ever work in the pana package. That's not the _only_ reason it only works here (the fact that its in `tool/` rather than `lib/` is the other reason), but it will mean there are more changes we'd need to make if we wanted to move this somewhere else and share it.

The way we'd write this if we wanted it to be general purpose instead of hardcoded is to check `buildStep.inputId.package` instead of `pana` and I don't see that as any more complicated than the current code so I don't see value in leaving it hardcoded.

build.yaml
kevmoo: Fixed. I wish we had better – something – here. Like a warning if this doesn't align w/ the content in code?

tool/src/version_builder.dart
kevmoo: Fixed. It's good to show best practices, me thinks.

isoos:
I think we shall rename this as "refactoring the version.dart generation" and remove the `package:analyzer` upgrade part, filing it as a separate PR, don't you think?

isoos:
Created a PR for the analyzer-only parts: https://github.com/dart-lang/pana/pull/321

On the other hand, do we want to keep the generated version for the long term? Having a single field + test for matching with pubspec.yaml (like the ones we have on pub site) would do just fine, without the need for the custom build.

kevmoo:
Happy to split this up.

Was trying to dogfood other build features. Not urgent...

kevmoo:
Glad you got it in, @isoos

PR #324, opened 2018-05-23
Support changing part of the analysis result.

isoos:


isoos:
Merging, as test break is unrelated.

PR #325, opened 2018-05-23
Fix build: re-running codegen

isoos:
Similar to the pub site, a dependency upgrade requires re-running.

### DIFF
--- a/lib/src/model.g.dart
+++ b/lib/src/model.g.dart
@@ -23,13 +23,11 @@ Summary _$SummaryFromJson(Map<String, dynamic> json) => new Summary(
         ? null
         : new PkgResolution.fromJson(
             json['pkgResolution'] as Map<String, dynamic>),
-    json['dartFiles'] == null
-        ? null
-        : new Map<String, DartFileSummary>.fromIterables(
-            (json['dartFiles'] as Map<String, dynamic>).keys,
-            (json['dartFiles'] as Map).values.map((e) => e == null
-                ? null
-                : new DartFileSummary.fromJson(e as Map<String, dynamic>))),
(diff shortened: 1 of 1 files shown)


isoos:
strange: the generated code doesn't seem to be that much different, and looks like as if it wasn't able to deserialize an enum

lib/src/model.g.dart
```diff
@@ -232,14 +229,12 @@ abstract class _$PenaltySerializerMixin {
...
-                (json['uses'] as Map).values.map((e) => e == null
-                    ? null
-                    : PlatformUse.values
-                        .singleWhere((x) => x.toString() == 'PlatformUse.$e'))),
+        (json['uses'] as Map<String, dynamic>)?.map((k, e) => new MapEntry(
+            k,
+            e == null
+                ? null
+                : PlatformUse.values
+                    .singleWhere((e) => e.toString() == 'PlatformUse.$e'))),
```
isoos: The issue **is** in the generated code: Here we have two `e` variables: one from the top-level `map`, one inside the `singleWhere`, and with that, `(e) => e.toString() == 'PlatformUse.$e')` is evaluated as false all the time.

@kevmoo: do we have a quick fix for this, or shall I downgrade the versions?

kevmoo: Gah! Downgrade. File an issue on json_serial – will fix ASAP

kevmoo: https://github.com/dart-lang/json_serializable/issues/180

kevmoo: https://github.com/dart-lang/json_serializable/pull/181

kevmoo: Publish @isoos !

isoos: Thanks, this works great!

isoos:
PTAL

PR #326, opened 2018-05-24
Fix dartdoc timeout when using pub global run.

isoos:


isoos:
merging, as build break is unrelated.

PR #331, opened 2018-05-30
Use public analyzer APIs.

scheglov:


lib/src/library_scanner.dart
```diff
@@ -110,25 +84,17 @@ class LibraryScanner {
...
 
-    var context = AnalysisEngine.instance.createAnalysisContext()
-      ..analysisOptions = options
-      ..sourceFactory = new SourceFactory(resolvers);
+    var analysisContext = new ContextBuilder()
+        .createContext(contextRoot: roots.single, sdkPath: sdkPath);
 
     return new LibraryScanner._(
-        package, packagePath, packageResolver, context, overrides);
+        package, packagePath, analysisContext.currentSession, overrides);
```
isoos: Does this mean we don't have any global-scoped cache in `package:analyzer`?

scheglov: Yes, `package:analyzer` does not cache any resolution or ASTs. Once analysis session is gone, all this data is gone too. There is a global file content cache, but it is keyed in `Expando` with the `FileContentOverlay` (and file system, but it is always the same). We create a new `FileContentOverlay` instance for each new `AnalysisContext`, so it goes away too.

isoos:
@scheglov: merge at will or let me know if I should do it

kevmoo:
Please squash merge!

PR #334, opened 2018-05-31
Disable implicit casts

isoos:
Going ahead of these changes...

### DIFF
--- a/lib/src/license.dart
+++ b/lib/src/license.dart
@@ -17,15 +17,16 @@ Future<List<LicenseFile>> detectLicensesInDir(String baseDir) async {
   final rootFiles = await new Directory(baseDir).list().toList();
   final licenseCandidates = rootFiles
       .where((fse) => fse is File)
+      .cast<File>()
       .where(_isLicenseFile)
       .take(5) // only the first 5 files are considered
       .toList();
   if (licenseCandidates.isEmpty) return [];
 
   final licenses = await Future.wait(licenseCandidates.map(
-    (FileSystemEntity fse) {
-      final relativePath = p.relative(fse.path, from: baseDir);
-      return detectLicenseInFile(fse, relativePath: relativePath);
+    (File file) {
+      final relativePath = p.relative(file.path, from: baseDir);
+      return detectLicenseInFile(file, relativePath: relativePath);
     },
   ));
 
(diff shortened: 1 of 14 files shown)


lib/src/license.dart
```diff
@@ -17,15 +17,16 @@ Future<List<LicenseFile>> detectLicensesInDir(String baseDir) async {
   final rootFiles = await new Directory(baseDir).list().toList();
   final licenseCandidates = rootFiles
       .where((fse) => fse is File)
+      .cast<File>()
```
jakobr-google: I believe the recommendation is to replace these two lines with a single:
```dart
  .whereType<File>()
```

isoos: I've changed it, but getting

````
  UnimplementedError: whereType is not yet supported
  dart:collection/list.dart 198:7               _ListBase&Object&ListMixin.whereType
  package:pana/src/license.dart 19:8            detectLicensesInDir
````

Changing it back for now.

File: lib/src/model.g.dart [hunk 2]
```diff
@@ -15,7 +15,7 @@ Summary _$SummaryFromJson(Map<String, dynamic> json) => new Summary(
     json['packageName'] as String,
     json['packageVersion'] == null
         ? null
-        : new Version.parse(json['packageVersion']),
+        : new Version.parse(json['packageVersion'] as String),
```
kevmoo: Did you make these changes by hand?

isoos: Updated the `tool/src/version_helper.dart`

kevmoo:
Ah!

isoos:
Interesting thing: the PR is passes test right now, but if I run it with preview-dart-2, the end-to-end test fail:

````
 find test -name \*_test.dart -exec dart --preview-dart-2 {} \;
````

The error message is: `Which: has different length and is missing map key 'constraint' at location ['pkgResolution']['dependencies'][1]`, which would indicate that it fails to serialize the `VersionConstraint` instance. @kevmoo: any idea what could be wrong there?

kevmoo:
No clue. Very weird!

lib/src/license.dart
kevmoo: Need to wait until preview-dart-2 is enabled everywhere...

isoos:
Found the issue: `Map<String, dynamic> get dependencies` in `Pubspec` was returning null because of the type check failing. Using wrapping solved it for now, added TODO to check back on it later.

PR #337, opened 2018-06-04
Use new pubspec_parse package

kevmoo:
Local Pubspec class is now a thin wrapper over the pkg class


isoos:
cool!

kevmoo:
FYI: this is pretty strict – so we should keep an eye out when it's deployed...

PR #338, opened 2018-06-05
Upgrade CI to dev.60 SDK

isoos:


isoos:
To fix the CI build, I've also updated this PR with the end2ends test golden data, as the `package:path` dependency did change and it is now pulling in new files.

isoos:
Removed test fixes as #341 did it already.

PR #339, opened 2018-06-07
Add identifier code to suggestions.

isoos:
Drive by https://github.com/dart-lang/pub-dartlang-dart/issues/1354

The names are arbitrary, but follow a pattern: the first part(s) can be used to group the issues and if there is an additional help page for the entire group, it can be applied with a simple prefix-matching. 

@kevmoo: do we have any preference on how to make the code-parsing easier? (E.g. `/` instead of `.` between the parts).


isoos:
There is an overlap with #338, will need to rebase after that gets merged.

kevmoo:
> do we have any preference on how to make the code-parsing easier? (E.g. / instead of . between the parts).

Not really. `.` is fine with me

PR #340, opened 2018-06-08
Better expose platform conflict reasons.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1355

Removing early platform-conflict detection, relying on a more detailed report.

### DIFF
--- a/lib/src/platform.dart
+++ b/lib/src/platform.dart
@@ -205,9 +205,12 @@ DartPlatform classifyPkgPlatform(
       libraries.keys.where((key) => libraries[key].hasConflict).toList();
   if (conflicts.isNotEmpty) {
     conflicts.sort();
-    var sample = conflicts.take(3).map((s) => '`$s`').join(', ');
+    var sample = conflicts.take(3).map((s) {
+      final components = libraries[s].components.map((s) => '`$s`').join(', ');
+      return '`$s` (components: $components)';
+    }).join(', ');
     if (conflicts.length > 3) {
-      sample = '$sample (and ${conflicts.length - 3} more).';
+      sample = '$sample and ${conflicts.length - 3} more.';
     }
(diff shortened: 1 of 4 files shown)


isoos:
CI will be red, until a related PR gets merged and rebased.

lib/src/platform.dart
```diff
@@ -232,9 +235,13 @@ DartPlatform classifyPkgPlatform(
...
       flutterConflicts.sort();
-      var sample = flutterConflicts.take(3).map((s) => '`$s`').join(', ');
+      var sample = flutterConflicts.take(3).map((s) {
+        final components =
+            libraries[s].components.map((s) => '`$s`').join(', ');
+        return '`$s` (components: $components)';
+      }).join(', ');
       if (flutterConflicts.length > 3) {
-        sample = '$sample (and ${flutterConflicts.length - 3} more).';
+        sample = '$sample and ${flutterConflicts.length - 3} more.';
```
sortie: Does this "x, y, z, and 5 more" pattern happen other places? If it does, it could be good to make a utility function for formatting the first few elements of an iterable nicely. This pattern sometimes annoys me on websites when they say "foo and 1 more." but saying the one more element would have been shorter than saying "and 1 more", that's less of a problem here with 3 elements.

isoos: It is only this place and the one a few lines above it. Yeah that "1 more" annoys me too. I was considering to do it like: display it up until 4, but when it is five, display 3 and add "and 2 more", but then it has this weird feeling: why should we display less items when it is more?

I'll do a follow-up refactoring for the utility function.

sortie: Yeah I think a good solution is something like trying to fit as many in X characters as possible, but at least 2 (or 1 if long) and the rest as "and more".

PR #342, opened 2018-06-12
Sanity test for analysis options.

isoos:


isoos:
> This makes it harder to update the analysis options?

It acts as a gateway, if we change something in the code that affects the way we generate the options file, we should be aware of it. We have similar test cases in this file, I just happened to notice that we don't have a case for the non-Flutter defaults.

> What is it that can go wrong so the wrong analysis options are used?

We analyze packages with different settings. The main difference is either generic or Flutter, and we have a few edge-cases where we rely on the developer-provided analysis_options to use some of the options from there.

PR #344, opened 2018-06-12
Compact sample building for platform conflict description.

isoos:
I've settled with displaying either three items in full, or two + "and X more" where is always X>=2.

### DIFF
--- a/lib/src/messages.dart
+++ b/lib/src/messages.dart
@@ -51,3 +51,14 @@ String runningDartanalyzerFailed(bool usesFlutter, errorMsg) {
   return 'Running `${dict.dartanalyzerShortCmd}` failed with the following output:\n\n'
       '```\n$errorMsg\n```\n';
 }
+
+/// Build a list like "A, B and 3 more".
+String buildSample(Iterable<String> items) {
+  final total = items.length;
+  final fullCount = total <= 3 ? total : 2;
+  var sample = items.take(fullCount).join(', ');
+  if (total > fullCount) {
+    sample = '$sample and ${total - fullCount} more.';
+  }
+  return sample;
+}
--- a/lib/src/platform.dart
+++ b/lib/src/platform.dart
@@ -1,5 +1,6 @@
 library pana.platform;
 
+import 'messages.dart' show buildSample;
 import 'model.dart'
     show ComponentNames, DartPlatform, PlatformNames, PlatformUse;
 import 'pubspec.dart';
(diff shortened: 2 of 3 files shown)


lib/src/utils.dart
```diff
@@ -247,3 +247,13 @@ Stream<ProcessSignal> getSignals() => Platform.isWindows
     ? ProcessSignal.sigint.watch()
     : StreamGroup
         .merge([ProcessSignal.sigterm.watch(), ProcessSignal.sigint.watch()]);
+
```
sortie: Suggestion: "utils" files tends to become the kitchen sink. How about sample.dart or a collection of presentational string APIs.

lib/src/utils.dart
```diff
@@ -247,3 +247,13 @@ Stream<ProcessSignal> getSignals() => Platform.isWindows
...
+/// Build a list like "A, B and 3 more".
+String buildSample<T>(List<T> items, String format(T item)) {
```
sortie: Use an Iterable<T> instead so it's not required to build a full list and only the needed elements can be built lazily.

The caller could be required to format the strings themselves, which lets you remove the second parameter and the genericness. For instance, buildSample(objectIterator.map((x) => x.toString()). This would work lazily and produce just the needed elements.

Maybe have an optional parameter with the amount of items to include? Perhaps a number of characters parameter, where it tries to fills as many items in the output while not producing too long a string.

Is sample the standard word for this feature?

lib/src/utils.dart
```diff
@@ -247,3 +247,13 @@ Stream<ProcessSignal> getSignals() => Platform.isWindows
...
+  final fullCount = items.length <= 3 ? items.length : 2;
+  var sample = items.take(fullCount).map(format).join(', ');
+  if (items.length > fullCount) {
+    sample = '$sample and ${items.length - fullCount} more.';
```
sortie: Is a string buffer more efficient?

isoos: It is, but the overhead of this is so little (compared to the entire process of the analysis) that I lean more towards compact code over the most efficient code. Can change if you feel strongly about it.

lib/src/utils.dart
isoos: Moving to `messages.dart`.

lib/src/utils.dart
isoos: Skipped the optional parameter for amount/length for now, but did the rest.

Re: naming: I'm not sure either, but sample seems to be close enough.

PR #346, opened 2018-06-14
Remove temp directory path from dartfmt error message.

isoos:


lib/src/sdk_env.dart
```diff
@@ -176,10 +176,9 @@ class ToolEnvironment {
         continue;
       }
 
-      throw new Exception([
-        'dartfmt on $dir/ failed with exit code ${result.exitCode}',
-        result.stderr
-      ].join('\n').toString());
+      final output = result.stderr.toString().replaceAll(packageDir, '');
```
sortie: Maybe remove the slash following packageDir as well, as "/bin/packages/stream/src/rspc/tag.dart" looks like an absolute path, but really the relative path "bin/packages/stream/src/rspc/tag.dart" is meant.

isoos: ack

PR #347, opened 2018-06-15
Fix 'null' text in platform conflict descriptions.

isoos:
The current `package:test` analysis has a "null" string in it, this fixes it.


isoos:
CI should be fixed after #345 gets merged.

PR #349, opened 2018-06-15
Preparing 0.11.4 release

isoos:


isoos:
> Ok. What was the command for publishing the new release?

`pub publish`

sortie:
I have now published pana 0.11.4.

Which pub version am I supposed to be using? I just used whatever I had in my PATH and it worked for me.

isoos:
Confirmed:
https://pub.dartlang.org/packages/pana

> Which pub version am I supposed to be using? I just used whatever I had in my PATH and it worked for me.

Any version should work (I believe even very old ones too, because the pub client APIs haven't changed for a long time).

PR #353, opened 2018-06-20
Upgrade pubspec_parse and adding end2end test.

isoos:
The `fs_shim` test was failing with the previous version of `pubspec_parse`, added it for an early warning.


isoos:
@sortie: PTAL, I needed to update code and tests to make it pass (separated in the second commit).

PR #354, opened 2018-06-20
Do not check existence of non-external URLs.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
 
 * Do not check existence of non-external URLs.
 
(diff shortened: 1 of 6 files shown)


isoos:
PTAL: I've update the PR with `UrlChecker`, which could be used by the pub site to cache the HEAD requests.

isoos:
@kevmoo: could you please review this too? It stands in the way (or will raise conflicts) of a larger, format-breaking refactor and restructuring.

PR #357, opened 2018-06-22
CI on SDK dev.64.1

isoos:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: dart
 
 dart:
-  - 2.0.0-dev.63.0
```
kevmoo: Try updating the dependency on build_runner to `^0.8.10` – should fix travis

isoos: Thanks!

PR #359, opened 2018-06-28
Increase CI SDK to dev.65.0

isoos:


lib/src/library_scanner.dart
```diff
@@ -163,6 +163,7 @@ class LibraryScanner {
     var uri = Uri.parse(libUri);
     var package = uri.pathSegments.first;
 
+    // ignore: deprecated_member_use
     var source = _session.sourceFactory.forUri2(uri);
```
isoos: @scheglov, @kevmoo: could somebody help me out here? What is the replacement for `session.sourceFactor.forUri2`?

scheglov: `uriConverter.pathToUri()`

isoos: Thanks!

PR #361, opened 2018-06-28
Update suggestion messages.

### LINKED ISSUE #360: Update messages
- [ ] s/general direction API/general direction of the API/g?
- [ ] s/you to detect bugs/you detect bugs/g
- [ ] s/earlier.Start/earlier. Start/g
- [ ] 'Fix analysis and formatting errors|hints|errors and hints

isoos:
Fixes #360.

I've eventually opted for "Fix additional 12 files with analysis or formatting issues.", as it makes it clear that it is about 12 files and not 12 issues.

### DIFF
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -258,7 +258,7 @@ Future<Maintenance> detectMaintenance(
     maintenanceSuggestions.add(new Suggestion.hint(
         SuggestionCode.analysisOptionsWeakMode,
         'Enable strong mode analysis.',
-        'Strong mode helps you to detect bugs and potential issues earlier.'
+        'Strong mode helps detect bugs and potential issues earlier. '
         'Start your `analysis_options.yaml` file with the following:\n\n'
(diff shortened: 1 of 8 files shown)


lib/src/maintenance.dart
```diff
@@ -358,20 +361,21 @@ Future<Maintenance> detectMaintenance(
...
-      final level = (errorCount > 0 || warningCount > 0)
+      final level = (bulkErrorCount > 0 || bulkWarningCount > 0)
           ? SuggestionLevel.warning
           : SuggestionLevel.hint;
       maintenanceSuggestions.add(
         new Suggestion(
           SuggestionCode.bulk,
           level,
-          'Fix analysis and formatting issues.',
+          'Fix additional ${restSuggestions.length} files with analysis or formatting issues.',
```
sortie: Is restSuggestions.length always more than 1? Should pluralize be used here?

isoos: Yes, it is always 3+, because it starts to work when the total is >= 5, and 2 of those will be taken out.

PR #364, opened 2018-07-03
Reduce the time-invariability of the end2end tests.

isoos:


test/end2end_test.dart
```diff
@@ -58,28 +58,29 @@ void main() {
           fail('Set `_regenerateGoldens` to `false` to run tests.');
         }
 
+        void removeDependencyDetails(Map map) {
+          if (map.containsKey('pkgResolution') &&
+              (map['pkgResolution'] as Map).containsKey('dependencies')) {
```
sortie: What happens if map['pkgResolution'] exists but is not a map? Is the data always well-formed?

isoos: In that case the test will fail, as in the golden file it is a map. Which would be a good thing, because that is our alert that something changed underneath and the previous assumption are changing too.

PR #367, opened 2018-07-04
Fix issue with multiple context roots.

### LINKED ISSUE #366: Bug in analyzer or in our processing?
````
SEVERE     Could not create LibraryScanner
           Bad state: Expected exactly one context root, got [Instance of 'ContextRootImpl', Instance of 'ContextRootImpl', Instance of 'ContextRootImpl', Instance of 'ContextRootImpl', Instance of 'ContextRootImpl', Instance of 'ContextRootImpl']
           #0      new LibraryScanner (package:pana/src/library_scanner.dart:85:7)
           #1      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:222:30)
           <asynchronous suspension>
           #2      PackageAnalyzer.inspectPackage.<anonymous closure> (package:pana/src/package_analyzer.dart:85:22)
           <asynchronous suspension>
           #3      _rootRun (dart:async/zone.dart:1126:13)
           #4      _CustomZone.run (dart:async/zone.dart:1023:19)
           #5      _runZoned (dart:async/zone.dart:1518:10)
           #6      runZoned (dart:async/zone.dart:1465:12)
           #7      withLogger (package:pana/src/logging.dart:13:61)
           #8      PackageAnalyzer.inspectPackage (package:pana/src/package_analyzer.dart:71:12)
           <asynchronous suspension>
           #9      main (file:///Users/isoos/work/google/pub/pana/bin/pana.dart:147:34)
           <asynchronous suspension>
           #10     _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:277:32)
           #11     _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:165:12)
````

The related code is:
````
    var contextLocator = new ContextLocator();
    var roots = contextLocator.locateRoots(includedPaths: [packagePath]);
    if (roots.length != 1) {
      throw new StateError('Expected exactly one context root, got $roots');
    }
````

### PR DESCRIPTION
**Pull request by isoos**
Fixes #366.

### CODE REVIEW

File: lib/src/library_scanner.dart [hunk 1]
```diff
@@ -81,12 +81,22 @@ class LibraryScanner {
...
     var contextLocator = new ContextLocator();
     var roots = contextLocator.locateRoots(includedPaths: [packagePath]);
-    if (roots.length != 1) {
-      throw new StateError('Expected exactly one context root, got $roots');
+    var root = roots.firstWhere(
+        (r) => r.packagesFile.parent.path == packagePath,
+        orElse: () => null);
+    if (root == null) {
+      log.warning(
+          'No context root on the default path, selecting the one with the most files.');
```
sortie: This seems a little magic to me. Does the user get this warning on the package page? Does this happen in practice for packages (which? Is that what we're finding out here? The package name, package path, and other information isn't mentioned in the log message). Is the "most files" fallback needed? It may produce the wrong result, so it could be better to have packages strictly specify the right path?

test/end2end_test.dart
```diff
@@ -102,6 +102,7 @@ void main() {
     }, timeout: const Timeout.factor(2));
   }
 
+  _verifyPackage('dartdoc-0.20.0.json', 'dartdoc', '0.20.0');
```
sortie: Unrelated?

isoos: Not unrelated, I'm adding the test because `dartdoc 0.20.0` (and earlier versions) were failing without this change, and the PR unblocks its analysis. If something happens again underneath, we should have better chances to notice it earlier.

lib/src/library_scanner.dart
isoos: This is for us to get notified through the logs. It is a non-blocking situation at this point, because we may have valid context roots to work with, but it does need investigation from our side.

PR #370, opened 2018-07-11
 Updated report on dartanalyzer suggestions.

isoos:
- `dartanalyzer` suggestions become one per file, with the summary issue count included
- `maintenance` needed to be adjusted to handle the now-reduced suggestions per file


isoos:
> Is this preparing a release?

Yes, it does. Shall we merge or would you rather split it?

sortie:
sgtm

sortie:
That is, go ahead and merge

PR #381, opened 2018-07-31
Use flutter format instead of dartfmt for Flutter packages.

isoos:
- `flutter format` emits different output and doesn't communicate with exit codes
- kept the same structure of per-directory scanning
- fixed a small bug with the non-Flutter code path where an empty `bin` directory would have ignored the `lib` entirely.
- https://github.com/dart-lang/pub-dartlang-dart/issues/1484

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -150,37 +150,56 @@ class ToolEnvironment {
     }
   }
 
-  Future<List<String>> filesNeedingFormat(String packageDir) async {
+  Future<List<String>> filesNeedingFormat(
+      String packageDir, bool usesFlutter) async {
     final dirs = await listFocusDirs(packageDir);
     if (dirs.isEmpty) {
       return const [];
     }
-    final files = <String>[];
+    final files = new Set<String>();
     for (final dir in dirs) {
-      var result = await runProc(
-        _dartfmtCmd,
-        ['--dry-run', '--set-exit-if-changed', p.join(packageDir, dir)],
-        environment: _environment,
-      );
-      if (result.exitCode == 0) {
-        return const [];
-      }
+      final fullPath = p.join(packageDir, dir);
 
-      final lines = LineSplitter.split(result.stdout as String)
-          .map((file) => p.join(dir, file))
-          .toList();
-      if (result.exitCode == 1) {
-        assert(lines.isNotEmpty);
+      if (usesFlutter) {
+        final result = await runProc(
+          _flutterCmd,
+          ['format', fullPath],
+          workingDirectory: packageDir,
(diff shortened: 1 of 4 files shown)


lib/src/sdk_env.dart
```diff
@@ -150,37 +150,56 @@ class ToolEnvironment {
...
+          workingDirectory: packageDir,
+          environment: _environment,
+        );
+
+        final lines = LineSplitter.split(result.stdout as String)
+            .where((line) =>
+                line.startsWith('Formatted ') && line.endsWith('.dart'))
+            .map((line) => line.substring(10).trim())
+            .map((file) => p.join(dir, file))
+            .toList();
```
sortie: Should we talk to the flutter people to get --dry-run and --set-exit-if-changed supported?

lib/src/sdk_env.dart
```diff
@@ -150,37 +150,56 @@ class ToolEnvironment {
...
+        final result = await runProc(
+          _flutterCmd,
+          ['format', fullPath],
+          workingDirectory: packageDir,
+          environment: _environment,
+        );
+
+        final lines = LineSplitter.split(result.stdout as String)
+            .where((line) =>
+                line.startsWith('Formatted ') && line.endsWith('.dart'))
```
sortie: I dislike parsing human readable output, it tends to be fragile. Should we talk to the flutter people?

test/end2end/dartdoc-0.20.0.json
```diff
@@ -1322,6 +1322,17 @@
             "amount": 1608,
             "fraction": 4000
           }
+        },
+        {
+          "code": "dartfmt.warning",
+          "level": "hint",
+          "title": "Format `lib/src/model.dart`.",
+          "description": "Run `dartfmt` to format `lib/src/model.dart`.",
```
sortie: Do we tell people to 'flutter format' if flutter?

lib/src/sdk_env.dart
isoos: +1

lib/src/sdk_env.dart
isoos: The output of `dartfmt` is not that great either, but it emits only the filename on each line.

test/end2end/dartdoc-0.20.0.json
isoos: Yes.

isoos:
> How does flutter format format differently?

It does not format differently, but it is tied to the Dart VM version included in Flutter SDK. Prior to this PR, we were using only our Dart SDK, and the difference in versions could result difference in the formatting, and caused confusion in the analysis results.

PR #382, opened 2018-08-02
misc: upgrade to the latest source_gen, build_runner, and json_serial

kevmoo:


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: pana
 description: Evaluate the health and quality of a Dart package
-version: 0.11.8
+version: 0.11.9-dev
```
isoos: I don't think we'll have 0.11.9, I'm targeting 0.12.0 with breaking changes.

kevmoo: Ack. Feel free to run over it...

PR #389, opened 2018-08-22
Clarify how to call pana

mit-mit:


isoos:
You'll need to modify the README and the tests to make them pass too. (Yeah, it is a bit paranoid)
I can do it later today if you are busy with other stuff..

bin/pana.dart
```diff
@@ -44,8 +44,8 @@ void _printHelp({String errorMessage}) {
     print(red.wrap(errorMessage));
     print('');
   }
-  print('''Usage: pana [<options>] <package> [<version>]
-       pana [<options>] --source path <directory>
+  print('''pana [<options>] --source path <absolute directory path>
```
sortie: Can we implement relative paths, e.g. by normalizing them if needed?

isoos: Agreed: https://github.com/dart-lang/pana/pull/392

isoos:
Closing PR, https://github.com/dart-lang/pana/pull/392 will handle it better.

sortie:
I think we should still do the --help string improvement, it's a good clarification that it's on pub.dartlang.org by default.

Issue #386, opened 2018-08-10
Homepage and Documentation issues and suggestions don't make sense.


nshahan:
I'm seeing messages like: 

Homepage is not helpful.
Update the homepage property: create a website about the package or use the source repository URL.

Documentation URL is not helpful.
Update the documentation property: create a website about the package or remove it.

For packages that supply homepage and documentation links. 

https://pub.dartlang.org/packages/angular#-analysis-tab-
https://pub.dartlang.org/packages/angular_components#-analysis-tab-

isoos:
Ah, because `angular` points to `webdev.dartlang.org`, while we consider `*.dartlang.org` as not helpful. We'll need to make an exception with angular there.

PR #393, opened 2018-08-23
Support internal packages that won't get homepage/documentation warning.

isoos:
https://github.com/dart-lang/pana/issues/386

The list of packages/patterns will be provided by the pub site.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -100,6 +100,7 @@ enum UrlStatus {
 
 class UrlChecker {
   final _internalHosts = new Set<Pattern>();
+  final _internalPackages = new Set<Pattern>();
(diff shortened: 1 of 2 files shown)


lib/src/download_utils.dart
```diff
@@ -166,7 +167,7 @@ class UrlChecker {
     return false;
   }
 
-  Future<UrlStatus> checkStatus(String url) async {
+  Future<UrlStatus> checkStatus(String url, {String packageName}) async {
```
sortie: Alternatively this pattern logic does not need to exist in pana, but could accept a bool 'isInternal' that could be provided by the pub site code.

isoos: Yeah, it could become part of `InspectOptions`.

Issue #390, opened 2018-08-22
Better error message when passing a local path


mit-mit:
pana requires an absolute path when using the `--source path` argument. If you pass a local path, we give a pretty bad error message:

```
$ pub global run pana --source path ./foobar --no-warning
INFO       Running `/Users/mit/homebrew/Cellar/dart/2.0.0/libexec/bin/dart --version`...
INFO       Running `flutter --version --machine`...
INFO       Parsing pubspec.yaml...
INFO       Running `/Users/mit/homebrew/Cellar/dart/2.0.0/libexec/bin/dartfmt --dry-run --set-exit-if-changed ./foobar/lib`...
INFO       Running `flutter packages pub upgrade --verbosity io --no-precompile`...
WARNING    Weird: `foobar` at `foobar/lib/`.
WARNING    No context root on the default path, selecting the one with the most files.
INFO       Scanning direct dependencies...
SEVERE     Error scanning direct libraries
           Invalid argument(s): Only absolute paths are supported: ./foobar/lib/foobar.dart
           #0      AnalysisDriver._throwIfNotAbsolutePath (package:analyzer/src/dart/analysis/driver.dart:1755:7)
           #1      AnalysisDriver.getUnitElement (package:analyzer/src/dart/analysis/driver.dart:832:5)
           #2      AnalysisSessionImpl.getUnitElement (package:analyzer/src/dart/analysis/session.dart:141:20)
           #3      LibraryScanner._getLibraryElement (package:pana/src/library_scanner.dart:254:38)
           <asynchronous suspension>
```

PR #394, opened 2018-08-23
Update CLI help message.

isoos:
https://github.com/dart-lang/pana/issues/390

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Usage: pana [<options>] <published package name> [<version>]
 Options:
       --flutter-sdk     The directory of the Flutter SDK.
   -j, --json            Output log items as JSON.
-  -s, --source          The source where the package is located (hosted on http://pub.dartlang.org, or local directory path).
+  -s, --source          The source where the package is located (hosted on https://pub.dartlang.org, or local directory path).
                         [hosted (default), path]
   
       --hosted-url      The server that hosts <package>.
--- a/README.md
+++ b/README.md
@@ -22,13 +22,13 @@ Used by the [Dart Package site](https://pub.dartlang.org/).
 You can specify either a package (+ version) or a local directory to analyze:
 
 ```
-Usage: pana [<options>] <package> [<version>]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,13 +22,13 @@ Used by the [Dart Package site](https://pub.dartlang.org/).
...
+       pana [<options>] --source path <local directory>
 
 Options:
       --flutter-sdk     The directory of the Flutter SDK.
   -j, --json            Output log items as JSON.
-  -s, --source          The source used to find the package.
+  -s, --source          The source where the package is located (hosted on http://pub.dartlang.org, or local directory path).
                         [hosted (default), path]
   
       --hosted-url      The server that hosts <package>.
```
sortie: Possibly mention that this overrides the default http://pub.dartlang.org.

sortie: And its dependencies?

README.md
```diff
@@ -22,13 +22,13 @@ Used by the [Dart Package site](https://pub.dartlang.org/).
...
-Usage: pana [<options>] <package> [<version>]
-       pana [<options>] --source path <directory>
+Usage: pana [<options>] <published package name> [<version>]
+       pana [<options>] --source path <local directory>
 
 Options:
       --flutter-sdk     The directory of the Flutter SDK.
   -j, --json            Output log items as JSON.
-  -s, --source          The source used to find the package.
+  -s, --source          The source where the package is located (hosted on http://pub.dartlang.org, or local directory path).
```
sortie: httpS:// please.

README.md
isoos: I think the line below is a good indication for that (defaults to https://pub.dartlang.org, and if you set to any other value, then you are on your own).

isoos: I'm not sure, it will be up to the `pub` client to download them. If the `pubspec.yaml` is set up properly, they should be able to fetch them. We had a developer in the community who was using this, and AFAICT they got it to work for their case.

PR #405, opened 2018-09-13
Increased dartfmt timeout to 5 minutes.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1605

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -20,6 +20,8 @@ import 'version.dart';
 
 final _logger = new Logger('pana.env');
 
+const _dartfmtTimeout = const Duration(minutes: 5);
+
 class ToolEnvironment {
   final String dartSdkDir;
   final String pubCacheDir;
@@ -166,6 +168,7 @@ class ToolEnvironment {
           ['format', fullPath],
           workingDirectory: packageDir,
           environment: _environment,
+          timeout: _dartfmtTimeout,
         );
 
         final lines = LineSplitter.split(result.stdout as String)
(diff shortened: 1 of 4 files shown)


lib/src/sdk_env.dart
```diff
@@ -20,6 +20,8 @@ import 'version.dart';
 
 final _logger = new Logger('pana.env');
 
+const _dartfmtTimeout = const Duration(minutes: 5);
```
sortie: Is it reasonable for a package to take two minutes to dart format, or five minutes? It seems a bit extreme. Did you test that the package in question does complete within 5 minutes? If we do this change, we indirectly encourage more packages to take longer by not failing. It'd be interesting to look at the package and see if it's any particular file or coding style that slows it down.

But this is fine and solves the problem, probably.

isoos: In that case I think it is the size of the files, e.g. one of the generated files in that package is 284kb: https://github.com/a14n/dart-google-maps/blob/master/lib/src/google_maps_src.g.dart

I see these timeouts as protection against a sticky, but in general transient issues, e.g. network outage, disk IO or process preempted for whatever reason... As long as we accept the uploaded files, we shall try our best to provide info about them. Maybe we could record the total amount of time for the analysis, and give feedback and/or maintenance penalty on the too large ones? But then: what is too large?

sortie: We're spending computing resources / money to analyze these packages. Too large would be when we think it's unreasonable to spend that much resources on a single package. I think this is fine. It would be interesting to have some statistics on the distribution on how long it takes to analyze packages.

isoos: I'll land this, and also add time tracking stats for the analysis (and parts of the analysis). We can draft better policy when we have the data :).

PR #406, opened 2018-09-14
Store basic stats in the analysis summary.

isoos:


test/end2end/dartdoc-0.20.3.json
```diff
@@ -1293,5 +1296,11 @@
       "isFormatted": true,
       "codeProblems": []
     }
+  },
+  "stats": {
+    "analyzeProcessElapsed": 8287,
+    "formatProcessElapsed": 1454,
+    "resolveProcessElapsed": 9938,
+    "totalElapsed": 33722
```
sortie: These aren't deterministic. Instead of having these values change needlessly when on every commit updating these tests, you should normalize then with mock values when writing out these .json files.

isoos: Do you have a normalization in mind? I was thinking of rounding the numbers to the second (to keep parts of the proprotions around), but that will have some movement in it too. Setting them to an arbitrary number would be my last resort, because we would loose a useful indicator (e.g. if the proportions shift too much).

sortie: I'm thinking some predefined numbers like 1234, they're just there to show the generation happened successfully. This is not a benchmark system, this is a deterministic unit testing system. We can build some benchmarks / runtime monitoring of production using the statistics made here. (Ideally even both)

PR #411, opened 2018-09-20
Upgrade CI SDK version.

isoos:


isoos:
I'll update the version string override in a separate PR, as it requires a bit of code change too.

PR #414, opened 2018-09-24
Track and expose health scoring for analyzer suggestions.

isoos:
This makes the report on health scoring consistent with maintenance scores:
- analyzer issues are calculated on a per-file base, starting from the one with most penalties
- bulk issues have the aggregate numbers

We may also expose the exact rule, but it may not be necessary anymore.

@sortie: I've included the version bump, please publish after merging.

### DIFF
--- a/lib/src/health.dart
+++ b/lib/src/health.dart
@@ -4,6 +4,8 @@
 
 library pana.health;
 
+import 'dart:math' as math;
+
 import 'package:meta/meta.dart';
 
 import 'messages.dart' as messages;
@@ -100,9 +102,10 @@ Health calcHealth({
   });
   var remainingScore = 100.0;
   for (final s in analyzerSuggestions) {
-    final score = (100.0 * remainingScore * (1.0 - s.percent)).round() / 100.0;
-    remainingScore *= s.percent;
-    suggestions.add(s.suggestion.change(score: score));
+    var penalty = (100.0 * remainingScore * (1.0 - s.percent)).round() / 100.0;
+    penalty = math.min(penalty, remainingScore);
+    remainingScore = math.max(0.0, remainingScore - penalty);
+    suggestions.add(s.suggestion.change(score: penalty));
   }
 
   suggestions.sort();
--- a/lib/src/health.dart
+++ b/lib/src/health.dart
@@ -57,7 +57,7 @@ Health calcHealth({
     }
   }
 
-  final analyzerSuggestions = <_Suggestion>[];
+  final analyzerSuggestions = <_SuggestionWithPercent>[];
 
   final reportedFiles = analyzerItems.map((cp) => cp.file).toSet();
   for (final path in reportedFiles) {
@@ -89,7 +89,7 @@ Health calcHealth({
         'Analysis of `$path` $failedWith $issueCounts$including:$issueList',
         file: path,
       );
-      analyzerSuggestions.add(new _Suggestion(suggestion,
+      analyzerSuggestions.add(new _SuggestionWithPercent(suggestion,
           calculateBaseHealth(errorCount, warningCount, hintCount)));
     }
   }
@@ -99,7 +99,7 @@ Health calcHealth({
     return -a.percent.compareTo(b.percent);
   });
   var remainingScore = 100.0;
-  for (var s in analyzerSuggestions) {
+  for (final s in analyzerSuggestions) {
     final score = (100.0 * remainingScore * (1.0 - s.percent)).round() / 100.0;
     remainingScore *= s.percent;
     suggestions.add(s.suggestion.change(score: score));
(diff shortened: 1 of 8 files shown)


lib/src/health.dart
```diff
@@ -161,3 +181,10 @@ List<Suggestion> _compact(
   suggestions.sort();
   return suggestions;
 }
+
+class _Suggestion {
```
sortie: The distinction between Suggestion and _Suggestion is confusing. _SuggestionWithPercent? Can we avoid this class or make it more like a special purpose class?

lib/src/health.dart
```diff
@@ -80,15 +82,28 @@ Health calcHealth({
...
+          calculateBaseHealth(errorCount, warningCount, hintCount)));
     }
   }
+  analyzerSuggestions.sort((a, b) {
+    final x = a.suggestion.compareTo(b.suggestion);
+    if (x != 0) return x;
+    return -a.percent.compareTo(b.percent);
+  });
+  var remainingScore = 100.0;
+  for (var s in analyzerSuggestions) {
```
sortie: 's' isn't changed, make it final

lib/src/health.dart
```diff
@@ -57,6 +57,8 @@ Health calcHealth({
     }
   }
 
+  final analyzerSuggestions = <_Suggestion>[];
```
sortie: Instead of a one-off special purpose class, you could have a map from the Suggestion to the percentage?

lib/src/health.dart
```diff
@@ -80,15 +82,28 @@ Health calcHealth({
...
     }
   }
+  analyzerSuggestions.sort((a, b) {
+    final x = a.suggestion.compareTo(b.suggestion);
+    if (x != 0) return x;
+    return -a.percent.compareTo(b.percent);
+  });
+  var remainingScore = 100.0;
+  for (var s in analyzerSuggestions) {
+    final score = (100.0 * remainingScore * (1.0 - s.percent)).round() / 100.0;
```
sortie: Does this introduce rounding errors?

lib/src/health.dart
```diff
@@ -80,15 +82,28 @@ Health calcHealth({
...
+    remainingScore *= s.percent;
+    suggestions.add(s.suggestion.change(score: score));
```
sortie: This seems to be the remaining score. Did you mean to add the score penalty of this suggestion, i.e. the difference between the old remainingScore and the new remainingScore?

lib/src/model.dart
```diff
@@ -16,6 +16,11 @@ import 'utils.dart' show toRelativePath;
 
 part 'model.g.dart';
 
+/// NOTE: In case this changes, update README.md
```
sortie: s/this/these/

lib/src/model.dart
```diff
@@ -282,6 +287,11 @@ class Suggestion extends Object
 
   @override
   String toString() => 'Sugestion: $level - $description';
+
+  Suggestion change({double score}) {
```
sortie: Why not support changing all the other fields?

lib/src/model.dart
```diff
@@ -664,15 +674,21 @@ class Health extends Object with _$HealthSerializerMixin {
...
-        math.pow(0.995, analyzerHintCount);
-    // TODO: document why and how platform conflict influence the score
+    var score = calculateBaseHealth(
+        analyzerErrorCount, analyzerWarningCount, analyzerHintCount);
     score -= 0.25 * platformConflictCount;
     return math.max(0.0, score);
   }
 }
 
+double calculateBaseHealth(
```
sortie: Dartdoc?

Write a couple unit tests for this function specifically?

lib/src/health.dart
isoos: We'll still need to sort and process them in order, I think a Map would make that slightly more complex.

lib/src/health.dart
isoos: refactoring a bit to make it clear

lib/src/health.dart
isoos: It does a bit, but with the refactored version it shouldn't be a problem anymore (e.g. values won't add up to more than 100.0).

isoos:
A few notes before merging, because there is no good way of calculating the score penalties per issues. Let's explore a package with two issues in two files:
- file-A with 1 error: -25
- file-B with 1 warning: -5
- if both are present the penalty is -28.75, health score is 71.25
- if one gets fixed, the score increase is either 23.75 (health score = 95) or 3.75 (health score=75).

Right now we will display file-A with -25 and file-B with -3.75, which is 50% correct. However, we can't use neither the higher, nor the lower values, as that way the sum of the numbers won't add up to the health score.

I think this is the better compromise between the alternatives, and for either case we'll still need to put a disclaimer on the pub site.

isoos:
After a bit of back-and-forth, I think #417 would be better.

PR #417, opened 2018-09-25
Estimate health score penalties.

isoos:
An alternative to https://github.com/dart-lang/pana/pull/414
The tradeoff here is:
- the impact of the score is not adjusted (the calculation is simpler),
- the sum of the individual penalties is larger than the overall penalty,
- there is no dependency on the order of the suggestions, it is symmetrical, and does not change if 1 out of N issues get fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.12.4
+
+* Documented how scoring works.
+
+* Estimate health score penalties.
+
 ## 0.12.3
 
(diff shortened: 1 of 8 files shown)


isoos:
> As long as it's explained.

I was thinking the following:
- it will be displayed like the rest (e.g. -25 score), but with an additional `(*)` asterisk, and 
- in the bottom of the block, it will be explained how scores are calculated, and that score is the maximum impact of that issue.

sortie:
Published

PR #423, opened 2018-10-05
Penalty for description using too many non-ASCII characters.

isoos:
There is a small, but growing portion of packages that use non-English description (and readme) to describe what the package does (e.g. [work](https://pub.dartlang.org/packages/work) or [tencent_cos](https://pub.dartlang.org/packages/tencent_cos)). We shall encourage using English, and this is a good proxy for it.

### DIFF
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -367,6 +367,16 @@ Future<Maintenance> detectMaintenance(
         score: 10.0));
   }
 
+  // Checking the non-English characters in the description
+  if (nonAsciiRuneRatio(description) > 0.1) {
+    maintenanceSuggestions.add(new Suggestion.warning(
+        SuggestionCode.pubspecDescriptionAsciiOnly,
+        'The description contains too many non-ASCII characters.',
+        'The primary audience of this site will speak and use English, please '
+        'use fewer foreign characters in the description.',
+        score: 20.0));
+  }
+
   if (unconstrainedDeps != null && unconstrainedDeps.isNotEmpty) {
     final count = unconstrainedDeps.length;
     final pluralized = count == 1 ? '1 dependency' : '$count dependencies';
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -372,8 +372,8 @@ Future<Maintenance> detectMaintenance(
     maintenanceSuggestions.add(new Suggestion.warning(
         SuggestionCode.pubspecDescriptionAsciiOnly,
         'The description contains too many non-ASCII characters.',
-        'The primary audience of this site will speak and use English, please '
-        'use fewer foreign characters in the description.',
+        "The site uses English as it's primary language. Please use a "
+        'description that primarily contains characters used when writing English.',
         score: 20.0));
   }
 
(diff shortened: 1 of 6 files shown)


lib/src/utils.dart
```diff
@@ -246,3 +246,24 @@ Stream<ProcessSignal> getSignals() => Platform.isWindows
...
+/// Returns the ratio of non-ASCII runes (Unicode characters) in a given text:
+/// (number of runes that are non-ASCII) / (total number of character runes).
+///
+/// The return value is between [0.0 - 1.0].
+double nonAsciiRuneRatio(String text) {
+  if (text == null || text.isEmpty) {
+    return 0.0;
+  }
+  // 160 - non-breaking space
+  final total = text.runes.where((r) => r > 32 && r != 160).length;
```
sortie: ASCII is from 32 to 128. 128 through 255 is not ASCII.

lib/src/utils.dart
```diff
@@ -246,3 +246,24 @@ Stream<ProcessSignal> getSignals() => Platform.isWindows
...
+  if (total == 0) {
+    return 0.0;
+  }
+  final nonAscii = text.runes.where((r) => r >= 256).length;
+  if (nonAscii > total) {
+    // should not happen
+    return 1.0;
```
sortie: I don't see how this is possible, remove this condition?

lib/src/maintenance.dart
```diff
@@ -367,6 +367,16 @@ Future<Maintenance> detectMaintenance(
...
   }
 
+  // Checking the non-English characters in the description
+  if (nonAsciiRuneRatio(description) > 0.1) {
+    maintenanceSuggestions.add(new Suggestion.warning(
+        SuggestionCode.pubspecDescriptionAsciiOnly,
+        'The description contains too many non-ASCII characters.',
+        'The primary audience of this site will speak and use English, please '
+        'use fewer foreign characters in the description.',
+        score: 20.0));
```
sortie: Code looks good to me.

@mit-mit Is this the semantics and language that we want?

mit-mit: Let's change "The primary audience of this site will speak and use English, please use fewer foreign characters in the description" to "The site uses English as it's primary language. Please use a description that primarily contains characters used when writing English."

lrhn:
This approach is using charset as a proxy for recognizing non-English languages. 

That may seem too biased against non-"western" languages. I'm fairly sure pure Danish would pass this filter (we only have three letters that are non-ASCII), so would most German and many other similar languages covered by the ISO-8859-1 (Latin-1) or Windows 1252 charsets (Afrikaans, Basque, Catalan, Danish, Dutch, English, Faeroese, Finnish, French, Galician, German, Icelandic, Irish, Italian, Norwegian, Portuguese, Spanish and Swedish are listed as languages of Latin-1).

I'd rather run the text against an English dictionary than rely on charset only. It might not seem fair to people hit by the filter if other languages can easily bypass it.
(Or maybe I'm over-worrying :)

isoos:
> I'd rather run the text against an English dictionary than rely on charset only. It might not seem fair to people hit by the filter if other languages can easily bypass it.

I think the character-range filter is a cheap pre-filter before we could apply a dictionary:
- if the characters are not in the range, it is not even worth it to check the dictionary,
- if the characters are in the range, we could check the dictionary.

Using a dictionary could also introduce a new set of issues, for example: a long list of rarely used tech names will be unlikely in it, and we could miss-classify a text based on it.

I'd rather apply this filter now, as it solves a specific issue, and worry about the rest later if they became an issue. Even if they become, the character range check would be a good thing to have.

PR #425, opened 2018-10-12
Bulk processing and comparison.

isoos:


bin/bulk.dart
```diff
@@ -0,0 +1,260 @@
...
+  @override
+  String get description => 'Analyzes multiple packages';
+
+  AnalyzeCommand() {
+    argParser
+      ..addOption('concurrency',
+          defaultsTo: '1', help: 'The concurrent execution of the analyis.')
+      ..addOption('dart-sdk', help: 'The directory of the Dart SDK.')
+      ..addOption('flutter-sdk', help: 'The directory of the Flutter SDK.')
+      ..addOption('input', help: 'The input file with the list of packages.')
```
sortie: I don't think --input is the best option name (all options are inputs after all), that option would be better if you supplied the packages yourself. How about --packages-from-file, --packages-list, --packages, something?

bin/bulk.dart
```diff
@@ -0,0 +1,260 @@
...
+      ..addOption('output',
+          help: 'The output directory to store the per-package results.');
+  }
+
+  @override
+  Future run() async {
+    final concurrency = int.tryParse(argResults['concurrency'] as String) ?? 1;
```
sortie: This doesn't throw on malformed numbers. I suggest int.parse(argResults['concurrency'] ?? "1").

bin/bulk.dart
```diff
@@ -0,0 +1,260 @@
...
+
+    final packages = new Set<String>();
+    packages.addAll(argResults.rest);
+
+    final String inputFileName = argResults['input'];
+    if (inputFileName != null) {
+      final inputFile = new File(inputFileName);
+      if (!(await inputFile.exists())) {
+        print('Input file $inputFileName does not exists.');
+        exit(-1);
```
sortie: exit(255). exit -1 doesn't make sense as program return codes are from 0 to 255 (they are modulo 256). The Dart VM uses exit code 255 for unhandled exceptions and other critical problems. Personally I encourage using exit code 1 when the program handles errors properly, like here, but the below reading of the files will exit 255 if the file doesn't have permissions set.

bin/bulk.dart
```diff
@@ -0,0 +1,260 @@
...
+      }
+      final lines = await inputFile.readAsLines();
+      packages.addAll(lines.where((l) => l.isNotEmpty));
+    }
+
+    if (packages.isEmpty) {
+      print('No package to analyze.');
+      exit(-1);
```
sortie: exit(1). Or possibly exit(0), this is not bad condition. Could we proceed through the below code as a no-operation and provide the normal output with information about zero packages?

bin/bulk.dart
```diff
@@ -0,0 +1,260 @@
...
+
+  @override
+  Future run() async {
+    final concurrency = int.tryParse(argResults['concurrency'] as String) ?? 1;
+    final pool = new Pool(concurrency);
+
+    final packages = new Set<String>();
+    packages.addAll(argResults.rest);
+
+    final String inputFileName = argResults['input'];
```
sortie: Is there a supported, documented & easy method for getting a list of all packages that is easy to pass here?

bin/bulk.dart
isoos: `--packages-list` seems to be the better choice for me.

PR #428, opened 2018-10-25
Bulk: repeated run should analyze only the unsuccessful ones.

isoos:


bin/bulk.dart
```diff
@@ -39,13 +39,16 @@ class AnalyzeCommand extends Command {
       ..addOption('packages-list',
           help: 'The input file with the list of packages.')
       ..addOption('output',
-          help: 'The output directory to store the per-package results.');
+          help: 'The output directory to store the per-package results.')
+      ..addFlag('force',
+          help: 'Re-do analysis when the output file already exists.');
```
sortie: 'for packages where the output file already exists'

Issue #418, opened 2018-09-26
Is Pana flutter only?


shamblett:
Its handy to check my packages with pana before publishing, however if I try this I get 

WARNING    Unable to detect Flutter version.
           ProcessException: The system cannot find the file specified.

Seems as though you need a flutter installed, the package is marked as Flutter,Other so I assumed it would work ina VM environment but it seems not to.

isoos:
@shamblett: It does not require Flutter for analysing non-Flutter packages, although the environment scan will try to detect it and it will log its failure. But if the package does not require Flutter then the analysis should go on.

When I run pana, I get this:

```
INFO       Running `flutter --version --machine`...
WARNING    Unable to detect Flutter version.
           ProcessException: No such file or directory
             Command: flutter --version --machine
           #0      runProc (package:pana/src/utils.dart:41:17)
           <asynchronous suspension>
           #1      ToolEnvironment.getFlutterVersion (package:pana/src/sdk_env.dart:230:15)
           <asynchronous suspension>
           #2      ToolEnvironment._init (package:pana/src/sdk_env.dart:60:31)
           <asynchronous suspension>
           #3      ToolEnvironment.create (package:pana/src/sdk_env.dart:115:19)
           <asynchronous suspension>
           #4      PackageAnalyzer.create (package:pana/src/package_analyzer.dart:62:54)
           <asynchronous suspension>
           #5      main (file:///Users/isoos/work/google/pub/pana/bin/pana.dart:144:46)
           <asynchronous suspension>
           #6      _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:287:32)
           #7      _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:171:12)
INFO       Downloading package [xxx] latest
```

As you can see, the warning does not block the analysis, and it goes on to download the package. If the analysis is blocked for you, there should be another reason for it, could you please share the full console output?

> the package is marked as Flutter,Other so I assumed it would work ina VM environment but it seems not to.

In this case the "Other" category means that it could be running in a console application, which may be a server, a regular app, or part of a code generator toolchain. Unfortunately we don't distinguish between those yet, hence the "other" category.

shamblett:
Ok, I think were getting there, when I run 

pana --source path .

for my package on linux its all Ok, on Windows however I get this 

INFO       Parsing pubspec.yaml...
[...]

PR #436, opened 2018-10-31
Use .exe and .bat extensions on Windows.

isoos:
#418

@shamblett is there any chance that you can try this branch on your windows setup?


shamblett:
Tested on my Windows10 box, works OK now, json output produced as expected.

PR #449, opened 2018-11-20
Display Dart SDK warning if package doesn't allow the latest Dart 2.X version.

### LINKED ISSUE #448: Redesign: "Support Dart 2 in pubspec.yaml."
We have the following suggestion in packages where the SDK constraint does not support 2.0:

screen shot 2018-11-19 at 8 34 11

This was designed for package with constraints of e.g. `<2.0`, but it also includes packages that have a constraint of `>=2.1 <3.0`.

Couple of options:
- We remove this, as pre-2.0 is legacy, and they won't get analyzed, and such message won't be displayed anyway.
- We increase the limit to always match the latest stable. Couple of options:
  - We specify the latest stable to check for.
  - We introduce a fake version of 2.9999, and if a constraint supports it, we are happy.
  - We check for 2 minor versions up the currently used SDK (e.g. 2.0 will check for 2.2).

/cc @mit-mit @jonasfj - any preference there?

isoos:
Fixes #448.

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -7,7 +7,7 @@ import 'package:yaml/yaml.dart' as yaml;
 
 import 'utils.dart';
 
-final _dart2First = Version.parse('2.0.0');
+final _dart2Last = Version.parse('2.9999.0');
 
 class Pubspec {
   final pubspek.Pubspec _inner;
@@ -105,6 +105,6 @@ class Pubspec {
     if (vc == null) {
       return false;
     }
-    return !vc.allows(_dart2First);
+    return !vc.allows(_dart2Last);
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Reduce the amount of logging when `dartfmt` fails due to an issue in the package.
 
+* Display Dart SDK warning if package doesn't allow the latest Dart 2.X version.
+
 ## 0.12.6
 
 * Bulk processing and comparison.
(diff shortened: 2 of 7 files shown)


lib/src/pubspec.dart
```diff
@@ -105,6 +106,6 @@ class Pubspec {
     if (vc == null) {
       return false;
```
jonasfj: so if there is no constraint on the sdk we don't warn? Ah, because that's a different problem: `hasDartSdkConstraint`

lib/src/pubspec.dart
```diff
@@ -105,6 +106,6 @@ class Pubspec {
...
     }
-    return !vc.allows(_dart2First);
+    return !vc.allows(_dart2First) && !vc.allows(_dart2Last);
```
jonasfj: Why don't we turn this into what sortie suggested and just ask: `!vc.allows(_dart2Last)`
If I understand it correctly that means: "warn if latest dart 2.x release is not supported", if we do this we should probably also change the name of the property...

lib/src/pubspec.dart
isoos: Yeah, `hasDartSdkConstraint` handles that part.

lib/src/pubspec.dart
isoos: Oh, I was thinking about a bit more complex solution (passing in the version for both the check and the error message). I guess using only _dart2Last with a message that is vague on the actual version works for us, I'll update this PR.

isoos:
@jonasfj PTAL, I've updated it to check the latest stable SDK.

PR #458, opened 2018-12-12
update analysis options

pq:
* removes stale analyzer options
* resyncs w/ flutter user options

/cc @isoos @kevmoo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.12.9
+
+* Analysis options used in code health scoring updated remove stale options and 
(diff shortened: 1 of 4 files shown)


isoos:
In general LGTM, but please also:
-  add a line to changelog about it (with 0.12.9 version number)
- fix the test (should be easy, I think they are testing an override or something, and these new values need to be adjusted accordingly)

pq:
Thanks for the feedback.  (And sorry I missed the test in the first place; that's what I get for coding without ☕️!)

Test fixed.  PTAL.

pq:
Thanks!

PR #462, opened 2018-12-18
Downgrade package:analyzer

isoos:
While investigating https://github.com/dart-lang/pub-dartlang-dart/issues/1888, I haven't found much that we can do on our side. The safest path seems to be rolling back https://github.com/dart-lang/pana/pull/456 and adding a test that should indicate we are safe(r) to upgrade again. I'll also report the issue in `package:analyzer`.

- Added a new end2end test: `angular_components`, which first surfaced the `analyzer` issue.
- Updated other tests because `package:convert/src/identity_codec.dart` is a new entry (other changes are ignored in the tests.)


pq:
fyi @stereotype441

Issue #469, opened 2018-12-30
pana's pub upgrade behaves differently from the cli one.


GregorySech:
I'm running pana on a new package but I get this error:
```
SEVERE     Problem with pub upgrade
           Bad state: For analyzer, the parsed version null did not match the locked version0.34.1.
           #0      _validateLockedVersions.<anonymous closure> (package:pana/src/pkg_resolution.dart:67:13)
           #1      SplayTreeMap.forEach (dart:collection/splay_tree.dart:430:8)
           #2      _validateLockedVersions (package:pana/src/pkg_resolution.dart:63:14)
           #3      createPkgResolution (package:pana/src/pkg_resolution.dart:48:5)
           #4      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:187:25)
           <asynchronous suspension>
           #5      PackageAnalyzer.inspectDir (package:pana/src/package_analyzer.dart:103:12)
           #6      main (file:///Users/gregorysech/.pub-cache/hosted/pub.dartlang.org/pana-0.12.9/bin/pana.dart:160:34)
           <asynchronous suspension>
           #7      _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:287:32)
           #8      _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:171:12)
```
However running `pub upgrade` alone does not generate this error.
This is a part of the pubspec file with all dependencies and sdk information
```
version: 0.0.1
environment:
  sdk: ">=2.0.0-dev.68.0 <3.0.0"

dependencies:
  build: ^1.0.2
  source_gen: ^0.9.3
  code_builder: ^3.1.3
  pref_gen_annotations: ^0.0.1

dev_dependencies:
  build_runner: ^1.1.2
```
The problem seems to vanish once the build_runner dev_dependency is moved into production dependencies.

jonasfj:
@isoos, do you know why pana concern its self with `pubspec.lock`? 
Maybe we could leave a comment explaining why we care in:

https://github.com/dart-lang/pana/blob/b8e0b16785259b6be32be7018d05041bd5335953/lib/src/pkg_resolution.dart#L55-L75

A guess might be that it's to ensure the analysis is made against packages of the correct versions. Hence, ensuring that the developer did a `pub get` before running `pana`.

----
And maybe we have a bug with dev_dependencies?

isoos:
[...]

PR #470, opened 2019-01-03
Fix: 'pub upgrade' parsing.

isoos:
Fixes https://github.com/dart-lang/pana/issues/469


isoos:
Rebased, PTAL.

isoos:
> if the printed output format is changing let's bump the minimum SDK version to 2.1.0?

I'm not sure if it is a change from pre-2.1 to 2.1, rather it is a use case which we haven't covered (and probably ignored if it came up in prod).

PR #472, opened 2019-01-07
Prepare 0.12.10 release.

isoos:
I think the parsing fix makes it worth to release it now, I'd like to see it on pub site.


jonasfj:
Poke me to publish tomorrow morning if I forget..

PR #477, opened 2019-01-16
Suppress more log when the pub upgrade error could be git-related.

### LINKED ISSUE #476: warning -> info, for "pub get" messages for packages with git dependencies
We don't want to see warnings in the logs that some git dependency isn't available anymore, or doesn't contain valid content anymore.

If there is git dependencies, we just say that "pub get" failed, instead of writing any logs or warnings.

isoos:
Fixes #476

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -212,9 +212,13 @@ class PackageAnalyzer {
         message = LineSplitter.split(upgrade.stderr as String).first;
       }
 
-      final isUserProblem = pubspec.hasGitDependency ||
-          message.contains('version solving failed') ||
-          message.contains('Git error.');
+      // 1: Version constraint issue with direct or transitive dependencies.
+      //
+      // 2: Code in a git repository could change or disappear.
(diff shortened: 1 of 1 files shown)


lib/src/package_analyzer.dart
```diff
@@ -212,8 +212,10 @@ class PackageAnalyzer {
         message = LineSplitter.split(upgrade.stderr as String).first;
       }
 
-      final isUserProblem = message.contains('version solving failed') ||
-          message.contains('Git error.');
+      final isUserProblem = pubspec.hasGitDependency ||
+          message.contains('version solving failed') ||
+          message.contains('Git error.') ||
+          message.contains('Could not find a file named "pubspec.yaml" in');
```
jonasfj: The last one could happen if a nested package or something weird didn't contain a `pubspec.yaml`?
Hmm.. I guess it's a user problem that you depended on said package then...

jonasfj: Least leave a comment arguing why these error messages are ignored...

isoos: Removing the last one as it should be covered with the git ones.

PR #479, opened 2019-01-16
Strict SDK constraint status

isoos:
I'd like to re-use the same status class in pub site, e.g. to fix the issue, I'd use:

````
bool get supportsOnlyLegacySdk => sdkConstraintStatus.isNotDart2Compatible;
````

https://github.com/dart-lang/pub-dartlang-dart/issues/1952

### CODE REVIEW

**Comment by isoos** [comment]
Please add the constraints here if you have more to check in the test.

**Comment by isoos** [comment]
Build is broken because a new release of `analyzer` did break our end2end tests, will be fixed with the other pending PR.

File: lib/src/pubspec.dart [hunk 1]
```diff
@@ -102,19 +100,75 @@ class Pubspec {
...
+final _range2Latest = VersionConstraint.parse('>=2.9999.0 <3.0.0');
+final _futureRange = VersionConstraint.parse('>=3.0.0');
+
+/// Detailed support coverage for the SDK constraint.
+class SdkConstraintStatus {
+  /// Whether it is non-empty, bounded constraint.
+  final bool hasConstraint;
+
+  /// Whether it allows anything from the <1.0.0 range.
+  final bool enablesDart0;
```
jonasfj: Do we need this information anywhere? Otherwise I suggest we shouldn't expose it...

lib/src/pubspec.dart
```diff
@@ -102,19 +100,75 @@ class Pubspec {
...
+      enablesFutureVersions:
+          hasConstraint && constraint.allowsAny(_futureRange),
+    );
+  }
+
+  /// Whether it is compatible with Dart 2 SDKs.
+  bool get isDart2Compatible =>
+      hasConstraint &&
+      !enablesDart0 &&
+      !enablesDart1 &&
```
jonasfj: Something can be dart2 compatible and dart1 compatible at the same time.

jonasfj: See: https://github.com/dart-lang/convert/blob/master/pubspec.yaml#L8

lib/src/pubspec.dart
isoos: It is useful for completeness, but we could remove it, because I cannot think of a constraint that enables <1.0 and 2.0 but not 1.0, so `enablesDart1` should cover for `isDart2Compatible`.

lib/src/pubspec.dart
isoos: Good point, I haven't considered that use case, will re-work this logic.

isoos: PTAL.

lib/src/pubspec.dart
```diff
@@ -102,19 +100,64 @@ class Pubspec {
...
+final _range2Latest = VersionConstraint.parse('>=2.9999.0 <3.0.0');
+final _futureRange = VersionConstraint.parse('>=3.0.0');
+
+/// Detailed support coverage for the SDK constraint.
+class SdkConstraintStatus {
+  /// Whether it is non-empty, bounded constraint.
+  final bool hasConstraint;
+
+  /// Whether it allows anything from the ^1.0.0 range.
+  final bool enablesDart1;
```
jonasfj: can't we make all of these private?

jonasfj: that way I don't have to wonder if they are accessed from somewhere... and if we refactor it becomes a lot easier...

Also the names are not ideal... or well, they can be misunderstood... I'm not proposing better names :) hehe
I'm just proposing that we hide things other files have no business reading.

PR #491, opened 2019-01-23
Linear penalties for description.

isoos:
I've found it annoying to have 58-character long description with full penalty.

### DIFF
--- a/lib/src/maintenance.dart
+++ b/lib/src/maintenance.dart
@@ -421,19 +421,21 @@ Future<Maintenance> detectMaintenance(
         'Ideal length is between 60 and 180 characters.',
         score: 20.0));
   } else if (description.length < 60) {
+    final penalty = min(20.0, 60.0 - description.length);
     maintenanceSuggestions.add(Suggestion.hint(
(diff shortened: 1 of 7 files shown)


lib/src/maintenance.dart
```diff
@@ -421,19 +421,21 @@ Future<Maintenance> detectMaintenance(
...
+    final penalty = min(20.0, 60.0 - description.length);
     maintenanceSuggestions.add(Suggestion.hint(
         SuggestionCode.pubspecDescriptionTooShort,
         'The description is too short.',
         'Add more detail about the package, what it does and what is its target use case. '
         'Try to write at least 60 characters.',
-        score: 20.0));
+        score: penalty));
   } else if (description.length > 180) {
+    final penalty = min(10.0, description.length.toDouble() - 180.0);
```
lrhn: The `.toDouble()` is unnecessary. Subtracting a double will automatically convert the first operand to `double`

isoos: Thanks, good to know!

PR #492, opened 2019-01-23
Use package:pedantic as default package analysis ruleset.

### LINKED ISSUE #459: Consider using pedantic as a baseline for linting?
Currently there's a hand-crafted list of enforced lints:

https://github.com/dart-lang/pana/blob/master/lib/src/analysis_options.dart

It would be cool to migrate to the [pedantic package](https://pub.dartlang.org/packages/pedantic) rule-set and be able to badge projects "pedantic" clean or something.  

FWIW we're tagging pedantic and flutter user lints now in the generated docs [like so](https://dart-lang.github.io/linter/lints/avoid_empty_else.html):

image

(The flutter badge corresponds to lints in flutter user options.)

/cc @isoos @kevmoo @jonasfj

isoos:
Fixes #459.

We need to add `package:pedantic` in the `dev_dependencies`, and subsequently remove it from the list of resolved packages (unless it is referenced in the original `pubspec.yaml`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Link to package layout conventions in example-related suggestions.
 
+* Use `package:pedantic` as default package analysis ruleset.
+
 ## 0.12.11
 
 * Maintenance score penalty for:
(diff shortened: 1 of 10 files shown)


pq:
Fantastic.  Thank you!

PR #497, opened 2019-01-29
Fix pedantic analysis of package:pedantic.

isoos:
fun fact: you cannot validate `package:pedantic` with arbitrary version of `pedantic`, as it would be a duplicate dependency and won't resolve.

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -436,7 +436,17 @@ class ToolEnvironment {
     final parsed = yamlToJson(original);
     parsed.remove('dev_dependencies');
     parsed.remove('dependency_overrides');
-    if (parsed['name'] != 'pedantic') {
+
+    // Adding pedantic as a dev dependency, because the analysis_options.yaml
+    // file references its ruleset.
+    //
+    // If the package under analysis is pedantic, or pedantic is already a
+    // dependency of the package, we don't need to add it.
+    final isPedantic = parsed['name'] == 'pedantic';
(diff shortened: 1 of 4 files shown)


lib/src/sdk_env.dart
```diff
@@ -436,7 +436,9 @@ class ToolEnvironment {
     final parsed = yamlToJson(original);
     parsed.remove('dev_dependencies');
     parsed.remove('dependency_overrides');
-    parsed['dev_dependencies'] = {'pedantic': '^1.0.0'};
+    if (parsed['name'] != 'pedantic') {
```
jonasfj: Maybe add a comment explaining that this is to support analysis of pedantic?

I'm a little bit wondering why we do it this way... rather than feeding the analyzer rules using a command line argument..  but that seems like cleanup for another day.

isoos: > I'm a little bit wondering why we do it this way... rather than feeding the analyzer rules using a command line argument.. but that seems like cleanup for another day.

We do feed the `analysis-options.yaml` as a command line parameter, but it needs to reference to pedantic's own ruleset, and (AFAICT) that can be only done when `package:pedantic` is added as a dependency.

isoos: In addition to the above, we should also check if `pedantic` is already in the `dependencies` block.

isoos: PTAL

isoos:
Closing in favour of doing direct file references instead.

PR #498, opened 2019-01-29
Fine-tune checked json exception message.

isoos:
error.message may be null


isoos:
> Are we fixing this upstream too?

I'll file a bug for it.

isoos:
https://github.com/dart-lang/json_serializable/issues/378

PR #500, opened 2019-01-29
Apply pedantic rules without referencing package:pedantic in the inspected package's pubspec.

isoos:


lib/src/sdk_env.dart
```diff
@@ -142,11 +143,18 @@ class ToolEnvironment {
     if (await originalOptionsFile.exists()) {
       originalOptions = await originalOptionsFile.readAsString();
     }
+    final pedanticResource =
+        const Resource('package:pedantic/analysis_options.yaml');
+    final pedanticContent = await pedanticResource.readAsString();
+    final pedanticFileName =
+        'pedantic_analyis_options_${DateTime.now().microsecondsSinceEpoch}.g.yaml';
+    final pedanticOptionsFile = File(p.join(packageDir, pedanticFileName));
+    await pedanticOptionsFile.writeAsString(pedanticContent);
```
jonasfj: Why create a new file? Can't we simply reference by global file path? (I didn't test it)

isoos: My first attempt was just that, but we allow a few settings as pass-through, and we can't just replace the user-provided options file.

PR #505, opened 2019-02-04
Support the latest pkg:analyzer, prepare for +1 release

kevmoo:


isoos:
Looks like there were a few changes in `package:analyzer`, is it just a straight upgrade?

kevmoo:
More than a few changes, but things are running fine.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.13+1
```
isoos: 0.12.14

pubspec.yaml
```diff
@@ -1,14 +1,14 @@
 name: pana
 description: Evaluate the health and quality of a Dart package
-version: 0.12.13
+version: 0.12.13+1
```
isoos: 0.12.14-dev

pubspec.yaml
```diff
@@ -1,14 +1,14 @@
...
 homepage: https://github.com/dart-lang/pana
 author: Dart Team <misc@dartlang.org>
 
 environment:
   sdk: '>=2.0.0 <3.0.0'
 
 dependencies:
-  analyzer: ^0.34.2
+  analyzer: '>=0.34.2 <0.36.0'
```
isoos: ^0.35.0

pubspec.yaml
kevmoo: was hoping to publish right away, if it's okay w/ you

pubspec.yaml
kevmoo: was hoping to support a range of versions – or is that not how ya'll roll?

isoos:
> was hoping to publish right away, if it's okay w/ you

It won't get into the pub site for a while, unless it is critical fix to something.

CHANGELOG.md
kevmoo: No behavior change. No new features. Was going to call it a patch bump.

isoos:
> was hoping to support a range of versions – or is that not how ya'll roll?

`package:analyzer` is considered a critical dependency for `pana`, it is better to have a tighter control around it.

Is there any pressing issue you want to have this published right away?

pubspec.yaml
jonasfj: There is no crime in supporting a range of versions..

But since `pana` is usually installed as a tool and not a dependency, what value does it bring.
Does it in any way out weight the cost of maintenance? Clearly, we're not checking the lower bounds :)

To be fair, I don't care, it's very unlikely anyone will install pana with anything but the latest version, since it's a usually installed as a tool with `pub global activate pana`.

PR #512, opened 2019-02-19
Detect usage section in README.md (substitute for example/).

### LINKED ISSUE #510: Don't penalize a package without an example if there is also no public API
Examples:
- https://pub.dartlang.org/packages/devtools
- https://pub.dartlang.org/packages/peanut

isoos:
- Fixes #510
- detects non-empty section headers with text like "Example" or "Usage" 
- `stream` end-to-end test is updated correctly


lib/src/maintenance.dart
```diff
@@ -552,3 +555,25 @@ Future<Maintenance> detectMaintenance(
 
 bool _isMarkdown(String fileName) =>
     p.extension(fileName).toLowerCase() == '.md';
+
+bool _hasExampleInReadmeMarkdown(String text) {
```
jonasfj: Give this a doc comment...

jonasfj: ie. explain what pulls out... give an example README... for how to put examples in READMEs...
```
/// Detects if the README [text] contains an example section.
///
/// ```
/// # Example
/// ....
/// ```
```

lib/src/maintenance.dart
```diff
@@ -552,3 +555,25 @@ Future<Maintenance> detectMaintenance(
...
+  if (text == null || text.isEmpty) return false;
+  final lines = text.split('\n').where((l) => l.trim().isNotEmpty).toList();
```
jonasfj: What about:
```md
Examples
---------
....
```

this is also valid markdown...


I would propose:
1) not doing this
2) adding an `extractTableOfContents` to the `markdown` package on pub (we don't own it, but they'll probably accept contributions).

isoos: Good point.

kevmoo:
CC @kwalrath – should we update our package guidance for this change? - https://www.dartlang.org/tools/pub/package-layout#examples

jonasfj:
@kevmoo, we closed this PR for now.

It's not that we don't wan this, but if we do it should be done properly with markdown parsing. But yeah, some day we might get around to doing this.

Issue #503, opened 2019-02-01
get pana score from cli?


pq:
Is there an easy easy way to get the score for a package when running the command line interface?

pq:
In addition to local package maintenance, we're also looking at doing experiments across pub package code bases to do things like gauge API change impacts.  To help focus our analysis it would be great if we could sort on pana score (or otherwise query it).

pq:
Just seeing #514; cheers for taking a look @isoos!

isoos:
@pq There is an unofficial pub API that exposes the scores, e.g.:
- https://pub.dartlang.org/api/packages/http/metrics?pretty
- https://pub.dartlang.org/api/packages/http/metrics?pretty&reports

pq:
Oh!  For the second use case this is fantastic.  (In that case we don't want to have to run `pana` ourselves.)

Cheers @isoos!

isoos:
#514 has been published with `pana` `0.12.14`

pq:
Many thanks @isoos!

pq:
Thanks again for doing this.  I've added a bot that does some baseline checking for the linter.

https://github.com/dart-lang/linter/pull/1464

I'll be interested to see how useful this regular feedback is.  (Or whether the baseline, currently just "health" and "maintenance" scores, should be more granular.)

PR #514, opened 2019-02-19
Display calculated health and maintenance scores.

isoos:
I wanted to enable #503, but didn't wanted to include the scores in the output model by default, as we are adding additional items on top of it.


pq:
fyi @stereotype441

Issue #515, opened 2019-02-25
Non-actionable feedback for the "analyzer" package


stereotype441:
Looking at https://pub.dartlang.org/packages/analyzer#-analysis-tab-, the analyzer shows up as having health zero.  The only info in "Analysis issues and suggestions" is as follows:

```
Running dartanalyzer . failed with the following output:

Invalid argument(s): Provided content does not align with expectations.
`STDERR exceeded 100000 lines.`
```

I'm not sure how to act on this, and it leads me to suspect a bug in the pana tool.  When I run the analyzer on itself from the command line, I get a much more sensible result:

```
Analyzing analyzer...
  hint • 'fileContentOverlay' is deprecated and shouldn't be used at lib/src/dart/analysis/analysis_context_collection.dart:47:9 • deprecated_member_use_from_same_package
  hint • 'package:analyzer/source/analysis_options_provider.dart' is deprecated and shouldn't be used at lib/src/lint/analysis.dart:13:1 • deprecated_member_use_from_same_package
  hint • 'file:///usr/local/google/home/paulberry/dart-new/sdk/pkg/analyzer/test/generated/constant_test.dart' is deprecated and shouldn't be used at test/generated/test_all.dart:12:1 • deprecated_member_use_from_same_package
  hint • 'file:///usr/local/google/home/paulberry/dart-new/sdk/pkg/analyzer/test/source/embedder_test.dart' is deprecated and shouldn't be used at test/source/test_all.dart:8:1 • deprecated_member_use_from_same_package
4 hints found.
```

stereotype441:
@isoos @jonasfj FYI

isoos:
@stereotype441: pana overrides the package-provided analysis options file in order to create a uniform metric across packages. [the current](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/analysis_options.yaml) options file uses very little linters enabled. We are following [stagehand's defaults](https://github.com/dart-lang/stagehand/blob/master/templates/package-simple/analysis_options.yaml) which now uses `pedantic` as the default linter set.

Could you please update the options file and re-run with that?

stereotype441:
@isoos With this as my analysis options file:
```
include: package:pedantic/analysis_options.yaml
```
I still get a reasonable looking result:
```
Analyzing analyzer...
  hint • 'fileContentOverlay' is deprecated and shouldn't be used at lib/src/dart/analysis/analysis_context_collection.dart:47:9 • deprecated_member_use_from_same_package
[...]

PR #518, opened 2019-02-28
Deduplicate dartanalyzer output lines, and increase timeout.

### LINKED ISSUE #517: dartanalyzer produces a lot of duplicate rows
Debugging #515 highlighted an interesting thing with `dartanalyzer`: running it with normal output will produce `7092 lints and 2 hints found` (`7094` total).

Running it with `--format machine` produced `1443845` lines of output, but after sorting the duplicates, it was only `7094` lines of output. We cut the analysis because the output process is above 100k lines, but we could rather deduplicate the lines as we read them.

Also, reporting issue to `dartanalyzer` repo.

isoos:
Fixes #517, improves #515.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -66,6 +66,7 @@ Future<ProcessResult> runProc(
   var items = await Future.wait(<Future<Object>>[
     process.exitCode,
     byteStreamSplit(process.stdout).forEach((outLine) {
+      // TODO: Remove deduplication when https://github.com/dart-lang/sdk/issues/36062 gets fixed
       if (deduplicate && stdoutLines.contains(outLine)) {
         return;
       }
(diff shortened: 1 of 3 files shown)


lib/src/utils.dart
```diff
@@ -65,6 +66,9 @@ Future<ProcessResult> runProc(
   var items = await Future.wait(<Future<Object>>[
     process.exitCode,
     byteStreamSplit(process.stdout).forEach((outLine) {
+      if (deduplicate && stdoutLines.contains(outLine)) {
```
jonasfj: Why is this necessary? it seems like something that should be fixed else where...

isoos: I've filed an issue for dartanalyzer, but I also want us to unblock from waiting for it to happen.

kevmoo: Perhaps add a todo here w/ a pointer to the issue?

PR #519, opened 2019-03-01
Ask for example only for packages with public libraries.

### LINKED ISSUE #510: Don't penalize a package without an example if there is also no public API

isoos:
Fixes #510

### DIFF
--- a/test/end2end_test.dart
+++ b/test/end2end_test.dart
@@ -13,7 +13,7 @@ import 'package:pana/src/version.dart';
 
 const String goldenDir = 'test/end2end';
 
-final _regenerateGoldens = false;
+final _regenerateGoldens = true;
 
 void main() {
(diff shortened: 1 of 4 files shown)


test/end2end_test.dart
```diff
@@ -13,7 +13,7 @@ import 'package:pana/src/version.dart';
 
 const String goldenDir = 'test/end2end';
 
-final _regenerateGoldens = false;
+final _regenerateGoldens = true;
```
kevmoo: oops?

kevmoo: fixed this for you, @isoos

isoos: Thanks!

PR #523, opened 2019-03-18
Fix: delete local temporary pedantic_analyis_options_[timestamp].g.yaml.

### LINKED ISSUE #522: Unnecessary additional *.g.yaml file isn't deleted from analyzed package.
Why file `pedantic_analizys_option_[digits].g.yaml` doesn't autodelete after analyzing package? I am using a full list of [linter options](https://dart-lang.github.io/linter/lints/options/options.html). But in the created file part of these are just duplicated. 
Pana version 0.12.14.

isoos:
Fixes #522.

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -178,6 +178,7 @@ class ToolEnvironment {
       }
       return output;
     } finally {
+      // TODO: create a withTempFile utility method that deletes these
       await customOptionsFile.delete();
       await pedanticOptionsFile.delete();
     }
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -179,6 +179,7 @@ class ToolEnvironment {
       return output;
     } finally {
       await customOptionsFile.delete();
+      await pedanticOptionsFile.delete();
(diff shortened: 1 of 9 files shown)


lib/src/sdk_env.dart
```diff
@@ -179,6 +179,7 @@ class ToolEnvironment {
       return output;
     } finally {
       await customOptionsFile.delete();
+      await pedanticOptionsFile.delete();
```
jonasfj: How about refactoring this as:
```dart
await withTempFile(name, content, () async {
  ...
});
```

Then have `withTempFile` write the file and delete it after the function given as last parameter is finished...
That way you can easily have both files in their own `finally {` block... which ensures that the second file is deleted, even if the first file wasn't deleted...

side note: couldn't we put these files in `/tmp/`, ideally using https://api.dartlang.org/stable/2.2.0/dart-io/Directory/systemTemp.html

jonasfj: Just a suggestion :)

isoos: I've added a `TODO` to address this later.

PR #525, opened 2019-03-29
Update end2end tests.

isoos:
`package:stream_channel` removed its direct dependency on isolate (got a separate library).


isoos:
> This is because we have tightly coupled tests, right?

Yeah. Keeping track of the changes in dependencies too.

PR #527, opened 2019-05-10
Bump to latest SDK, update end2end files

kevmoo:


kevmoo:
Actually, I want to hold off.

pedantic did something interesting where the lints are actually stored in versioned files.

I think we're missing those now!

kevmoo:
@isoos – thoughts?

isoos:
I'm not sure what's going on: why do we have fewer lint suggestions? Is that a 2.3 SDK change or a pedantic change?

isoos:
> Is that a 2.3 SDK change or a pedantic change?

I'm pretty sure now that this is a change in pedantic that affects us in an unexpected way. Currently the way we do this is the following:
- we load the resource `package:pedantic/analysis_options.yaml` and save it a separate file
- we write our customized `analysis_options.yaml` to point to this file
- we run dartanalyzer point to the custom `analysis_options.yaml`

This is in order to not pollute the package dependency report with pedantic (so that we don't need to remove it either). I'll update the related code to match the new pedantic layout, let's re-do the end2end tests after that.

isoos:
And for the record: CI is auto-upgrading the dependencies, and the new `pedantic` version broke the end2end tests when it came out a week ago or so...

kevmoo:
@isoos – can close this out, right?

isoos:
> @isoos – can close this out, right?

Yes, I'll update the other PR as needed

PR #531, opened 2019-05-21
use latets pkg:json_serializable

kevmoo:


isoos:
> Could you please also increase the version number?

Bumping the version in a separate commit (as we have released 0.12.15 this morning)...

PR #532, opened 2019-05-21
Change `flutter packages` to `flutter pub`

mit-mit:
Related to flutter/flutter#33041

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 A library for analyzing Dart packages. It invokes executables from the Dart SDK
 (or from the Flutter SDK if the package uses Flutter).
 
-* Checks for outdated dependencies (calls `pub upgrade` or `flutter packages upgrade`).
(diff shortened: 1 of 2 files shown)


isoos:
Do we also need to change the executed `flutter packages pub upgrade` -> `flutter pub pub upgrade` or will the `packages` work for a while?

mit-mit:
> Do we also need to change the executed flutter packages pub upgrade -> flutter pub pub upgrade or will the packages work for a while?

`flutter packages` is still supported (it's technically now an alias for `flutter pub`).

But no reason not to shorten that to just `flutter pub upgrade`; will do that now.

isoos:
Just checked: for the command execution we need to use `flutter pub pub upgrade`, because `flutter pub upgrade` does not output the package information (e.g. resolved and available versions).

mit-mit:
Ah, I see, so we rely on the raw output from `pub`? Yes, then `flutter pub pub` is needed.

mit-mit:
@isoos PTAL now

PR #534, opened 2019-05-24
Recognize more file patterns.

isoos:
Following the specification in https://github.com/dart-lang/site-www/pull/1640


lib/src/license.dart
```diff
@@ -166,13 +166,17 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.''');
 
+final _licenseFileNames = <String>[
+  'copying',
+  'license',
+  'unlicense',
```
jonasfj: These have to be uppercase, see: https://github.com/dart-lang/pub/blob/master/lib/src/validator/license.dart

PR #535, opened 2019-05-28
Penalize package constraints that does not support the latest published versions of their dependencies.

isoos:
Fixes https://github.com/dart-lang/pub-dev/issues/2319


lib/src/maintenance.dart
```diff
@@ -531,6 +534,26 @@ Future<Maintenance> detectMaintenance(
         score: 20.0));
   }
 
+  final outdatedDeps = pkgResolution?.outdated ?? <PkgDependency>[];
```
jonasfj: Can we give this block a comment?

PR #563, opened 2019-09-11
Draft of a step abstraction

jonasfj:
This isn't quite ready to do... the logic for the `changelog` step is mostly just a quick hack..

Probably we have to fit the `pubspec` from `pubspec_parse` in to the `AnalysisContext`, probably along with the `.packages` and `pubspec.lock` (which shouldn't exist in the `packageFolder`).

I'm tempted to give up using `package:file`, as it would be somewhat attractive to use [package:test_descriptor](https://pub.dev/packages/test_descriptor). In practice we'll probably always want to create actual files anyways.

This is just a WIP, feel free to leave some suggestions... I'm not exactly sure how we can gracefully migrate existing code to use this flow. IMO, it's okay if we just leave a bunch of the old check as is..

### DIFF
--- a/lib/src/steps/common/changelog.dart
+++ b/lib/src/steps/common/changelog.dart
@@ -1,0 +1,78 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert' show utf8;
+import 'package:file/file.dart' show File;
+import 'package:source_span/source_span.dart';
+import 'package:pubspec_parse/pubspec_parse.dart';
+import '../step.dart';
+
+final _changelogFilePattern = RegExp(r'^CHANGELOG($|\.)', caseSensitive: false);
+
+/// The changelog [Step] will check that:
+///  1. A changelog file exists,
+///  2. The changelog is valid UTF-8 encoded,
+///  3. The changelog mentioned the current version.
+final changelog = Step('changelog', (ctx) async {
+  // Find the changelog file
+  final changelogFile = await ctx.packageFolder.list().firstWhere(
+        (e) => e is File && _changelogFilePattern.hasMatch(e.basename),
+        orElse: () => null,
+      ) as File;
+  // If there is no changelog file, then we'll leave a hint and be done.
+  if (changelogFile == null) {
+    BreakStep.hint(
+      kind: 'changelog-not-found',
+      title: 'CHANGELOG.md missing',
+      description: 'Consider including a `CHANGELOG.md` file outlining the '
+          'changes from one version to the next.',
+      penalty: 0.1,
+    );
+  }
+
+  final pubspecFile = ctx.packageFolder.childFile('pubspec.yaml');
+  if (!await pubspecFile.exists()) {
+    return; // Someone else will handle this
+  }
+  final pubspec = Pubspec.parse(
+    await pubspecFile.readAsString(),
+    lenient: true,
+    sourceUrl: pubspecFile.path,
+  );
+
+  // Read the changelog file
+  final changelogBytes = await changelogFile.readAsBytes();
+  String changelog;
+  try {
+    changelog = utf8.decode(changelogBytes);
+  } on FormatException {
+    ctx.add.warning(
+      kind: 'changelog-encoding',
+      title: 'Changelog encoding',
+      description: 'Use utf8 encoding for `${changelogFile.basename}`',
+      location: SourceFile.decoded(
+        changelogBytes,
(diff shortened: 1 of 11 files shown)


jonasfj:
@isoos, your thoughts on this would be appreciated :)

lib/src/steps/common/changelog.dart
```diff
@@ -0,0 +1,78 @@
...
+        orElse: () => null,
+      ) as File;
+  // If there is no changelog file, then we'll leave a hint and be done.
+  if (changelogFile == null) {
+    BreakStep.hint(
+      kind: 'changelog-not-found',
+      title: 'CHANGELOG.md missing',
+      description: 'Consider including a `CHANGELOG.md` file outlining the '
+          'changes from one version to the next.',
+      penalty: 0.1,
```
isoos: I think 10% penalty should be a warning instead, not a hint.

lib/src/steps/result.dart
```diff
@@ -0,0 +1,312 @@
...
+  ///  * [title], a human-readable title for the suggestion,
+  ///  * [description], a human-readable markdown description of the suggestion,
+  ///  * [location], (optional) span of source code that caused this suggestion,
+  ///  * [penalty], (optional) penalty for package score ranging from 0 to 1.
+  void error({
+    String kind,
+    @required String title,
+    @required String description,
+    SourceSpan location,
+    double penalty = 0,
```
isoos: Let's make this `@required penalty,` with no predefined value.

lib/src/steps/result.dart
```diff
@@ -0,0 +1,312 @@
...
+  ///  * Missing dependencies.
+  ///
+  /// A suggestion have following parameters:
+  ///  * [kind], (optional) kind identifier for statistics and grouping,
+  ///  * [title], a human-readable title for the suggestion,
+  ///  * [description], a human-readable markdown description of the suggestion,
+  ///  * [location], (optional) span of source code that caused this suggestion,
+  ///  * [penalty], (optional) penalty for package score ranging from 0 to 1.
+  void error({
+    String kind,
```
isoos: should we make this `@required`?

lib/src/steps/result.dart
```diff
@@ -0,0 +1,312 @@
...
+  final String kind;
+  final String _level;
+  final String _title;
+  final String _description;
+  final SourceSpan _loc;
+  final double _penalty;
+
+  BreakStep._({
+    this.kind,
+    String level,
```
isoos: I imagine that throwing an exception should happen when we find a serious issue, and the `kind`, `level` and `penalty` should be `@required`.

lib/src/steps/step.dart
```diff
@@ -0,0 +1,41 @@
...
+@sealed
+class Step {
+  /// Unique name of this analysis step.
+  ///
+  /// This **must** be a unique lowercase identifier matching `^[a-z0-9_-]+$`.
+  final String name;
+
+  /// Run analysis given an [AnalysisContext].
+  ///
+  /// This may throw [BreakStep] to report a final suggestion and stop analysis.
```
isoos: Wondering about this: why should we combine the final suggestion with the exception?

jonasfj: Throwing it is optional...

It allows us to make helper functions that throws an exception that stops analysis for the given step and sets a suggestion, without access to the context...

It's heavily inspired by what we did with `ResponseException` in pub-dev.

I'm realizing now, we should stop prefixing `kind` with the step name. And then we should remove duplicate suggestions...

Then we can make a function that reads and parses the `pubspec.yaml`. If it fails it throw a `BreakStep` exception, thus, a step that uses this function doesn't need to check if the return value is `null`, nor does it need to catch the exception. Most steps that need a parsed pubspec will just give up doing anything else and return.

jonasfj: Note to self: refactor Step into having a constructor that creates a single step from a list of steps...

Then I can also simplify to `runStep` instead of `runSteps`.

lib/src/steps/result.dart
jonasfj: Probably... How do you feel if we give up the automatic prefixing... It's natural to write kinds like `changelog-missing` anyways...

And if we have clashing codes it's not really a big deal, anyways, we use them for stats, collapsing and testing... At least that's what I imagine..

lib/src/steps/step.dart
isoos: > It allows us to make helper functions that throws an exception that stops analysis for the given step and sets a suggestion, without access to the context...

Please add this paragraph to the comment block.

kevmoo:
@jonasfj – guessing we can nuke this, right?

jonasfj:
Yes, I think we've abandoned this idea..

PR #565, opened 2019-11-04
Fix end2end tests on CI

isoos:
- some of it is only diffferent JSON key order
- the breaking parts were: new dependencies and different analysis errors


test/end2end/stream-2.0.1.json
```diff
@@ -1075,231 +1077,6 @@
           "col": 9,
           "description": "Unnecessary new keyword."
         },
-        {
```
sigurdm: Why did all of these disappear?

isoos: either a change in the analyzer or a change in the dependencies

PR #567, opened 2019-11-15
Flutter platform spec usage

sigurdm:


sigurdm:
cc @mit-mit

lib/src/pubspec.dart
```diff
@@ -112,6 +112,15 @@ class Pubspec {
 
   SdkConstraintStatus get sdkConstraintStatus =>
       SdkConstraintStatus.fromSdkVersion(_inner.environment['sdk']);
+
+  bool get usesOldFlutterPluginFormat =>
+      usesFlutter &&
+      _inner.flutter['plugin'] != null &&
+      const {
```
isoos: `_inner.flutter['plugin'] is Map &&`

sigurdm: Good catch.
Done

sigurdm:
We should only land this after support for the new format is in flutter/stable

sigurdm:
@jonasfj we are ready to land this now, right?

PR #573, opened 2019-11-25
Fix: pre-release versions are not considered when detecting outdated dependencies.

### LINKED ISSUE #572: Don't penalise for not supporting latest dev dependencies
My plugin depends on the latest stable release of rxdart (0.22.6) but is penalised -10 points because 0.23.0-dev.1 is available.

isoos:
- fixes #572 
- end2end test confirmed fix on [angular_components 0.10.0](https://pub.dev/packages/angular_components/versions/0.10.0#-analysis-tab-)

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -628,6 +628,7 @@ class PkgDependency implements Comparable<PkgDependency> {
 
   bool get isLatest => available == null;
 
+  // TODO: investigate if `pub upgrade` reports the latest stable or the latest uploaded
   bool get isOutdated => available != null && !available.isPreRelease;
 
   bool get isHosted =>
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -628,7 +628,7 @@ class PkgDependency implements Comparable<PkgDependency> {
 
   bool get isLatest => available == null;
 
-  bool get isOutdated => !isLatest;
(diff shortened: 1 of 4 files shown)


lib/src/model.dart
```diff
@@ -622,7 +622,7 @@ class PkgDependency implements Comparable<PkgDependency> {
 
   bool get isLatest => available == null;
 
-  bool get isOutdated => !isLatest;
+  bool get isOutdated => available != null && !available.isPreRelease;
```
sigurdm: Can it happen that there is both a newer prerelease and a  newer nonprerelease available, and the prerelease shadows the non-prerelease making the dependency seem not outdated even though it is?

isoos: The information is coming from the `pub` client, and we'll take here whatever it provides.

Any memory what `pub` uses? (latest stable, latest by date, or just highest version value?) I don't see much risk of using this at face value right now, but later on we could also query it from  from pub.dev.

sigurdm: True, at worst we will warn too little.

I don't know off hand what the pub client does.
Perhaps add a TODO that we might want to investigate edge cases here.

isoos:
CI is green after rebase, but did not update the status here. Merging.

PR #574, opened 2019-11-25
Calculate sdk tags

sigurdm:


lib/src/package_analyzer.dart
```diff
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:async/async.dart';
```
isoos: do you need this?

lib/src/pubspec.dart
```diff
@@ -112,6 +112,21 @@ class Pubspec {
 
   SdkConstraintStatus get sdkConstraintStatus =>
       SdkConstraintStatus.fromSdkVersion(_inner.environment['sdk']);
+
+  /// The sdk tags associated with [this].
+  List<String> sdkTags() {
+    final result = <String>[];
+    final dependentSdks = this.dependentSdks;
+    if (dependentSdks.every(['dart'].contains)) {
```
isoos: I find this hard to parse. Isn't this the same as:
```dart
if (dependentSdks.length == 1 && dependentSdks.single =='dart') {
```

lib/src/pubspec.dart
```diff
@@ -112,6 +112,21 @@ class Pubspec {
...
+      result.add('sdk:dart');
+    }
+    if (dependentSdks.every(['dart', 'flutter'].contains)) {
```
isoos: Same here:
```dart
if (dependentSdks.length == 2 && dependentSdks.containsAll(['dart', 'flutter']) {
```

lib/src/package_analyzer.dart
sigurdm: Nope - removed

lib/src/pubspec.dart
sigurdm: Yes, I wrote it like that initially, but changed it to follow the same pattern as below.

lib/src/pubspec.dart
sigurdm: No - not exactly the same.
the expression i wrote is true for `dependentSdks = ['dart']` (You get the 'sdk:flutter' tag as soon as you can be used in flutter, not if you require flutter).
I agree that it is cryptic, and would be happy to write it in a better way.

isoos: I see, makes sense. Please add a code comment with this information, so that we'll remember it a few months down the line :).

sigurdm: Good idea! Thanks

PR #576, opened 2019-11-26
Utility for analyzing configurable imports/exports

jonasfj:
@sigurdm, I think we can resolve the configurable imports/exports by simply analyzing the AST.
Looking at the grammar it seems very simple.

I'm not entirely sure what we should do when:
 * Parsing fails,
 * Declared variables are undefineds,

I suspect  we can ignore parsing errors (these will be analyzer errors, anyways), and as long as declared variables are always `true` or `false`, then I suspect the cases where they are unrecognized can be reported as warning for the package. This way, we are sure to get a bug report, when users start using a new variable and we haven't updated pana to reflect that.

The only problem I see is that part files might one day be allowed to do configurable imports, but this might not break the AST to a level that we'll remember to update this. But maybe the best long-term solution is to have the analyzer do this for us.

Upsides to this approach is that it's a bit faster than a full analysis, and if the code has analysis errors we can still do the dependency analysis.

Note. this only resolved immediate imports/exports, so we need propagate these until we reach a fixed-point. But I think this is the tool we need to get configurable imports working :)

### DIFF
--- a/lib/src/dependency_analysis.dart
+++ b/lib/src/dependency_analysis.dart
@@ -1,0 +1,96 @@
+import 'package:analyzer/dart/analysis/session.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+
+/// Dependencies for library at `package:` [uri] under some configuration of
+/// declared variables.
+class LibraryDependencies {
+  final Uri uri;
+  final Set<Uri> dependencies;
+
+  LibraryDependencies(this.uri, Iterable<Uri> dependencies)
+      : dependencies = Set.from(dependencies);
+
+  @override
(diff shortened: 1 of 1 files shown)


sigurdm:
yes! I think this is the way we should go.

> I'm not entirely sure what we should do when:
> Parsing fails,

Stop the analysis (hopefully some other part of pana will yell at the program)

> Declared variables are undefineds,

I think that is the same as that they are false.

> The only problem I see is that part files might one day be allowed to do configurable imports, but this might not break the AST to a level that we'll remember to update this. 

I don't think we'll see that day. And in any case, if the language changes under us we'll anyway have to adapt.

PR #578, opened 2019-11-29
tagging with sdk: platform: and runtime: tags

sigurdm:


lib/src/tag_detection.dart
```diff
@@ -0,0 +1,372 @@
...
+
+import 'package:analyzer/dart/analysis/session.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/analysis/context_builder.dart';
+import 'package:analyzer/dart/analysis/context_locator.dart';
+import 'package:collection/collection.dart';
+import 'package:pana/pana.dart';
+import 'package:path/path.dart' as path;
+
+abstract class Graph<T> {
```
jonasfj: Document that this is a directed graph.

lib/src/tag_detection.dart
```diff
@@ -0,0 +1,372 @@
...
+}
+
+class Runtime {
+  final String name;
+  final Set<String> enabledLibs;
+
+  Runtime(this.name, this.enabledLibs);
+
+  Map<String, String> get declaredVariables =>
+      {for (final lib in enabledLibs) 'dart.library.$lib': 'true'};
```
jonasfj: Need to make the `dart.library.js` to `false` when not present? or is that not how it works? I realize that our logic is largely ignorant of this.

lib/src/tag_detection.dart
```diff
@@ -0,0 +1,569 @@
...
+  Set<Uri> directSuccessors(Uri uri) {
+    return _cache.putIfAbsent(uri, () {
+      final uriString = uri.toString();
+      if (uriString.startsWith('dart:') || uriString.startsWith('dart-ext:')) {
+        return <Uri>{};
+      }
+      final unitResult = _analysisSession.getParsedUnit(
+        _analysisSession.uriConverter.uriToPath(uri),
+      );
+      if (unitResult.isPart || unitResult.unit == null) {
```
jonasfj: This means we ignore errors `unitResult.unit == null`.

PR #581, opened 2019-12-05
Avoid crashing the tagging when there is no lib/ folder.

sigurdm:


lib/src/tag_detection.dart
```diff
@@ -498,20 +498,28 @@ class Tagger {
         .currentSession;
     final pubspecCache = _PubspecCache(session);
     final pubspec = pubspecCache.pubspecOfPackage(packageDir);
-    final files = Directory('$packageDir/lib')
-        .listSync(recursive: false)
+    List<FileSystemEntity> libFiles;
```
isoos: Why not check for `Directory('$packageDir/lib').existsSync()`?

sigurdm: It might not be relevant here, but in general you cannot trust that because you checked a file exists it will exist when you read it afterwards. It is usually more robust to try the operation, and handle errors.
I'm happy to change the code here, as I don't think it will cause real problems.

isoos: Fair point in the general case, but I don't think we'll need to protect the processing against file changes like that.

sigurdm:
Could not merge this for some reason.
Superseded by https://github.com/dart-lang/pana/pull/582.

PR #583, opened 2019-12-06
Fix CI

isoos:
`package:pedantic` has a new version with much more lints, most of them fixed, one of them ignored with comment (that code may go away soon).

Probably we should update the packages (or their versions) in the end2end tests, it is getting way too large...

### DIFF
--- a/bin/bulk.dart
+++ b/bin/bulk.dart
@@ -17,7 +17,7 @@ import 'package:pool/pool.dart';
 
 final _jsonEncoder = const JsonEncoder.withIndent('  ');
 
-main(List<String> args) async {
+Future main(List<String> args) async {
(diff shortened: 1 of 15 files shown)


bin/bulk.dart
```diff
@@ -17,7 +17,7 @@ import 'package:pool/pool.dart';
 
 final _jsonEncoder = const JsonEncoder.withIndent('  ');
 
-main(List<String> args) async {
+Future main(List<String> args) async {
```
sigurdm: Can we use `Future<void>` here and below?

sigurdm:
+1 for updating the packages (after doing this).
At least for those packages that do not exhibit some special behavior we want to track.

PR #584, opened 2019-12-09
Fix tag calculation

sigurdm:
Two issues:
primary library was calculated wrong
flutter-web should allow dart:io imports

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.13.1+3
 
-* Allow import of dart:io on platform web when calculating tags.
+* Allow import of platform:flutter on platform web when calculating tags.
 * Fix the finding of primary library.
 
 ## 0.13.1+2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.13.1+3
+
+* Allow import of dart:io on platform web when calculating tags.
+* Fix the finding of primary library.
+
 ## 0.13.1+2
 
 * More handling of degenerate cases when computing sdk, platform and runtime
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.13.1+3
+
+* Allow import of dart:io on platform web when calculating tags.
```
jonasfj: Let's special case `package:flutter` instead, importing `dart:io` won't work, see:
https://flutter.dev/docs/development/platform-integration/web#can-i-use-dartio-with-a-web-app

lib/src/tag_detection.dart
```diff
@@ -224,6 +228,9 @@ class _PackageGraph implements _DirectedGraph<String> {
   Set<String> directSuccessors(String packageDir) {
     final pubspec = _pubspecCache.pubspecOfPackage(packageDir);
     return pubspec.dependencies.keys
+        // HACK: Ignore package:flutter as it is provided by the SDK.
+        // TODO(jonasfj): Look at the package dependency kind, we should ignore SDK dependencies
+        .filter((name) => name == 'flutter')
```
jonasfj: @sigurdm, does this seem right?

lib/src/tag_detection.dart
```diff
@@ -153,6 +153,10 @@ class LibraryGraph implements _DirectedGraph<Uri> {
       if (uriString.startsWith('dart:') || uriString.startsWith('dart-ext:')) {
         return <Uri>{};
       }
+      // HACK: Package flutter comes from the SDK, we do not want to look at its import graph.
+      if (uriString.startsWith('package:flutter/') {
+        return <Uri>{};
```
jonasfj: @sigurdm, does this seem right?

lib/src/tag_detection.dart
sigurdm: Why is this needed?

CHANGELOG.md
jonasfj: update changelog

PR #592, opened 2019-12-16
Updated tag detection for packages without a primary library.

### LINKED ISSUE #591: linter: missing sdk.version... a little help?
As of the lastest `pana` release, the linter's maintenance score has dropped w/ this warning:

```yaml
 "maintenance": {
  "missingChangelog": false,
  "missingExample": false,
  "missingReadme": false,
  "missingAnalysisOptions": false,
  "oldAnalysisOptions": false,
  "strongModeEnabled": true,
  "isExperimentalVersion": false,
  "isPreReleaseVersion": false,
  "suggestions": [
   {
    "code": "sdk.missing",
    "level": "error",
    "title": "No valid SDK.",
    "description": "The analysis could not detect a valid SDK that can use this package.",
    "score": 20.0
   }
  ]
 },

```

Build failure: https://travis-ci.org/dart-lang/linter/jobs/624732589

Linter pubspec: https://github.com/dart-lang/linter/blob/master/pubspec.yaml

Could you help me resolve?

/cc @isoos @jonasfj

isoos:
- Fixes #591.
- Top-level libraries are the files in `lib/*.dart`. However, if there is a primary library in the package (`/lib/[package_name].dart`), it will be used as a single value in top-level libraries.
- Tags are added when all of the top-level libraries agree on supporting the given SDK or platform.
- Binary-only packages are handled via a temporary heuristic.

### DIFF
--- a/lib/src/tag_detection.dart
+++ b/lib/src/tag_detection.dart
@@ -490,11 +490,12 @@ class Tagger {
   String packageDir;
   final AnalysisSession _session;
   final _PubspecCache _pubspecCache;
+  final bool _isBinaryOnly;
   final List<Uri> _topLibraries;
   final _PackageGraph _packageGraph;
 
   Tagger._(this.packageDir, this._session, _PubspecCache pubspecCache,
-      this._topLibraries)
+      this._isBinaryOnly, this._topLibraries)
       : _pubspecCache = pubspecCache,
         _packageGraph = _PackageGraph(pubspecCache);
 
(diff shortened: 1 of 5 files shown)


pq:
Thank you @isoos!

lib/src/tag_detection.dart
```diff
@@ -509,28 +509,49 @@ class Tagger {
...
-      return Tagger._(packageDir, session, pubspecCache, null);
+    final libSrcDir = Directory(path.join(packageDir, 'lib', 'src'));
+    final libSrcDartFiles = !libSrcDir.existsSync()
+        ? <String>[]
+        : libSrcDir
+            .listSync(recursive: false)
+            .where((e) => e is File && e.path.endsWith('.dart'))
+            .map((f) => path.relative(f.path, from: libDir.path))
+            .toList();
+    // Sort to make the arbitrary use of first file deterministic.
```
sigurdm: Update this comment. We are now no longer using the first file.

isoos:
> ideally we should recognise packages that are mainly executable and have another tagging scheme for them, or at least use their declared executables from the pubspec as root libraries.

I'm not entirely sure that we should treat them that much different on this stage of the analysis. E.g. in case of `package:linter`, while it may be a "binary-only" package, I can still import its internals (`lib/src/.../whatever.dart`), use them, and they may just work fine.

Of course we should also tag  "binary-only" status, which would change the instructions on the "Install tab"...

sigurdm:
> I can still import its internals (lib/src/.../whatever.dart), use them, and they may just work fine.

You _can_, but the [convention](https://dart.dev/guides/language/effective-dart/usage#dont-import-libraries-that-are-inside-the-src-directory-of-another-package) is to not import internals, and I don't think we should provide an analysis that suggests what works in case you break the convention.

isoos:
That's a good point. Should I remove the `lib/src` part from this one, or are doing that later?

sigurdm:
Could you look up the [executables](https://dart.dev/tools/pub/pubspec#executables), and use those in case there are nothing in lib/ ?

isoos:
I've updated the PR, checking the `bin/` directory for files, but only with a (probably temporary) heuristic:
- not checking the `pubspec.executables`, because it is optionals, and many binary packages, including `package:linter` and my own codegen library ignored it (it is useful for having an alias, but not a requirement for `pub run ...`)
- it returns `sdk:dart` and `runtime:native-*` as tags, without really checking the referenced libraries

pq:
Fantastic!  Any chance you could push out a release with this fix?

Much appreciated!

isoos:
@pq: Unfortunately there are no production deployments around the holidays, I'd expect this to get into prod in early January.

pq:
No worries.  That makes perfect sense.

Hope you enjoy some time off! ❄️

Issue #605, opened 2020-01-15
Tag and analysis failure for packages that have libraries, but not in the lib/*.dart


isoos:
e.g. https://pub.dev/packages/getflutter

PR #606, opened 2020-01-15
Use lib/**/*.dart as top-level libraries when there is no lib/*.dart

isoos:
#605


isoos:
this needs approval to merge

sigurdm:
Sorry - approval should be in now...

PR #609, opened 2020-01-27
Fix end2end tests.

isoos:
New `package:logging` files in the full dependency lists.


sigurdm:
Should we rewrite the end2end tests to be in a world with a fixed sub-set of packages?
This seems like a never-ending task

isoos:
> Should we rewrite the end2end tests to be in a world with a fixed sub-set of packages?
> This seems like a never-ending task

It is a tradeoff: if we restrict the tests to be only on a subset of features or frozen in time: the test will pass, but we may miss something. It would be interesting to have a change-monitoring/change-verification outside of the CI tests, but I don't know how it could work with much less effort...

PR #618, opened 2020-03-09
Fix CI

isoos:


isoos:
> This is mostly because new packages have been published, so golden files have new scores..

At first I thought it was only that, but we also have this:

````
Analyzing pana...
  hint • 'package:package_config/packages_file.dart' is deprecated and shouldn't be used. Use the package_config.json based API. • lib/src/library_scanner.dart:13:1 • deprecated_member_use
1 hint found.
````

jonasfj:
Yeah, latest version of `package_config` is 1.9.1 where everything has been deprecated and the new API that will be 2.0.0 has been added. So anyone using package_config will get this message.

isoos:
> Yeah, latest version of package_config is 1.9.1 where everything has been deprecated and the new API that will be 2.0.0 has been added. So anyone using package_config will get this message.

Suppressing the message for now, as the migration also requires us to use an async API instead of the current sync one.

PR #619, opened 2020-03-09
Use new package:package_config API

isoos:


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.13.6
 
 * Fix: detect Dart files on Windows
+* Migrated to use new `package:package_config` API.
```
jonasfj: Also update the pubspec.yaml to say `package_config: >=1.9.1 <3.0.0`

lib/src/library_scanner.dart
```diff
@@ -40,40 +38,34 @@ class LibraryScanner {
...
       throw StateError('A package configuration file was not found at the '
           'expected location.\n$dotPackagesPath');
     }
-    var dotPackagesBytes = File(dotPackagesPath).readAsBytesSync();
-    var packageMap =
-        package_config.parse(dotPackagesBytes, p.toUri(dotPackagesPath));
+    final dotPackagesFile = File(dotPackagesPath);
+    final config = await package_config.loadPackageConfig(dotPackagesFile);
 
     String package;
```
jonasfj: Can we add a comment, it's not obvious what this is trying to do.

lib/src/library_scanner.dart
```diff
@@ -40,40 +38,34 @@ class LibraryScanner {
...
-      }
-
-      if (p.isWithin(packagePath, p.fromUri(v))) {
-        packageNames.add(k);
+    final packageNames = <String>[];
+    config.packages.forEach((pkg) {
+      if (package != null) return;
+
+      final pkgRootFilePath = Directory.fromUri(pkg.root).path;
+      if (pkgRootFilePath == '$packagePath${Platform.pathSeparator}') {
```
jonasfj: Wouldn't it be safer to normalize paths and compare them?

lib/src/library_scanner.dart
```diff
@@ -40,40 +38,34 @@ class LibraryScanner {
   LibraryScanner._(
       this.packageName, this._packagePath, this._session, this._overrides);
 
-  factory LibraryScanner(String dartSdkPath, String packagePath,
-      {List<LibraryOverride> overrides}) {
+  static Future<LibraryScanner> create(String dartSdkPath, String packagePath,
+      {List<LibraryOverride> overrides}) async {
```
jonasfj: Wouldn't it be safer to pass in a `packageName` to be analyzed instead of `packagePath`? Or read the package name from `pubspec.yaml`..

lib/src/library_scanner.dart
```diff
@@ -40,40 +38,34 @@ class LibraryScanner {
...
-        package = k;
-      }
-
-      if (p.isWithin(packagePath, p.fromUri(v))) {
-        packageNames.add(k);
+    final packageNames = <String>[];
+    config.packages.forEach((pkg) {
+      if (package != null) return;
+
+      final pkgRootFilePath = Directory.fromUri(pkg.root).path;
```
jonasfj: Seems safer to use `Uri` + `Uri.normalize()` or something like that.

lib/src/library_scanner.dart
jonasfj: We can also leave a comment to cleanup this logic.

lib/src/library_scanner.dart
```diff
@@ -40,40 +38,34 @@ class LibraryScanner {
...
+        package = pkg.name;
+      } else if (p.isWithin(packagePath, pkgRootFilePath)) {
+        packageNames.add(pkg.name);
       }
     });
 
     if (package == null) {
       if (packageNames.length == 1) {
         package = packageNames.single;
-        log.warning('Weird: `$package` at `${packageMap[package]}`.');
```
jonasfj: I don't see this in logs anywhere

isoos:
> But it might just be easier to cleanup and the right thing... unless I'm missing what this is supposedly doing.

The `package_config` reports the directory with an ending `/` while our code gets it without that. I'm not sure what kind of cleanup we could do, somewhere we need to convert between the two.

lib/src/library_scanner.dart
isoos: That's the normalization right there. :)
Do you know any method in `package:path` that takes a directory path and returns it without the end file separator? e.g. `a/b/c/` -> `a/b/c`

lib/src/library_scanner.dart
isoos: Yeah, that would make much more sense, I don't remember why we have it this way...

lib/src/library_scanner.dart
isoos: I don't think `Uri.normalizePath` removes the last `/` at the end of the path.

isoos:
Closing this, #620 seems to be better.

PR #620, opened 2020-03-09
Fix TODO: do not need to auto-detect packageName in LibraryScanner

isoos:


kevmoo:
@isoos – need to tweak formatting!

PR #622, opened 2020-03-16
Ignore file: imports when analyzing tags

sigurdm:
Fixes https://github.com/dart-lang/pub-dev/issues/3430

### DIFF
--- a/test/tag_detection_test.dart
+++ b/test/tag_detection_test.dart
@@ -398,5 +398,43 @@ name: my_dependency
       expect(tagger.runtimeTags(),
           {'runtime:native-aot', 'runtime:native-jit', 'runtime:web'});
     });
+
+       test('file: imports are ignored', () async {
+      final decriptor = d.dir('cache', [
+        d.dir('my_package', [
+          d.dir('lib', [
+            d.file('my_package.dart', '''
+import 'file:/abc/def.dart';
+int fourtyTwo() => 42;
+'''),
+          ]),
+          d.file('.packages', '''
+my_package:lib/
+'''),
+          d.file(
+            'pubspec.yaml',
+            '''
+name: my_package
+''',
+          )
+        ])
+      ]);
+      await decriptor.create();
+      final tagger = Tagger('${decriptor.io.path}/my_package');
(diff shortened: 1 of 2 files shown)


test/tag_detection_test.dart
```diff
@@ -397,6 +397,44 @@ name: my_dependency
...
       expect(tagger.runtimeTags(),
           {'runtime:native-aot', 'runtime:native-jit', 'runtime:web'});
+    });
+
+       test('file: imports are ignored', () async {
+      final decriptor = d.dir('cache', [
+        d.dir('my_package', [
+          d.dir('lib', [
+            d.file('my_package.dart', '''
+import 'file:/abc/def.dart';
```
isoos: Shouldn't we rather fail the analysis of this library? We don't have the file, the package likely won't be able to work properly.

sigurdm: I guess a package failing here would have analyzer errors, that will already give a penality elsewhere.

The question is: should we even attempt the tag detection if there is analyzer errors?

I still think it makes sense to attempt the tag detection, and then it needs to be robust.
@jonasfj WDYT?

jonasfj: uh, `file:` imports... interesting, haha,

Such packages must be inherently broken -- right?

I'm okay not giving tags if there is analyzer bugs of this character, I'm also okay, ignoring the error.
So long as `pana` doesn't crash.. The most correct thing might be to not classify the packages..

HOWEVER; any libraries in Dart / Flutter SDK could use file: imports... I suppose.. I would be surprised, but if the file exists then I think we should assume it's part of the SDK.

sigurdm: Created https://github.com/dart-lang/pana/pull/623/files that avoids the tag-detection when errors have been detected.

We could land both.

> HOWEVER; any libraries in Dart / Flutter SDK could use file: imports... I suppose.. I would be surprised, but if the file exists then I think we should assume it's part of the SDK.

I would be extremely surprised too. What would that even mean?

jonasfj: > What would that even mean?

I suppose importing from an absolute `file:` known to be inside the SDK is a reasonable thing to do.. that said I doubt we do it, because the SDK can be moved around :)

PR #625, opened 2020-03-19
Update README.md

GroovinChip:
Add note about potential IDE restart required

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Used by the [Dart Package site](https://pub.dev/).
 > pub global activate pana
 ```
 
+NOTE: You may need to restart your IDE for pana commands to be recognized.
+
 ### Usage
 
 You can specify either a package (+ version) or a local directory to analyze:


README.md
```diff
@@ -19,6 +19,8 @@ Used by the [Dart Package site](https://pub.dev/).
 > pub global activate pana
 ```
 
+NOTE: You may need to restart your IDE for pana commands to be recognized.
```
sigurdm, line 22: Is this a general thing for pub global activate?

What ide did you experience this with?

GroovinChip:
I experienced this with Android Studio

sigurdm:
And how do you call pana from android studio? Does it have its own button? Or is it from the shell?

GroovinChip:
It is from the shell

sigurdm:
Usually bash picks up new binaries from `PATH` automatically without a restart - so I'm not sure why this is needed.

GroovinChip:
The new text only says a restart *may* be needed

sigurdm:
Hmm - so many things *may* be needed under different circumstances we cannot enumerate them all.

If a restart of the IDE **is** needed I think the advice is very helpful.

I would like to understand the mechanism at play here better before including more instructions here. Then we can also make more precise instructions. Maybe we should write: 

> NOTE: You may need to restart your shell for pana commands to be recognized.

But without understanding why bash doesn't pick this up, it is hard to recommend something specific.

GroovinChip:
Yeah, I don't know what would cause the problem so a generic statement like
that would be helpful to me, personally. I can change it to your suggestion
of you like?

PR #627, opened 2020-03-24
Bump version

sigurdm:


isoos:
You'll also need to run `pub run build_runner build` to update the generated version.

Issue #628, opened 2020-03-25
NSME in DartSdkInfo.parse (pana:0.13.6) 


pq:
As of ` 0.13.6` we're seeing failures in our build that uses pana:

https://travis-ci.org/github/dart-lang/linter/jobs/666901407?utm_medium=notification&utm_source=github_status

```
Running pana...
{"logName":"pana","level":"INFO","message":"Running `/home/travis/dart-sdk/bin/dart --version`..."}
Unhandled exception:
NoSuchMethodError: The method '[]' was called on null.
Receiver: null
Tried calling: [](1)
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
#1      new DartSdkInfo.parse (package:pana/src/sdk_env.dart:507:38)
#2      ToolEnvironment._init (package:pana/src/sdk_env.dart:63:37)
<asynchronous suspension>
#3      ToolEnvironment.create (package:pana/src/sdk_env.dart:135:19)
<asynchronous suspension>
#4      PackageAnalyzer.create (package:pana/src/package_analyzer.dart:68:50)
#5      main (file:///home/travis/.pub-cache/hosted/pub.dartlang.org/pana-0.13.6/bin/pana.dart:131:44)
<asynchronous suspension>
#6      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#7      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

Is this a known issue?

PR #629, opened 2020-03-25
Handle new output from --version in latest Dart SDK

kevmoo:
It may contain a date, which needs to be considered in the parser

Fixes https://github.com/dart-lang/pana/issues/628
Prepare for v0.13.7

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -493,7 +493,7 @@ class DartdocResult {
 
 class DartSdkInfo {
   static final _sdkRegexp = RegExp(
-    r'Dart VM version:\s([^\s]+)\s\(([^\)]+)\)(\s\(([^\)]+)\))? on "(\w+)"',
+    r'Dart VM version:\s([^\s]+)(?:\s\((?:[^\)]+)\))?\s\(([^\)]+)\) on "(\w+)"',
   );
 
   // TODO: parse an actual `DateTime` here. Likely requires using pkg/intl
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -492,8 +492,9 @@ class DartdocResult {
 }
 
 class DartSdkInfo {
-  static final _sdkRegexp =
-      RegExp('Dart VM version:\\s([^\\s]+)\\s\\(([^\\)]+)\\) on "(\\w+)"');
+  static final _sdkRegexp = RegExp(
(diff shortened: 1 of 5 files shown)


kevmoo:
CC @pq

sigurdm:
I think this is triggered by: https://github.com/dart-lang/sdk/commit/8350f423ee928b095249e82a081209fc349290e3

That means the new info in the version string is the channel

lib/src/sdk_env.dart
```diff
@@ -492,8 +492,9 @@ class DartdocResult {
 }
 
 class DartSdkInfo {
-  static final _sdkRegexp =
-      RegExp('Dart VM version:\\s([^\\s]+)\\s\\(([^\\)]+)\\) on "(\\w+)"');
+  static final _sdkRegexp = RegExp(
+    r'Dart VM version:\s([^\s]+)\s\(([^\)]+)\)(\s\(([^\)]+)\))? on "(\w+)"',
```
sigurdm: I think you need non capturing groups, or the date will no longer be match[3].
```
r'Dart VM version:\s([^\s]+)\s\(([^\)]+)\)(?:\s\((?:[^\)]+)\))? on "(\w+)"'
```

isoos: We'll also need to handle both the old and the new output (e.g. try the longer/new regex first, then fallback to the old one), or, if this is a breaking change, we'll need to be specific about it.

kevmoo: I do handle the optional 2nd item – there's a ? in there that's hard to read.

This works on the bots (without the new change) and on my machine (with the change).

A test would be good, though.

kevmoo:
Should I change this PR or land it? Linter is broken. 😢 

@sigurdm

sigurdm:
@jonasfj could you sanity check my update?

test/sdk_env_test.dart
```diff
@@ -13,5 +13,17 @@ void main() {
         'Dart VM version: 2.0.0-dev.49.0 (Wed Apr 18 20:41:36 2018 +0200) on "macos_x64"';
     final sdkInfo = DartSdkInfo.parse(version);
     expect(sdkInfo.version, Version.parse('2.0.0-dev.49.0'));
+    expect(sdkInfo.dateString, 'Wed Apr 18 20:41:36 2018 +0200');
+    expect(sdkInfo.platform, 'macos_x64');
+  });
+
+  test('parsing SDK version  new style', () {
```
jonasfj: Nit: Extra space on title..

kevmoo:
Thanks, @sigurdm !

PR #634, opened 2020-04-14
Tag null-safety compliant packages

sigurdm:


pubspec.yaml
```diff
@@ -43,3 +43,16 @@ dev_dependencies:
 
 executables:
   pana:
+
+dependency_overrides:
```
isoos, line 47: Does this mean we can't release pana for a while if this gets merged?

jonasfj: We can add a git override in `pub-dev` for `pana` and these two dependencies too. But yes, that's painful.

lib/src/tag_detection.dart
```diff
@@ -495,6 +533,54 @@ class Sdk {
   static List<Sdk> knownSdks = [dart, flutter];
 }
 
+/// A package is said to null-safety compliant if:
+///
+/// The package has opted-in by specifying a lower dart sdk bound >= 2.10
+/// No libraries in the package opts out of the null-safety enabled language
+/// version.
+/// All (non-dev) dependencies in the latest version resolvable by pub are
+/// fully null-safety-compliant by this definition.
```
jonasfj, line 542: Maybe a little markdown can make this prettier :D

lib/src/tag_detection.dart
```diff
@@ -701,4 +762,15 @@ class Tagger {
     }
     return result;
   }
+
+  List<String> nullSafetyTags() {
+    final nullSafety = NullSafety(_packageGraph, _pubspecCache, _session);
+    final nullSafetyResult = nullSafety.findNullSafetyViolation(packageDir);
+    if (nullSafetyResult.hasPath) {
+      log.info(
+          '$packageDir is not null-safety compliant because of ${nullSafetyResult.path}');
```
jonasfj, line 771: Pretty-print?

lib/src/tag_detection.dart
```diff
@@ -701,4 +762,15 @@ class Tagger {
...
+      return [];
+    }
+    return ['is:null-safety'];
```
jonasfj, line 774: How about `is:null-safe`, we are going to surface it differnetly in the UI, anyways.

We could also do `is:null-safety-compliant`, but that just seems very long.

pubspec.yaml
sigurdm: Yes unfortunately it will mean that - I am trying to ping for a new analyzer version here: https://github.com/dart-lang/sdk/issues/37487

lib/src/tag_detection.dart
sigurdm: Sure - done :)

lib/src/tag_detection.dart
sigurdm: Done

lib/src/tag_detection.dart
sigurdm: Sure - I intuitively prefer that. Only I'm learning from @mit-mit that that is the wrong name for the feature.

mit-mit: I think I'm ok with `is:null-safe` here; the alternatives seem worse

Issue #590, opened 2019-12-13
surface explanation of runtime/platform tags not granted


jonasfj:
We should surface why a platform/runtime tag was not assigned. Especially, now that we have lovely import-traces to explain this.

isoos:
This was implemented.

PR #635, opened 2020-04-20
Add suggestions when a tag doesn't apply

sigurdm:


test/end2end/pub_server-0.1.4-2.json
```diff
@@ -85,6 +85,14 @@
     "isExperimentalVersion": false,
     "isPreReleaseVersion": false
   },
+  "suggestions": [
+    {
+      "code": "tagging.notCompatible",
+      "level": "hint",
+      "title": "Package is not null safe",
```
isoos, line 92: I believe we don't want to expose this to the users, at this point, until they cannot do anything about it. The best configuration could be to make this as a dynamic check and emit only when the SDK used for analysis already supports null-safety... wdyt?

lib/src/tag_detection.dart
```diff
@@ -640,140 +742,132 @@ class Tagger {
...
-        final pathResult = finder.findRuntimeViolation(lib);
-        if (pathResult.hasPath) {
-          log.info('$packageDir does not support runtime ${runtime.name} '
-              'because of the import path: ${pathResult.path}');
-          supports = false;
-          break;
+      tags.addAll(<String>[Runtime.nativeAot.name, Runtime.nativeJit.name]);
+    } else if (_topLibraries.isEmpty) {
+      Suggestion.hint(
+          SuggestionCode.noToplevelLibraries,
```
isoos, line 824: yay!

lib/src/tag_detection.dart
```diff
@@ -640,140 +742,132 @@ class Tagger {
...
-    final pathResult = declaredSdkViolationFinder.findPath(packageDir);
-    if (pathResult.hasPath) {
-      log.info(
-          '$packageDir does not support ${sdk.name} because of the package: ${pathResult.path}');
-      return false;
+  /// Adds tags for the Flutter platforms that this package supports to [tags].
+  ///
+  /// Adds [Suggestion]s to [suggestions] for platforms not supported.
+  void flutterPlatformTags(List<String> tags, List<Suggestion> suggestions) {
+    if (_isBinaryOnly) {
```
isoos, line 778: please comment why it is empty and why that case is ignored

sigurdm: Hmm - I think I was unsure what to write, postponed it and forgot ...

lib/src/tag_detection.dart
```diff
@@ -265,30 +304,35 @@ class _PackageGraph implements _DirectedGraph<String> {
...
+    }
     return path.dirname(filePath);
   }
 
+  String packageName(Uri uri) {
+    if (uri.scheme != 'package') {
+      // We only resolve package: uris.
+      throw ArgumentError('Trying to get the package name of $uri');
+    }
+    return uri.path.substring(0, uri.path.indexOf('/'));
```
isoos, line 324: `uri.pathSegments.first`?

sigurdm: Nice!
Done

lib/src/tag_detection.dart
```diff
@@ -78,27 +78,38 @@ import 'package:analyzer/dart/analysis/session.dart';
...
-
 abstract class _DirectedGraph<T> {
   Set<T> directSuccessors(T t);
 }
 
+/// Creates a suggestion indicating an issue, using [path] to give the location
+/// of the issue.
+typedef Explainer<N> = Suggestion Function(List<N> path);
+
+/// Returns an [Explainer] if node fullfills the predicate (if it violates some
```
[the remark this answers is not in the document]
sigurdm: These should all be private - Done.

The idea is that the path is not ready at the point in the search where we find the 'bad' node (because we prepend old paths if we encounter a known bad node via `_PathFinder._cache`).

natebosch:
How does a VM package avoid getting annoying warnings about this? Do we need to specifically "tag" every single published package to say that it doesn't support flutter web? Does that happen in the pubspec? What is the format, is it documented somewhere?

natebosch:
It looks like maybe this was intended to solve the issue raised in https://github.com/dart-lang/pana/issues/590 ?

Perhaps we can find a better way to surface this in the UI? We should not appear to "suggest" that every package needs to support the web, but that is what we are doing putting this un "Analysis Suggestions". Maybe we can change the rendering for the tagged platforms and have a greyed-out platform for the web, and when you mouse-over it it shows a tip about why?

jonasfj:
> Do we need to specifically "tag" every single published package to say that it doesn't support flutter web? Does that happen in the pubspec?

Packages cannot at this stage specify which tags they have, this is inferred based on the libraries imported from the primary library of a package.

For Flutter plugins tags are derived from the platforms listed in the pubspec under the `flutter.plugins.platforms` key.

Regular VM plugins need no do anything. This suggestion is just to highlight why a tag was not given. This logic will change anyways, as scoring is currently re-envisioned to be simpler..

natebosch:
> Regular VM plugins need no do anything. This suggestion is just to highlight why a tag was not given.

Can we change the UX? Can we have a section for "Platform Restriction Reasons" instead of "Suggestions"? This isn't something that we should suggest someone change.

sigurdm:
Yes - these messages were shoehorned into the current "suggestions"

In the new scoring model this will be surfaced differently as part of a section about platform support.

PR #643, opened 2020-05-27
InspectOptions.lockedVersions

isoos:


lib/src/package_analyzer.dart
```diff
@@ -54,6 +59,7 @@ class InspectOptions {
     this.isInternal = false,
     this.lineLength,
     this.analysisOptionsUri,
+    this.lockedVersions,
```
sigurdm, line 62: Could we give this an empty map as a default, that would save a null check later.

sigurdm:
Sorry for only seeing this now.

isoos:
> would be awesome with a test.

maybe as part of end2end tests?

isoos:
This is now obsolete I think.

PR #645, opened 2020-06-04
Avoid destructive pub get

sigurdm:


lib/src/sdk_env.dart
```diff
@@ -276,15 +276,33 @@ class ToolEnvironment {
...
+      return await fn(tempDir);
+    } finally {
+      if (tempDir.existsSync()) {
+        tempDir.deleteSync(recursive: true);
+      }
+    }
+  }
+
+  /// Runs `pub upgrade` on a stripped copy of [pubspec].
+  Future<ProcessResult> runUpgrade(
```
isoos, line 291: This is being called from `pub-dev`'s `dartdoc_runner.dart` too, before running `dartdoc`. Is copying the `pubspec.lock` enough to support that?

lib/src/sdk_env.dart
```diff
@@ -445,23 +461,13 @@ class ToolEnvironment {
...
-
-    return backup;
+  /// Returns a (shallow) copy of the underlying JSON map from [pubspec] with
+  /// dev_dependencies and dependency_overrides removed.
+  Map<dynamic, dynamic> _stripPubspecYaml(Pubspec pubspec) {
+    final json = Map.from(pubspec.toJson());
+    json.remove('dev_dependencies');
+    json.remove('dependency_overrides');
+    return json;
   }
```
jonasfj, line 471: Couldn't we also create a `pubspec.yaml` on the form:

`{"name": "dummy pana", "dependencies": {"$package": {"path": "/path/to/folder"}}`

Then use this to create the `.dart_tool/package_config.json` file, which I suppose is all we're doing here?
Information about outdated packages we can get from `pub outdated --json` which has `--no-dependency-overrides` and `--no-dev-dependencies`.

End result is probably the same, I'm not sure creating an empty pubspec is cleaner..

sigurdm: Yes! That's even better!

sigurdm: Though - I'd like to do the `pub outdated` later in another PR.

lib/src/sdk_env.dart
sigurdm: Not sure.

Are we using the resolution afterwards then this will probably not be enough.
If we just want the fact that it resolved then this should be enough. (Actually we probably do not want to copy a pubspec.lock file when running on a hosted package, but rely on an open-ended resolution).

lib/src/sdk_env.dart
jonasfj: If I was unclear I just meant this as a suggestion... Feel free to ignore or leave for later..

I'm not sure it's cleaner to make a dummy pubspec.. it's just more similar to what we do in global activate..

lib/src/sdk_env.dart
isoos: I'd rather change the local package analysis path to copy the entire package to a temp directory instead of using thing - otherwise nice - trick...

sigurdm: I don't think this is very much a trick. Not more than the old `_stripPubspecYaml` method.
If we copy the entire folder we still have to think about the interaction with`dartdoc_runner.dart` from `pub-dev`.
So the question remains:
Does `dartdoc_runner.dart` rely on the side-effects of running the pub-upgrade?

isoos: > Does dartdoc_runner.dart rely on the side-effects of running the pub-upgrade?

The reason we run this before `dartdoc` is to have a full package setup (so that `dartdoc` doesn't need to run it), in order to separate the timeouts of fetching the dependencies from the already-long dartdoc content generation. I don't recall the details, but for a while this also fixed an issue where `dartdoc` failed to get some Flutter dependencies properly.

> If we copy the entire folder we still have to think about the interaction with dartdoc_runner.dart from pub-dev.

My thinking was:
- let's leave this method as-is (after all, it is called run upgrade)
- for the local analysis path let's prepare the analysis with copying the package to a temp dir, and run every destructive modification on the temp copy.

sigurdm: I will leave the method as-is, so dartdocRunner can continue working.
But pana itself will not copy the entire directory.
Though I think moving forward we should let dartdoc-runner do its own pub upgrade.

Does it work for you?

lib/src/sdk_env.dart
sigurdm: I'll leave it for now

isoos:
This has been implemented.

Issue #639, opened 2020-05-06
packages using old flutter plugin format should only be able to support android/ios


jonasfj:
Plugins which has:
`flutter.plugin.androidPackage` or `flutter.plugin.pluginClass` are using the old plugin format.

We should consider such plugins as supporting only Android and iOS.

PR #646, opened 2020-06-09
Add tag-detection for old-style flutter plugins

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/639

### DIFF
--- a/lib/src/tag_detection.dart
+++ b/lib/src/tag_detection.dart
@@ -513,10 +513,8 @@ class _DeclaredFlutterPlatformDetector {
         return _FlutterPlatform.recognizedPlatforms.toSet();
       }
       final pluginMap = fields['flutter']['plugin'] as Map;
-
-      if (pluginMap['platforms'] is Map) {
-        final declaredPlatforms =
-            fields['flutter']['plugin']['platforms'] as Map;
+      final declaredPlatforms = pluginMap['platforms'];
+      if (declaredPlatforms is Map) {
         for (final platform in _FlutterPlatform.recognizedPlatforms) {
           if (declaredPlatforms.containsKey(platform.name)) {
             result.add(platform);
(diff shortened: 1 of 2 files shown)


lib/src/tag_detection.dart
```diff
@@ -501,15 +506,29 @@ class _DeclaredFlutterPlatformDetector {
...
       }
-      final declaredPlatforms = fields['flutter']['plugin']['platforms'] as Map;
-      for (final platform in _FlutterPlatform.recognizedPlatforms) {
-        if (declaredPlatforms.containsKey(platform.name)) {
-          result.add(platform);
+      final pluginMap = fields['flutter']['plugin'] as Map;
+
+      if (pluginMap['platforms'] is Map) {
+        final declaredPlatforms =
+            fields['flutter']['plugin']['platforms'] as Map;
```
isoos, line 519: Nit: `pluginMap['platforms'] as Map`

lib/src/tag_detection.dart
```diff
@@ -501,15 +506,29 @@ class _DeclaredFlutterPlatformDetector {
...
+          }
+        }
+      } else {
+        // Legacy style plugin:
+        if (pluginMap['androidPackage'] is String) {
+          result.add(_FlutterPlatform.android);
+        }
+        if (pluginMap['iosPrefix'] is String) {
+          result.add(_FlutterPlatform.ios);
         }
```
jonasfj, line 532: The old plugin format didn't support either ios or android, but required both to be supported.
And if I recall correctly then either `androidPackage` or `iosPrefix` was optional.
In either case, if either `androidPackage` or `iosPrefix` is present then we assume `ios` and `android` support.

```dart
if (pluginMap['androidPackage'] is String || pluginMap['iosPrefix'] is String) {
  result.add(_FlutterPlatform.android);
  result.add(_FlutterPlatform.ios);
}
```

lib/src/tag_detection.dart
sigurdm: Thanks!

lib/src/tag_detection.dart
sigurdm: Done - PTAL

PR #650, opened 2020-06-09
Implement scoring section 'supports dart 2'

sigurdm:


lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+    if (environment is YamlMap) {
+      final sdk = environment.nodes['sdk'];
+      if (sdk is YamlNode) {
+        span = sdk.span;
+      }
+    }
+
+    issues.add(
+      _Issue(
+        'The current sdk constraint does not allow any Dart 2 versions',
```
isoos, line 44: Nit: I'd like to see these messages with a sentence-ending `.`.

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+String _makeSummary(String introduction, List<_Issue> issues) {
+  return [
+    introduction,
+    if (issues.isNotEmpty) ...[
+      '',
+      if (issues.length <= 2)
+        ...issues
+      else ...[
+        'Found ${issues.length} issues. Showing the first two:',
+        ...issues.take(2).map((x) => x.toString()),
```
isoos, line 109: Nit: I think `.map((x) => x.toString())` is not strictly needed.

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+    // TODO(sigurdm):Implement rest of sections.
+  ]);
+}
+
+ReportSection _supportsDart2(Pubspec pubspec) {
+  final supportsDart2 = pubspec.supportsDart2;
+  final issues = <_Issue>[];
+
+  if (!supportsDart2) {
+    final environment = pubspec.toJson()['environment'] as Map;
```
isoos, line 33: We should probably expose `pubspec._inner.environment` instead.

lib/src/pubspec.dart
```diff
@@ -121,6 +128,16 @@ class Pubspec {
         'iosPrefix',
         'pluginClass',
       }.any((_inner.flutter['plugin'] as Map<String, Object>).containsKey);
+
+  /// 'true' if any dart2 versions of the sdk are supported.
+  bool get supportsDart2 {
```
isoos, line 133: I think this could reuse `SdkConstraintStatus`.

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+        span = sdk.span;
+      }
+    }
+
+    issues.add(
+      _Issue(
+        'The current sdk constraint does not allow any Dart 2 versions',
+        span: span,
+      ),
+    );
```
jonasfj, line 47: Maybe split this into two explanations:
 * Case 1, there is no SDK constraint.
 * Case 2, there is an SDK constraint, but it doesn't allow `2.x.y`

In  case 2 we can even include the SDK constraint in the message like:
```md
The current SDK constraint `$sdk` does not allow any Dart 2 SDK versions.
```

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  return ReportSection(
+      title: 'Package supports Dart 2',
+      maxPoints: 20,
+      grantedPoints: supportsDart2 ? 20 : 0,
+      summary: _makeSummary(
+          'Package gets 20 points if their dart sdk constraint allows Dart 2',
```
jonasfj, line 55: In this case wouldn't it be fair to make two kinds of messages depending on whether or not Dart 2 is supported?

So:
```
Package gets 20 points because the Dart SDK constraint `$sdk` is compatible with Dart 2.
```
or 
```
Package gets 0 points because the Dart SDK constraint `$sdk` is not compatible with Dart 2.
```
or
```
Package gets 0 points because it does not feature an SDK constraint and is not considered Dart 2 compatible.
```

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+      if (suggestion != null) suggestion
+    ].join('\n');
+  }
+}
+
+extension on SourceSpan {
+  /// An attempt to render [SourceSpan]s in a markdown-friendly way
+  String get markdown {
+    assert(sourceUrl != null);
+    return '<${p.prettyUri(sourceUrl)}>:${start.line + 1}:${start.column + 1}\n\n'
```
jonasfj, line 93: How about using:
```
'`<${p.prettyUri(sourceUrl)}>:${start.line + 1}`:${start.column + 1}\n\n'
```

Adding ` around the filename:line:column.

lib/src/create_report.dart
```diff
@@ -0,0 +1,113 @@
...
+        '```${highlight()}```\n';
```
jonasfj, line 94:
```suggestion
        '```\n${highlight()}\n```\n';
```

lib/src/create_report.dart
sigurdm: Thanks. Done

lib/src/create_report.dart
sigurdm: Done

lib/src/create_report.dart
sigurdm: Done

lib/src/create_report.dart
sigurdm: Good idea!
Done

lib/src/create_report.dart
sigurdm: Done

lib/src/create_report.dart
sigurdm: Done

sigurdm:
Thanks for the thorough reviews. PTAL!

lib/src/create_report.dart
sigurdm: I think the summary headline should always be the same, and then the details below explain what you did wrong

lib/src/pubspec.dart
sigurdm: Ah - true, that was already there!

lib/src/create_report.dart
sigurdm: done

lib/src/pubspec.dart
```diff
@@ -43,6 +48,8 @@ class Pubspec {
 
   Map<String, Dependency> get devDependencies => _inner.devDependencies;
 
+  Map get environment => _content['environment'] as Map;
```
jonasfj, line 51: nit: `(_content['environment'] as Map) ?? {}` might be nicer in the future...

sigurdm: Done

Issue #655, opened 2020-06-15
Trust declared platforms of flutter plugins


sigurdm:
Today the platform tags of a plugin is the intersection of the the package's and those of its dependencies.

We should just trust what the package declares when showing platform tags

We should only count scoring for those platforms that we analyze to work.

jonasfj:
Discussing this on video, it was suggested that we instead warn about issues like this in analyzer or during `pub publish`.

PR #657, opened 2020-06-18
Use given flutter tags

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/655


test/goldens/end2end/path_provider-1.6.11.json
```diff
@@ -0,0 +1,124 @@
...
+      "codeProblems": null
+    }
+  },
+  "stats": {
+    "analyzeProcessElapsed": 1234,
+    "formatProcessElapsed": 567,
+    "resolveProcessElapsed": 899,
+    "totalElapsed": 4567
+  },
+  "tags": [],
```
jonasfj, line 120: Isn't there something missing here?

lib/src/tag_detection.dart
```diff
@@ -502,6 +512,12 @@ class _DeclaredFlutterPlatformDetector {
 
   _DeclaredFlutterPlatformDetector(this._pubspecCache);
 
+  bool _isFlutterPlugin(Uri uri) =>
+      uri.scheme != 'package' ||
+      _declaredFlutterPlatforms(_pubspecCache.packageName(uri)) != null;
```
jonasfj, line 517: I suppose there is a need to distinguish between "_package_" and "_plugin without any supported platforms_", right?

lib/src/tag_detection.dart
```diff
@@ -818,11 +864,22 @@ class Tagger {
...
-          final violationResult = violationFinder._findPlatformViolation(lib);
-          if (violationResult != null) {
-            suggestions.add(violationResult);
+          if (!hasFoundNonPrunedViolation) {
+            final violationResult = violationFinder._findPlatformViolation(lib);
+            if (violationResult != null) {
+              suggestions.add(violationResult);
+              hasFoundNonPrunedViolation = true;
+            }
+          }
```
jonasfj, line 878: nit. Maybe it would be easier to read if we did this in two separate loops... or used map/where/any, like:

```dart
// Report only the first non-pruned violation as suggestion 
final firstNonPrunedViolation = _topLibraries
  .map(violationFinder._findPlatformViolation)
  .firstWhere((e) => e != null, orElse: () => null);
if (firstNonPrunedViolation != null) {
  suggestions.add(firstNonPrunedViolation);
}

// Tag is supported, if there is no pruned violations
final supports = _topLibraries
  .map(prunedViolationFinder._findPlatformViolation)
  .any((e) => e != null);
```

lib/src/tag_detection.dart
sigurdm: These are the "declared" platforms. A non-plugin has no declared platforms.

test/goldens/end2end/path_provider-1.6.11.json
sigurdm: Seems I can only regenerate this on linux... I think this might be fixed by: https://github.com/dart-lang/pana/pull/652..

Uploaded new version

lib/src/tag_detection.dart
sigurdm: And a plugin with no declared platforms would return the empty set.

PR #658, opened 2020-06-18
Improve sentence

mit-mit:
'Package not compatible with runtime flutter-native on android' reads better than 'Package not compatible with runtime flutter-native of android'


sigurdm:
I'll patch the golden files

PR #662, opened 2020-06-19
New scoring analysis

sigurdm:


lib/src/create_report.dart
```diff
@@ -71,6 +79,113 @@ environment:
...
+      : <_Issue>[];
+
+  return ReportSection(
+    title: 'Code follows recommended code style',
+    maxPoints: 20,
+    grantedPoints: (errors.isEmpty && warnings.isEmpty)
+        ? (formattingIssues.isEmpty && lints.isEmpty ? 20 : 10)
+        : 0,
+    summary: _makeSummary('''
+10 points: code has no errors and warnings.
```
isoos, line 107: Nit: extra line break and `*10 points:*` (`<em>` tag) + same for `20 points`

sigurdm: Done

PR #665, opened 2020-06-25
Report sections for documentation points.

isoos:


lib/src/dartdoc_analyzer.dart
```diff
@@ -28,3 +32,37 @@ Suggestion getDartdocRunFailedSuggestion([DartdocResult result]) {
...
+  final percent = total <= 0 ? 1.0 : documented / total;
+  final accepted = percent >= 0.2;
+  final granted = accepted ? 10 : 0;
+  final undocumented = total - documented;
+  return ReportSection(
+    title: documentationSectionTitle,
+    grantedPoints: granted,
+    maxPoints: 10,
+    summary: accepted
+        ? 'At least 20% of the public API has documentation comments.'
```
sigurdm, line 51: I'd like the first line to be invariant to whether the check passed or not, just explaining the scoring. Then a paragraph explaining the eventual problems.

In this case it would probably make sense to print the percentage even if it's above 20%.

lib/src/model.dart
```diff
@@ -1037,6 +1037,37 @@ class Report {
...
+          sections: sections.map(
+        (s) {
+          if (s != matched) {
+            return s;
+          }
+          return ReportSection(
+            title: s.title,
+            maxPoints: s.maxPoints + section.maxPoints,
+            grantedPoints: s.grantedPoints + section.grantedPoints,
+            summary: [s.summary.trim(), section.summary.trim()].join('\n\n'),
```
sigurdm, line 1065: Maybe we need to special case this a bit more.

The first paragraph of the section should explain the scoring rule of that section, followed by details of any issues.

lib/src/dartdoc_analyzer.dart
isoos: PTAL.

lib/src/model.dart
isoos: This points towards having a specific data structure for a section's generic explanations of the scoring rule, and a separate part for listing the issues. I'd rather start with this simpler concatenation, and as we evaluate the result, we can go back and restructure it?

sigurdm: SG.

I don't think we had anticipated the combination of sections, and I think with sandboxing that won't be needed.

lib/src/dartdoc_analyzer.dart
```diff
@@ -28,3 +32,43 @@ Suggestion getDartdocRunFailedSuggestion([DartdocResult result]) {
...
+ReportSection documentationCoverageSection({
+  @required int documented,
+  @required int total,
+}) {
+  final maxPoints = 10;
+  final ratio = total <= 0 ? 1.0 : documented / total;
+  final accepted = ratio >= 0.2;
+  final percent = ratio.toStringAsFixed(1);
+  final summary = StringBuffer();
+  summary.write('$documented out of $total ($percent %) '
```
sigurdm, line 47: I would still like the first paragraph to give the rule. (`*10 points*: 20% or more of the public api has dartdoc.` or similar) Then the following paragraphs can give details

isoos: Done. (also for the example check)

PR #687, opened 2020-07-06
Don't require flutter constraint

sigurdm:


lib/src/create_report.dart
```diff
@@ -668,6 +663,9 @@ Future<ReportSection> _multiPlatform(String packageDir, Pubspec pubspec) async {
       }
     } else {
       tagger.runtimeTags(tags, explanations);
+      final issues = explanations
```
isoos, line 666: not sure, just double-checking: is this `issues` masking another `issues` variable in score?

test/goldens/end2end/angular_components-0.10.0.json_report.md
```diff
@@ -41,6 +41,12 @@ See [package layout](https://dart.dev/tools/pub/package-layout#examples) guideli
 
 ### [~] 10/20 points: Supports 1 of 2 possible platforms (native, js)
 
+* Package not compatible with runtime native-aot
```
isoos, line 44: Is there a case where `native-aot` and `native-jit` differs? Can we emit only one of them with the tag `native`?

test/goldens/end2end/angular_components-0.10.0.json_report.md
```diff
@@ -41,6 +41,12 @@ See [package layout](https://dart.dev/tools/pub/package-layout#examples) guideli
...
+
+Because of the import of dart:html via the import chain `package:angular_components/angular_components.dart` → `package:angular_components/theme/module.dart` → `package:angular/angular.dart` → `package:angular/src/platform/browser/tools/tools.dart` → `package:angular/src/platform/browser/tools/common_tools.dart` → `package:angular/src/core/linker/component_factory.dart` → `package:angular/src/core/linker/view_ref.dart` → `dart:html`
+* Package not compatible with runtime native-jit
```
isoos, line 47: newline before this?

test/goldens/end2end/angular_components-0.10.0.json_report.md
mit-mit: I think `dart:developer` is the one one: https://dart.dev/guides/libraries

I would prefer if we generally just say "Package is not compatible with the Dart native platform", and then only if we have a specific check for `dart:developer` then for that check say "Package is not compatible with the Dart native platform in AOT mode"

jonasfj: Generally, we don't mention "AOT"-mode on pub.dev.. We just say that something is "native" if it works with "native-jit".

lib/src/create_report.dart
sigurdm: This was accidentally mixed in from https://github.com/dart-lang/pana/pull/689 and doesn't belong here.

Thanks!

test/goldens/end2end/angular_components-0.10.0.json_report.md
sigurdm: This will be handled in a separate PR.

test/goldens/end2end/angular_components-0.10.0.json_report.md
sigurdm: See https://github.com/dart-lang/pana/pull/689

PR #691, opened 2020-07-06
Nicer report when pub get fails

sigurdm:


lib/src/sdk_env.dart
```diff
@@ -315,8 +315,7 @@ class ToolEnvironment {
       workingDirectory: packageDir,
     );
     if (getResult.exitCode != 0) {
-      throw Exception(
-          '`pub get` exited ${getResult.exitCode}. StdErr: ${getResult.stderr}');
+      throw Exception('`pub get` failed: ${getResult.stderr}');
```
jonasfj, line 318: Can we not make a `class PubGetFailed implements Exception` thing... See `DartFmtFailed` in https://github.com/dart-lang/pana/pull/692

If we want to pass on `stderr` I think it's best guarded in triple quotes, and it would be nice if we avoid mixing this with other exceptions/errors (so as to avoid catch-all).

sigurdm: > Can we not make a class PubGetFailed implements Exception thing... See DartFmtFailed in #692

Can we converge on a single `ToolException` ?

> If we want to pass on stderr I think it's best guarded in triple quotes, and it would be nice if we avoid mixing this with other exceptions/errors (so as to avoid catch-all).

Done

Issue #690, opened 2020-07-06
dartfmt failed exit code 65


jonasfj:
When running `dart bin/pana.dart --source hosted elite_kit 0.0.1` pana crashes.

PR #692, opened 2020-07-06
Handle dartfmt crashes

jonasfj:


jonasfj:
Fixes https://github.com/dart-lang/pana/issues/690

lib/src/sdk_env.dart
```diff
@@ -500,3 +508,13 @@ String _join(String path, String binDir, String executable) {
   }
   return cmd;
 }
+
+/// Thrown when `dartfmt` fails to run.
+class DartFmtFailed implements Exception {
```
sigurdm, line 513: Let's call this `ToolException` so we can reuse it across tools?

jonasfj: Love it!

jonasfj:
Blocked on https://github.com/dart-lang/pana/pull/691

PR #697, opened 2020-07-07
Use bullet lists for import-paths

sigurdm:
Makes long paths easier to read/comprehend


test/goldens/end2end/stream-2.0.1.json_report.md
```diff
@@ -67,14 +67,24 @@ See [package layout](https://dart.dev/tools/pub/package-layout#examples) guideli
 Package not compatible with runtime native-aot
 </summary>
 
-Because of the import of dart:mirrors via the import chain `package:stream/stream.dart` → `package:stream/plugin.dart` → `package:rikulo_commons/mirrors.dart` → `dart:mirrors`
+Because:
+
```
jonasfj, line 71: nit: Can't we do without this extra whitespace?

lib/src/tag_detection.dart
```diff
@@ -722,7 +714,7 @@ class _NullSafetyViolationFinder {
               if (version < _firstVersionWithNullSafety) {
                 return (path) => Explanation(
                       'Package is not null safe',
-                      'Because $file is opting out in package ${_PackageGraph.formatPath(path)}',
+                      'Because:\n\n${_PackageGraph.formatPath(path)} where $file is opting out from null-safety.',
```
jonasfj, line 717:
```suggestion
                      'Because:\n${_PackageGraph.formatPath(path)} where $file is opting out from null-safety.',
```

test/goldens/end2end/stream-2.0.1.json_report.md
sigurdm: Done

lib/src/tag_detection.dart
sigurdm: Done

PR #698, opened 2020-07-07
Fix goldens

jonasfj:


jonasfj:
I think travis broke because of something dependencies... let's see if it passes now :D

PR #700, opened 2020-07-07
Make a table of outdated data

sigurdm:


lib/src/create_report.dart
```diff
@@ -503,13 +508,62 @@ Future<ReportSection> _trustworthyDependency(
...
+          ['Package', 'Constraint', 'Compatible', 'Latest'],
+          [':-', ':-', ':-', ':-'],
+          for (final package in outdated.packages)
+            if (pubspec.dependencies.containsKey(package.package))
+              [
+                '[${package.package}]',
+                constraint(pubspec.dependencies[package.package]),
+                package.upgradable?.version ?? '-',
+                package.latest?.version ?? '-',
+              ],
```
jonasfj, line 539: nit, one could also do map... But this is also fine :D
```suggestion
...outdated.packages
           .where((p) => pubspec.dependencies.containsKey(p.package))
           .map((p) => [
  '[${p.package}]',
  constraint(pubspec.dependencies[p.package]),
  p.upgradable?.version ?? '-',
  p.latest?.version ?? '-',
]),
```

sigurdm: Yeah - its approximately the same.

PR #706, opened 2020-07-07
Use --no-dependency-overrides and ignore dependencies which are not u…

jonasfj:
…pgradable, that is because they have been filtered out


lib/src/create_report.dart
```diff
@@ -553,11 +559,12 @@ Future<ReportSection> _trustworthyDependency(
               ],
           ['**Transitive dependencies**'],
           for (final package in outdated.packages)
-            if (!pubspec.dependencies.containsKey(package.package))
+            if (!pubspec.dependencies.containsKey(package.package) &&
```
sigurdm, line 562: Perhaps add a comment with a link to:
https://github.com/dart-lang/pub/issues/2552

jonasfj:

test/goldens/end2end/stream-2.0.1.json_report.md
```diff
@@ -135,40 +135,40 @@ Dependencies
...
+|[`http_parser`]|-|3.1.4|3.1.4|
+|[`meta`]|-|1.2.1|1.2.1|
+|[`mime`]|-|0.9.6+3|0.9.6+3|
+|[`pedantic`]|-|1.9.1|1.9.1|
+|[`source_span`]|-|1.7.0|1.7.0|
+|[`string_scanner`]|-|1.0.5|1.0.5|
+|[`term_glyph`]|-|1.1.0|1.1.0|
+|[`typed_data`]|-|1.2.0|1.2.0|
 
 To reproduce run `pub outdated --no-dev-dependencies --up-to-date`.
```
sigurdm, line 155: Oh one more thing.
We should also update the reproduction command

PR #711, opened 2020-07-07
Leave out table when there are no dependencies

sigurdm:


jonasfj:
Can we hold off on merging this... it'll conflict with stuff in my pipeline..

jonasfj:
I'll steal this...

PR #723, opened 2020-07-10
Clarify issues from prerelease platforms

sigurdm:


lib/src/create_report.dart
```diff
@@ -778,46 +778,67 @@ Future<ReportSection> _multiPlatform(String packageDir, Pubspec pubspec) async {
...
+        List<String> tags,
+        List<Explanation> explanations) {
+      final unofficialExplanations = explanations
+          .where((e) => e.tag != null && !tagNames.containsKey(e.tag));
+      final officialExplanations = explanations
+          .where((e) => e.tag == null || tagNames.containsKey(e.tag));
+      final paragraphs = [
+        if (officialExplanations.isNotEmpty)
+          _RawParagraph(
+              'Consider supporting multiple platforms. Points were deducted due to the following issues:\n'),
```
mit-mit, line 793: Perhaps "issues" is a bit harsh? Maybe just omit it, so "Points were deducted due to the following:" ?

sigurdm: Done

PR #731, opened 2020-07-17
Update README.md

AndroidNils:
I think it should be "-100 Points", right? Because you lose 100 points, if parsing `pubspec.yaml` isn't possible.


sigurdm:
Good catch.

But we actually need to rewrite this section entirely to reflect the new scoring.

AndroidNils:
So should we merge or close this?

sigurdm:
Might as well merge it.

Thanks for the contribution!

PR #738, opened 2020-07-21
Dont explain aot

sigurdm:


lib/src/create_report.dart
```diff
@@ -845,6 +845,10 @@ Future<ReportSection> _multiPlatform(String packageDir, Pubspec pubspec) async {
         'runtime:web': 'js',
       };
 
+      // We don't want the native-aut runtime to be explained here.
+      explanations.removeWhere(
+          (explanation) => explanation.tag == 'runtime:native-aot');
```
isoos, line 850: nit: would be nice to reference this via a const value.

sigurdm: Done

PR #740, opened 2020-07-23
Remove bulk.dart and mini_sum.dart

isoos:
We'll remove most of the internal data that is tracked/aggregated in these utilities.

I reckon that @kevmoo was the main user of these at one point, I hope it is not used anymore.


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 **BREAKING CHANGES:**
 
 * Removed `Summary.suggestions`.
+* Removed `bulk.dart` and `mini_sum.dart`.
```
kevmoo, line 6: things in `lib/src` don't need to be documented.

Maybe mention `bin/bulk.dart`

PR #741, opened 2020-07-24
Remove Maintenance

isoos:


isoos:
> We also have to update the README (doesn't have to be this commit).

Yes, I'm planning to do that at the end of the cleanup.

PR #751, opened 2020-07-28
Simpler bin/pana CLI

isoos:


bin/pana.dart
```diff
@@ -32,12 +32,9 @@ final _parser = ArgParser()
       help: 'The server that hosts <package>.', defaultsTo: defaultHostedUrl)
   ..addOption('line-length',
       abbr: 'l', help: 'The line length to use with dartfmt.')
-  ..addOption('verbosity',
-      help: 'Configure the details in the output.',
-      allowed: ['compact', 'normal', 'verbose'],
-      defaultsTo: 'normal')
-  ..addFlag('scores', help: 'Include scores in the output JSON.')
-  ..addFlag('report', help: 'Print markdown report instead of JSON.')
+  ..addFlag('hosted', help: 'Shortcut to `--source hosted`.')
```
sigurdm, line 35:
```suggestion
  ..addFlag('hosted', help: 'Shortcut to `--source hosted`.', negatable = false)
```

sigurdm, line 35: Can we make this one be the default? And hide the  '--source' option, just keeping it for backwards compatibility?

README.md
```diff
@@ -37,10 +37,8 @@ Options:
       --hosted-url      The server that hosts <package>.
                         (defaults to "https://pub.dev")
   -l, --line-length     The line length to use with dartfmt.
-      --verbosity       Configure the details in the output.
-                        [compact, normal (default), verbose]
-      --[no-]scores     Include scores in the output JSON.
-      --[no-]report     Print markdown report instead of JSON.
+      --[no-]hosted     Shortcut to `--source hosted`.
+      --[no-]verbose    Print the full JSON output instead of the markdown report.
```
sigurdm, line 41: Should we call this `--json` or `--machine`?

isoos: We already have `--json` for the logs... which should be probably reused for this too (if you want json logs, why would you want non-json reports)?

isoos:
PTAL, I have also updated/fixed the first part of the help message.

PR #752, opened 2020-07-29
Move dart:io code from Pubspec

kevmoo:
Allows importing of this code via web
Needed for tools


isoos:
Curious: what kind of tool would import this (especially as we have no obligation of not breaking it)? Asking because I'm planning a few more, potentially breaking changes and/or cleanup for the 0.14.0 release...

If this is something that is expected to be used by other tools, maybe it would be better to refactor into a shared package?

Issue #92, opened 2017-09-13
Detect multiple licenses


isoos:
Examples:
- https://github.com/dart-lang/dart-samples/blob/master/LICENSE
- multiple license files at root
- Use https://spdx.org/licenses/ classifications.

kevmoo:
@isoos – do we have examples of this for published packages? Just curious...

isoos:
The updated package stats tool revealed that we have only a few packages that include multiple license files, many of the seem to include it by accident:

```
    "moreThanOne": {
      "sunmi": 3,
      "api": 2,
      "architecture_component": 2,
      "architecture_generator": 2,
      "fluent_validation": 2,
      "flute_music_player": 2,
      "flutter_ffmpeg": 2,
      "hello_example": 2,
      "hex": 2,
      "horizontal_select": 2,
      "omniverse": 2,
      "persian_utils": 2,
      "rondy_bottom_navigation_bar": 2,
      "save_image": 2,
      "spotify_playback": 2,
      "testhaipham_example": 2,
      "tinylog": 2,
      "widgets_visibility_provider": 2,
      "yui": 2
    }
```
(tool: https://github.com/dart-lang/pub-dev/pull/3900)

I think it is not worth to detect multiple files, rather the package should follow a predefined layout, where the license file selection is clearly defined.

kevmoo:
I completely agree here. Keep it simple. Absolutely not worth the complexity

jonasfj:
Your analysis is missing `package:yaml`, but I agree, a single LICENSE file should do.

isoos:
> Your analysis is missing package:yaml

Yeah, we were not detecting the `-license.txt` files. However, I'd argue that the package should have a single entry point where I can understand the (composite) license easily, without having to look whether there are additional files. That file could be a single (common) license block, or a description of what is the effective license of the package.

isoos:
Closing, as we've settled on detecting only a single license.

PR #755, opened 2020-07-31
Return only a single license file from the analysis.

isoos:
https://github.com/dart-lang/pana/issues/92


lib/src/license.dart
```diff
@@ -13,14 +13,14 @@ import 'package:path/path.dart' as p;
 import 'download_utils.dart';
 import 'model.dart';
 
-Future<List<LicenseFile>> detectLicensesInDir(String baseDir) async {
+Future<LicenseFile> detectLicensesInDir(String baseDir) async {
   final rootFiles = await Directory(baseDir).list().toList();
   final licenseCandidates = rootFiles
       .whereType<File>()
       .where(_isLicenseFile)
       .take(5) // only the first 5 files are considered
```
sigurdm, line 21: Could we change this to 1?

isoos:
> My only concern is that we might want to prioritize the license files

Good point, I'll add that (similar to the ones we have for readme, changelog and example).

lib/src/license.dart
```diff
@@ -11,55 +11,36 @@ import 'dart:io';
...
-  final licenses = await Future.wait(licenseCandidates.map(
-    (File file) {
-      final relativePath = p.relative(file.path, from: baseDir);
-      return detectLicenseInFile(file, relativePath: relativePath);
-    },
-  ));
-
-  licenses.sort((l1, l2) => Comparable.compare(l1.path, l2.path));
-  return licenses;
+Future<LicenseFile> detectLicensesInDir(String baseDir) async {
```
jonasfj, line 17: Rename the method if it's not plural anymore

lib/src/license.dart
```diff
@@ -11,55 +11,36 @@ import 'dart:io';
...
+  for (final candidate in licenseFileNames) {
```
jonasfj, line 18: Historically, the `pub` tool has a allow license files with many different names... `pana` and `pub.dev` needs to continue to support such old packages.

But I'm fine with saying that:
 * if there is a `LICENSE` file, then it is always the license we show, and,
 * if there is no `LICENSE` file, then we reject the package when published.

Note, that for existing packages, the old license extraction logic needs to be preserved, does it not?

isoos: > the old license extraction logic needs to be preserved

With the new rules we'll lose 4 licenses detected out of 10k+ (latest stable versions):
- `LICENSE.html` (2 instances)
- `LICENSE.GPLv3`
- `LICENSE.MIT`

Supporting and hosting a free-form `.html` is not something we want to do or encourage.

The remaining `LICENSE.<name>` is debatable - do we want to keep supporting them?

isoos: Btw. the current logic now follows the same selection precedence as we have in https://github.com/dart-lang/pub-dev/pull/3908

PR #779, opened 2020-08-26
Classify asset-only packages as multi-platform.

### LINKED ISSUE #604: Tag and analysis failure for packages that have no dart files
There are packages with resources only, e.g. `package:bulma_min`.

Updated: fixed package name.

### LINKED ISSUE #778: Platform identification fails for pure asset packages
E.g.: https://github.com/flutter/cupertino_icons is not identified. I suggest these (packages with no Dart code) are identified as supporting all platforms.

isoos:
- Fixes #604.
- Fixes #778.


lib/src/tag_detection.dart
```diff
@@ -806,15 +813,24 @@ class Tagger {
...
+      topLibraries,
+    );
   }
 
   void sdkTags(List<String> tags, List<Explanation> explanations) {
     try {
-      if (_isBinaryOnly) {
-        tags.add('sdk:dart');
+      if (_isAssetOnly) {
+        tags.addAll(_Sdk.knownSdks.map((s) => s.tag));
```
jonasfj, line 831: What if the package has an SDK constraint on `flutter`, or a dependency on something else that has an SDK constraint on flutter..

I agree, we don't need to look at the file in the dependenices because those are not imported.

isoos: Good point, we'll need to run the declared platform and sdk finders to only surface the ones that are declared.

isoos: Fixed. PTAL.

jonasfj (approved):

Issue #677, opened 2020-06-30
Add scoring section for null-safety compatibility


jonasfj:
@mit-mit, how many points do you think we should give for null-safety migration? 20 points?

**Blocked until** a migration guide is out and a pre-release of Dart advocates publishing null-safety migrated packages.

mit-mit:
I think I'd like to repurpose the Dart 2 section and it's 20 points for nulls safety.

sigurdm:
We already took out the dart 2 section for the documentation

mit-mit:
Ah, got it. Well, then we just add a new 20 points section. We can go higher with the new system!

mit-mit:
A concrete check that I'd like to see here is for packages that declare support for null safety (via SDK constraint), but don’t support sound null safety during analysis

sigurdm:
Do I understand the scoring here? One would get full points for either:

* Being a pre-null-safety package
* Being a package that opts in, and with all dependencies resolvable to null-safety-opted in packages (with no opt-outs in files).

But no points for opting in, but having not-opted-in files

leafpetersen:
Is this still planned?

mit-mit:
Yes, I believe so!

@sigurdm, can you elaborate on what you mean by being a "pre-null-safety" package? My idea was that you'd only get these 20 points if you support null safety.

sigurdm:
> @sigurdm, can you elaborate on what you mean by being a "pre-null-safety" package? 

I interpreted your statement: "A concrete check that I'd like to see here is for packages that declare support for null safety (via SDK constraint), but don’t support sound null safety during analysis" that we would give points to both packages that do not claim null-safety support in their language version, **and** packages  that claim support without having non-null-safety dependencies.

> My idea was that you'd only get these 20 points if you support null safety.

I guess that mean only the second category would get the points.

mit-mit:
> I guess that mean only the second category would get the points.

Yes, that is my suggestion.

PR #797, opened 2020-11-03
Add null-safety report (without points)

sigurdm:
Fixes https://github.com/dart-lang/pana/issues/677


lib/src/model.dart
```diff
@@ -477,6 +477,7 @@ abstract class ReportSectionId {
   static const dependency = 'dependency';
   static const documentation = 'documentation';
   static const platform = 'platform';
+  static const nullSafety = 'nullSafety';
```
isoos, line 480: let's keep the ids lowercase, and use `null-safety` or `null-safe` here.

lib/src/create_report.dart
```diff
@@ -755,6 +756,83 @@ Future<ReportSection> _trustworthyDependency(PackageContext context) async {
...
+            suggestion: 'Run `pub get` for more information.')
+      ],
+      0,
+      maxPoints,
+      _Status.bad,
+    );
+  }
+  return ReportSection(
+    title: 'Package supports null-safety',
+    maxPoints: 20,
```
isoos, line 824: maxPoints: 0?

sigurdm: Thanks!!!

lib/src/create_report.dart
```diff
@@ -755,6 +756,83 @@ Future<ReportSection> _trustworthyDependency(PackageContext context) async {
...
+                  suggestion: 'Try running `pub outdated --mode=null-safety`.\n'
+                      'Be sure to read the [migration guide](https://dart.dev/null-safety/migration-guide).')
+            ],
+            0,
+            maxPoints,
+            _Status.bad);
+      }
+    } else {
+      subsection = _Subsection(
+          'Package does not opt in to  null-safety',
```
isoos, line 798: Nit: extra space

lib/src/create_report.dart
```diff
@@ -755,6 +756,83 @@ Future<ReportSection> _trustworthyDependency(PackageContext context) async {
...
+          [
+            _Issue(
+              'Package version constraint lower bound is below 2.12.0.',
```
isoos, line 801: nit: can we use a const for `2.12.0`, maybe reuse `_firstNullSafetyVersion` for it?

lib/src/create_report.dart
```diff
@@ -755,6 +756,83 @@ Future<ReportSection> _trustworthyDependency(PackageContext context) async {
...
+                  'Consider [migrating](https://dart.dev/null-safety/migration-guide).',
+            )
+          ],
+          0,
+          maxPoints,
+          _Status.soso);
+    }
+  } else {
+    subsection = _Subsection(
+      'Supports 0 of 2 possible platforms (native, js)',
```
isoos, line 812: This feels strange here. Maybe "Unable to detect null safety"?

lib/src/model.dart
```diff
@@ -477,6 +482,24 @@ abstract class ReportSectionId {
   static const dependency = 'dependency';
   static const documentation = 'documentation';
   static const platform = 'platform';
+  static const nullSafety = 'null-safety';
+}
+
+enum ReportStatus {
```
isoos, line 488: As it becomes part of the public API, let's review the naming a bit, please consider:
- `miss`, `failed` (instead of `bad`)?
- `mixed`, `partial` (instead `soso`)?
- `pass`, `ok` (instead of `good`)?

isoos (commented):
lgtm in general, let's give the naming a short a reconsideration and it is good to go (even with the current names, I just wanted to have a quick review on it).

lib/src/model.dart
sigurdm: Yeah - I guess these sound more professional ;)

Fixed

isoos (approved):

lib/src/model.dart
```diff
@@ -477,6 +482,24 @@ abstract class ReportSectionId {
...
+  partial,
+  @JsonValue('passed')
+  passed,
+}
+
+ReportStatus summarizeStatuses(Iterable<ReportStatus> ss) {
+  return ss.fold(ReportStatus.passed, minStatus);
+}
+
+ReportStatus minStatus(ReportStatus a, ReportStatus b) {
```
jonasfj, line 501: nit: this code is not obvious, neat, but... could it have a comment?

lib/src/package_analyzer.dart
```diff
@@ -38,6 +39,7 @@ class InspectOptions {
     this.isInternal = false,
     this.lineLength,
     this.analysisOptionsUri,
+    this.nullSafetySectionEnabledBeforeOptin = true,
```
jonasfj, line 42: Shouldn't it default to `false` for now? or are we setting that in `pub.dev`... perhaps it's better to not give it a default value?

lib/src/create_report.dart
```diff
@@ -745,14 +749,88 @@ Future<ReportSection> _trustworthyDependency(PackageContext context) async {
...
+            maxPoints,
+            ReportStatus.passed);
+      } else {
+        subsection = _Subsection(
+            'Package declares support for null-safety, but there are issues',
+            [
+              ...explanations.map(_explanationToIssue).toList(),
+              _Issue('For more information',
+                  suggestion: 'Try running `pub outdated --mode=null-safety`.\n'
+                      'Be sure to read the [migration guide](https://dart.dev/null-safety/migration-guide).')
```
jonasfj, line 793: This will no longer give the full picture..

Let's add a TODO to explain it here...

jonasfj (approved):

lib/src/package_analyzer.dart
sigurdm: I'm ok with false.

Done.

lib/src/create_report.dart
sigurdm: Done

lib/src/model.dart
sigurdm: Done

PR #803, opened 2020-11-16
Tag packages with is:null-safe regardless whether the current SDK support

isoos:


jonasfj (approved):
cc @sigurdm in case you recall why we did this :)

sigurdm (approved):

Issue #804, opened 2020-11-19
Refine null-safety analysis to only consider imported libraries from dependencies


jonasfj:

sigurdm:
Do we start in all libraries from the root - or just from the "primary libraries" ?

jonasfj:
I think we would start from any library in `lib/` (maybe excluding `lib/src/` if not imported/exported from `lib/`).

mit-mit:
One example that is hitting this: Flutter plugins (e.g. https://pub.dev/packages/url_launcher/versions/6.0.0-nullsafety.1), which are currently not tagged as supporting null safety:

```
> url_launcher that depends on:
> url_launcher_windows that depends on:
> flutter that depends on:
> sky_engine where ui/ui.dart is opting out from null-safety.
```

isoos:
Another example reported here: https://github.com/dart-lang/pub-dev/issues/4277

PR #807, opened 2020-11-19
Use library import graph to determine null-safety

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/804


lib/src/null_safety.dart
```diff
@@ -6,7 +6,7 @@ import 'package:pub_semver/pub_semver.dart';
 
 bool isNullSafety(Version version, String packageName) =>
```
jonasfj, line 7: Can we document what `packageName == null` means=

lib/src/tag_detection.dart
```diff
@@ -983,23 +919,90 @@ class Tagger {
...
+          final optOutViolationFinder = _PathFinder<Uri>(
+            LibraryGraph(_session, runtime.declaredVariables),
+            (library) {
+              final unit = _parsedUnitFromUri(_session, library);
+              if (unit == null) return null;
+              final languageVersionToken = unit.languageVersionToken;
+              if (languageVersionToken == null) return null;
+              // dart:ui has no package name. So we cannot trivially look it up in
+              // the allowed experiments. We just assume it is opted in.
+              if (library == Uri.parse('dart:ui')) return null;
```
jonasfj, line 968: Let's make this assumption for all `dart:` libraries, let's assume they are opted-in and have no langauge-version-markers.

Maybe we can say that if `!library.isScheme('package')` then we `return null`, that way we never have to deal `package = null` (I assume it's only null for `dart:` libraries).

sigurdm: Good point
Done!

lib/src/tag_detection.dart
```diff
@@ -693,82 +693,40 @@ class _Sdk {
...
+  /// This is:
+  ///  * _primaryLibrary, if one exists,
+  ///  * _publicLibraries, if no _primaryLibrary exists,
+  ///  * all libraries, if no _publicLibraries exists.
+  ///
+  /// TODO(jonasfj): We should be using [_primaryLibrary] (with fallback to
+  ///     [_publicLibraries]) most places instead, and remove this property.
+  ///     If a package has no public libraries, then we should perhaps avoid
+  ///     assigning any tags or assign all tags.
   final List<Uri> _topLibraries;
```
jonasfj, line 718: This might be orthogonal, but `_topLibraries` seems like a bad name...
For null-safety detection we certainly want to look at "all public libraries".

For platform tag detection we only want to be looking at:
 * the primary library, or
 * the public libraries (if there is no single primary library).
 * MAYBE: all libraries if there is no public libraries -- I'm increasingly leaning towards the idea that we should look at these until someone imports from another library (possibly in another package).

So if a package has no public library, maybe we should just give it all tags or no tags. I'm guessing it's most correct to give it all tags, because it's most likely:
 * (i) A tool that generates code or does other useful things, or,
 * (ii) A package containing private libraries being used by other packages.

In the case of (ii) there is no harm in showing the package.. It's rather useless because it has no public libraries. And of the author doesn't want to pollute search results, the unlisted flag is there for this.

jonasfj:
I did a few fixes, but we probably still have tests failing :D

jonasfj (approved):

PR #812, opened 2020-12-02
latest analyzer

kevmoo:


pubspec.yaml
```diff
@@ -4,10 +4,10 @@ version: 0.14.10-dev
 homepage: https://github.com/dart-lang/pana
 
 environment:
-  sdk: '>=2.8.0 <3.0.0'
+  sdk: '>=2.10.4 <3.0.0'
```
isoos, line 7: `pub.dev` is running on `2.10.0`. Can we have 2.10.0 as minimum requirement here or is there something in 2.10.4 that we need to depend on?

test/goldens/end2end/_dummy_pkg-1.0.0-null-safety.0.json
```diff
@@ -23,32 +23,7 @@
     "name": "BSD",
     "url": "https://github.com/dart-lang/pub-dev/blob/master/LICENSE"
   },
-  "pkgResolution": {
```
isoos, line 26: Such a huge difference seems to be really off.

Not sure if this is the new version or a different SDK, but CI and end2end tests are on `2.12.0-0.0.dev`. Can you regenerate them with that SDK?

sigurdm (approved):
Yes, if we regenerate golden files with 2.12.0-0.0.dev and relax the lower sdk bound, I am also go.
LGTM

isoos:
@kevmoo Not sure what's missing in your env, I did an alternative PR: https://github.com/dart-lang/pana/pull/813

PR #815, opened 2020-12-15
Deprecate methods in ToolEnvironment

isoos:
- Also updated end2end tests for CI.
- Also removed unused method that was not exposed as public API.


sigurdm (approved):
LGTM

isoos:
Let's also publish the new version!

sigurdm:
Published!

PR #831, opened 2021-02-16
run `flutter` with `--no-version-check`

sigurdm:
This saves a bit of time on each invocation.


jonasfj (approved):

lib/src/utils.dart
```diff
@@ -19,27 +19,26 @@ Stream<String> _byteStreamSplit(Stream<List<int>> stream) =>
...
-    executable,
-    arguments,
+    arguments.first,
+    arguments.skip(1).toList(),
     workingDirectory: workingDirectory,
     environment: environment,
   );
 }
 
 Future<ProcessResult> runProc(
```
isoos, line 33: This is a breaking change, and should get mentioned in the changelog (IIRC we do use it in pub-dev).

sigurdm: Ah good point. I'll do a follow-up preparing for release.

Issue #827, opened 2021-02-05
Receiver: null Tried calling: where(Closure: (CodeProblem) => bool)


jonasfj:
```
Receiver: null Tried calling: where(Closure: (CodeProblem) => bool)
at Object.noSuchMethod (native)
at _analyzePackage (package:pana/src/create_report.dart:186)
at _staticAnalysis (package:pana/src/create_report.dart:104)
at createReport (package:pana/src/create_report.dart:60)
at PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:179)
at PackageAnalyzer.inspectPackage.<fn>.<fn> (package_analyzer.dart:71)
at PackageAnalyzer.inspectPackage.<fn>.<fn> (unknown location)
at withTempDir (package:pana/src/download_utils.dart:245)
at PackageAnalyzer.inspectPackage.<fn> (package_analyzer.dart:68)
at _rootRun (native)
at _CustomZone.run (native)
at _runZoned (native)
at runZoned (native)
at withLogger (package:pana/src/logging.dart:13)
at PackageAnalyzer.inspectPackage (package:pana/src/package_analyzer.dart:66)
```

jonasfj:
Updated:
```
pub.analyzer.pana: Failed (v0.14.10) - pointycastle/3.0.0-nullsafety.2

Error:
    NoSuchMethodError: The method 'where' was called on null.
    Receiver: null
    Tried calling: where(Closure: (CodeProblem) => bool)

Stack:
    #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
    #1      _analyzePackage (package:pana/src/create_report.dart:186:14)
    <asynchronous suspension>
    #2      _staticAnalysis (package:pana/src/create_report.dart:104:32)
    #3      createReport (package:pana/src/create_report.dart:60:11)
    <asynchronous suspension>
    #4      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:179:21)
    <asynchronous suspension>
    #5      PackageAnalyzer.inspectPackage.<anonymous closure>.<anonymous closure> (package:pana/src/package_analyzer.dart:71:22)
    <asynchronous suspension>
    #6      PackageAnalyzer.inspectPackage.<anonymous closure>.<anonymous closure> (package:pana/src/package_analyzer.dart)
    #7      withTempDir (package:pana/src/download_utils.dart:245:20)
    <asynchronous suspension>
[...]

PR #836, opened 2021-02-19
Survive analyzer crash

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/827


lib/src/package_analyzer.dart
```diff
@@ -142,11 +142,10 @@ class PackageAnalyzer {
     if (pkgResolution != null) {
       List<CodeProblem> analyzerItems;
       if (dartFiles.isNotEmpty) {
-        analyzerItems = await context.staticAnalysis();
-      } else {
-        analyzerItems = <CodeProblem>[];
```
isoos, line 147: I think this part may be important, but may not be anymore....

sigurdm: We have a null-check just below, and don't use the variable after that. I think it is all right.

lib/src/package_analyzer.dart
```diff
@@ -142,11 +142,11 @@ class PackageAnalyzer {
...
+        try {
+          analyzerItems = await context.staticAnalysis();
+          // ignore: empty_catches
+        } on ToolException {}
```
isoos, line 148: How about the following?
```
} on ToolException catch (e) {
  errors.add(messages.runningDartanalyzerFailed(usesFlutter, e.message));
}
```

isoos (approved):

isoos (approved):
Approval pending on the test fixes + changelog entry.

lib/src/package_analyzer.dart
sigurdm: Done

isoos (approved):

PR #847, opened 2021-03-12
Updated set of end2end test packages.

isoos:
I've added a small note on why that package and/or version has been selected to be part of the end2end suite:
- If we want to change it in the future, we shall have at least a starting point whether a given test is still considered useful.
- If we find a smaller package for that goal, replacing it would result faster running tests.


jonasfj (approved):

sigurdm (approved):
Nice! I have needed these comments before!

PR #851, opened 2021-03-22
Updated analysis_options.yaml for Flutter packages.

isoos:


lib/src/analysis_options.dart
```diff
@@ -45,6 +45,7 @@ linter:
     - avoid_empty_else
     - avoid_init_to_null
     - avoid_return_types_on_setters
+    - avoid_web_libraries_in_flutter
```
sigurdm, line 48: Could we fetch this from the current flutter installation?
Perhaps: 
```
File(p.join(Platform.environment['FLUTTER_ROOT'], 'packages', 'flutter',
          'lib', 'analysis_options_user.yaml'))
      .readAsStringSync()```

isoos: Maybe... but we won't have an easy test for it.

sigurdm: I wonder how hard it would be to have flutter enabled on CI?

isoos: We could use https://github.com/marketplace/actions/flutter-action
Or, alternatively, always download the latest from git

jonasfj: > Maybe... but we won't have a cheap test for it.

What do you mean, CI or locally?

-------

For CI, it would great to add `flutter-action`.

In production, we always a have a flutter SDK installed, so using the `analysis_options_user.yaml` file relative to `FLUTTER_ROOT` seems fine to me.

I guess locally we won't be able to run tests without setting `FLUTTER_ROOT` env var.

isoos:
#856 is a better way going forward, closing this

PR #856, opened 2021-04-16
New analysis-options.yaml handling.

isoos:
- Using `analysisOptionsYaml` instead of `analysisOptionsUri` (if `pub.dev` wants to pin the `pedantic` version).
- If not specified, falls back to (a) `analysis_options_user.yaml` inside the specified Flutter SDK (or on GitHub) or (b) the latest `pedantic` version (on GitHub).
- removes `package:resource` dependency
- allows the same passthrough options as the old version


lib/src/analysis_options.dart
```diff
@@ -3,79 +3,102 @@
...
+    }
+  }
+
+  // try to load latest from github
+  if (_cachedFlutterOptionsOnGithub != null) {
+    return _cachedFlutterOptionsOnGithub;
+  }
+  try {
+    final rs = await http.get(Uri.parse(
+        'https://raw.githubusercontent.com/flutter/flutter/master/packages/flutter/lib/analysis_options_user.yaml'));
```
jonasfj, line 50: When and why do we need this? It's only relevant when the flutter SDK isn't available, right?
Why do we need to support that use case?

-------

Also HTTP requests without retries is never a good idea.

lib/src/analysis_options.dart
```diff
@@ -3,79 +3,102 @@
...
+  } catch (_) {
+    // no-op
+  }
+
+  // fallback empty options
+  _logger.warning('Unable to load default Flutter analysis options.');
+  return '';
+}
+
+Future<String> _getPedanticAnalysisOptions() async {
```
jonasfj, line 64: When and why do we need this?
Why do we need to support that use case?

-------

Also HTTP requests without retries is never a good idea.

lib/src/analysis_options.dart
isoos: > When and why do we need this?

pub.dev doesn't need this, only standalone CI run without explicit Flutter SDK directory set. (We could detect where it is in the path and work out the subdirectory, but it may have issues on Windows).

lib/src/analysis_options.dart
isoos: Similarly: when `pana` is run as part of the CI workflow.

lib/src/analysis_options.dart
jonasfj: Okay, I'm down for that...

Let's add some retries through... or use `http_retry`.

jonasfj (approved):

lib/src/analysis_options.dart
isoos: Implemented with hand-rolled retry, because `http_retry` will conflict with `pub-dev` (at least for a while).

jonasfj (approved):

PR #867, opened 2021-04-27
Mask versions in end2end tests.

isoos:


jonasfj (commented):
We could also choose to only marks versions when we compare the two files... And not when stored on the file system...

Hmm, yeah, I can see the downsides with this too...

Maybe we should use a snapshot from pub.dev, like we have a fake server on pub-dev..
We could simply cache the version listing responses, maybe?

isoos:
> Maybe we should use a snapshot from pub.dev, like we have a fake server on pub-dev..
> We could simply cache the version listing responses, maybe?

I think it may have a marginal benefit, but needs a lot of effort and supporting infrastructure.

I've also contemplated masking the non-major part of the versions, e.g. `1.2.3-x+y` becomes `1.0.0` or `1.*.*`. That way the format remains reasonable. wdyt?

jonasfj:
@isoos, hmm, what if we proxy all requests to `pub.dev` over localhost, and then when proxying: `https://pub.dev/api/packages/<package>` we use the `published` (date-time) to remove everything newer than a hardcoded date?

I don't think we use API end-points other than `https://pub.dev/api/packages/<package>`, and filtering by `published` date, would effectively give us a fixed-in-time view of the repository.

It's not as robust a faking everything, but it allows us to test with real packages.

PR #869, opened 2021-05-04
WIP Add ability to Parse Gradle files

Sameerkash:
Related to issue  dart-lang/pub-dev#4615

cc @jonasfj

### DIFF
--- a/lib/src/gradle_parser.dart
+++ b/lib/src/gradle_parser.dart
@@ -1,0 +1,102 @@
+import 'package:petitparser/petitparser.dart';
+
+/// Grammar for Gradle
+class GradleGrammerDefinition extends GrammarDefinition {
+  @override
+  Parser start() => ref0(value).end();
+
+  /// A token, smallest unit to parse
+  Parser token(Object source, [String name]) {
+    if (source is String) {
+      return source.toParser(message: 'Expected ${name ?? source}').trim();
+    } else if (source is Parser) {
+      ArgumentError.checkNotNull(name, 'name');
+      return source.flatten('Expected $name').trim();
+    } else {
+      throw ArgumentError('Unknown token type: $source.');
+    }
+  }
+
+  /// Prodctions to parse a block of elements
+  Parser block() => ref1(token, '{') & ref0(elements) & ref1(token, '}');
+
+  /// A list of key-value pairs seperated by a new line
+  Parser elements() =>
+      ref0(pair).separatedBy(ref1(token, '\n'), includeSeparators: false);
+
+  /// Proudctions to parse each block
+  Parser members() =>
+      ref0(pair).separatedBy(ref1(token, '\n'), includeSeparators: false);
+
+  /// a key-value pair seperated by a whitespace
+  Parser pair() => ref0(stringToken) & ref1(token, ' ') & ref0(value);
+
+  /// Primitive value tokens
+  Parser trueToken() => ref1(token, 'true');
+  Parser falseToken() => ref1(token, 'false');
+  Parser stringToken() => ref2(token, ref0(stringPrimitive), 'string');
+  Parser numberToken() => ref2(token, ref0(numberPrimitive), 'number');
+
+  /// Possible values for parsing
+  Parser value() => [
+        ref0(members),
+        ref0(block),
+        ref0(stringToken),
+        ref0(numberToken),
+        ref0(trueToken),
+        ref0(falseToken),
+      ].toChoiceParser(failureJoiner: selectFarthestJoined);
+
+  Parser characterPrimitive() => pattern('^"\\');
+  Parser stringPrimitive() =>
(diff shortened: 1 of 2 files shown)


Sameerkash:
I'm not able to successfully parse a file, might be some mistake in my grammar as It involved some guesswork.

lib/src/gradle_parser.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  /// Prodctions to parse a block of elements
+  Parser block() => ref1(token, '{') & ref0(elements) & ref1(token, '}');
+
+  /// A list of key-value pairs seperated by a new line
+  Parser elements() =>
+      ref0(pair).separatedBy(ref1(token, '\n'), includeSeparators: false);
+
+  /// Proudctions to parse each block
+  Parser members() =>
```
Sameerkash, line 28: This doesn't seem to be working as expected.

lib/src/gradle_parser.dart
```diff
@@ -0,0 +1,102 @@
+import 'package:petitparser/petitparser.dart';
+
+/// Grammar for Gradle
+class GradleGrammerDefinition extends GrammarDefinition {
```
jonasfj, line 4: private classes please... Personally I haven't tried to use the `GrammarDefinition` stuff, but I suppose can might make sense.

lib/src/gradle_parser.dart
```diff
@@ -0,0 +1,102 @@
...
+      ].toChoiceParser(failureJoiner: selectFarthestJoined);
+
+  Parser characterPrimitive() => pattern('^"\\');
+  Parser stringPrimitive() =>
+      char('"') & ref0(characterPrimitive).star() & char('"');
+  Parser numberPrimitive() => (digit().plus());
+}
+
+/// Parser for Gradle
+class GradleParserDefinition extends GradleGrammerDefinition {
```
jonasfj, line 57: It might be overkill to not just add all the production actions in `GradleGrammerDefinition`...

Actually, it might even be simpler to just do simple grammar without a custom subclass of `GrammarDefinition`.. On the other hand I can see it as an elegant way to encapsulate things.. So long as the monstrosity is private -- as a type `GrammarDefinition` has a lot of members that have no business being publicly re-exported from a library -- IMO, it only make sense as a private type.

https://pub.dev/packages/petitparser#writing-a-simple-grammar

lib/src/gradle_parser.dart
```diff
@@ -0,0 +1,102 @@
...
+  Parser pair() => ref0(stringToken) & ref1(token, ' ') & ref0(value);
+
+  /// Primitive value tokens
+  Parser trueToken() => ref1(token, 'true');
+  Parser falseToken() => ref1(token, 'false');
+  Parser stringToken() => ref2(token, ref0(stringPrimitive), 'string');
+  Parser numberToken() => ref2(token, ref0(numberPrimitive), 'number');
+
+  /// Possible values for parsing
+  Parser value() => [
```
jonasfj, line 41: You might need to reference gradle documentation here... find the grammar and reference the rules...

Or we need to find a way to recover from errors...

Sameerkash: Okay, I could not find the perfect docs for this, just referred things here and there to find the structure specific to android, maybe I'll have to dig in a bit deeper.

pubspec.yaml
```diff
@@ -4,43 +4,44 @@ version: 0.16.0
...
-  args: '>=1.6.0 <3.0.0'
-  async: '>=1.13.3 <3.0.0'
-  cli_util: '>=0.2.0 <0.4.0'
-  html: '>=0.13.3 <0.16.0'
-  http: '>=0.11.3 <0.14.0'
-  io: '>=0.3.4 <2.0.0'
-  json_annotation: '>=2.0.0 <5.0.0'
-  logging: '>=0.11.3+1 <2.0.0'
-  markdown: '>=3.0.0 <5.0.0'
+  analyzer: ">=0.41.0 <2.0.0"
```
jonasfj, line 10: let's avoid formatting changes here

Sameerkash: oops yeah, sorry that automatically happens and goes without my knowledge, I'll revert it back.

Sameerkash:
Closing this for now since we don't have a proper implementation.

lib/src/gradle_parser.dart
jonasfj: Yeah, the stuff we talked about in [hackers-pub-dev-](https://github.com/flutter/flutter/wiki/Chat) is similar in many ways... Tokenize the input, that'll strip comments and ideally also handle strings (so we ignore brackets inside strings), then use some sort of heuristic / parser to find the relevant bits using the brackets as guides for how to find it :D

Maybe it's just build a tree where everything inside a backet of any sorts becomes a sub-tree... then look for a node with the label `defaultConfig`, find children named `minSdkVersion` look for the next integer.. or something like that.

Issue #879, opened 2021-05-18
Redirected Homepage URL is showing as broken


timsneath:
https://pub.dev/packages/win32/score complains that win32.pub is unreachable:
![image](https://user-images.githubusercontent.com/2319867/118702390-b650f300-b7c9-11eb-93b5-ebf3403252ca.png)

This is hosted on Google Domains and is just a redirect to https://pub.dev/documentation/win32/latest/

Surprised that this gave an error, since the backend infra is pretty durable. Is there a log to diagnose this? Or is it that pana gives an error for a redirected URL?

isoos:
We are using package:safe_url_check with the defaults, which is following 8 redirects:
https://github.com/google/dart-neats/blob/master/safe_url_check/lib/safe_url_check.dart#L107

I think the issue here may be ipv4/ipv6 or other network setting: from the browser, I can reach win32.dev, and it redirects properly, however curl gets me:
```
curl https://win32.dev/
curl: (6) Could not resolve
```

jonasfj:
Maybe we should try with `win32.pub` :see_no_evil: (credits athomas for noticing)

jonasfj:
```
[safe_url_check]$dart test -r expanded
00:00 +0: test/unique_local_ip_test.dart: isUniqueLocalIpV6
00:00 +1: test/private_ip_test.dart: isPrivateIpV4
00:00 +2: loading test/safe_url_check_test.dart
00:00 +2: test/safe_url_check_test.dart: safeUrlCheck("https://win32.pub")
00:00 +3: test/safe_url_check_test.dart: safeUrlCheck("https://google.com")
00:00 +4: test/safe_url_check_test.dart: safeUrlCheck("https://github.com")
00:01 +5: test/safe_url_check_test.dart: safeUrlCheck("https://github.com/google/dart-neats.git")
00:01 +6: test/safe_url_check_test.dart: !safeUrlCheck("https://github.com/google/dart-neats.git/bad-url")
00:02 +7: All tests passed!
```

Maybe we just need to analyze it again.

timsneath:
Looks like this URL resolved upon a new analysis run: 
https://pub.dev/packages/win32/score

sigurdm:
Looks like it is working.

Closing

PR #885, opened 2021-05-21
UrlChecker.exists to enable a caching hook for pub.dev

isoos:
- #879
- pub.dev should have a cache that stores the URLs that can be accessed successfully.
- The cache could have a longer timeout (e.g. 7 days).
- If a URL is not accessible, it doesn't become a negative cached entry, it would be tried via `super.exists()` and on success the status could be stored in cache.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -132,7 +132,7 @@ class UrlChecker {
 
   UrlChecker({
     int maxCacheSize,
-  }) : _maxCacheSize = maxCacheSize ?? 100 {
+  }) : _maxCacheSize = maxCacheSize ?? 10000 {
     addInternalHosts([
       'dart.dev',
       RegExp(r'.*\.dart\.dev'),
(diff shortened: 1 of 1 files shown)


lib/src/download_utils.dart
```diff
@@ -153,6 +153,10 @@ class UrlChecker {
     return _internalHosts.every((p) => p.allMatches(uri.host).isEmpty);
   }
 
+  Future<bool> exists(Uri uri) async {
```
jonasfj, line 156: Do you want to subclass and inject a custom `UrlChecker`?

If so, I think documenting all members would be a good idea.

But if you want some caching, why not just cache the last 1000 URLs we looked using an in-memory dictionary? Using a shared cache for this feels like overkill..

jonasfj (approved):

lib/src/download_utils.dart
isoos: Good point, updated it, PTAL.

However, I'd like to keep the option open to implement a redis-based cache too, so updated the API accordingly...

PR #890, opened 2021-05-26
Migrate to latest dependencies, selective migration to null safety

kevmoo:
- Migrate to latest dependencies
- selective migration to null safety
- changelog


kevmoo:
CC @natebosch

pubspec.yaml
```diff
@@ -1,46 +1,46 @@
...
-  args: '>=1.6.0 <3.0.0'
-  async: '>=1.13.3 <3.0.0'
-  cli_util: '>=0.2.0 <0.4.0'
-  html: '>=0.13.3 <0.16.0'
-  http: '>=0.11.3 <0.14.0'
-  io: '>=0.3.4 <2.0.0'
-  json_annotation: '>=2.0.0 <5.0.0'
-  logging: '>=0.11.3+1 <2.0.0'
-  markdown: '>=3.0.0 <5.0.0'
+  analyzer: ^1.0.0
```
isoos, line 10: Please don't force-update these dependencies yet, as it will be incompatible with `pub-dev`, and we'll be force-upgrade with too many breaking changes.

test/goldens/end2end/__timestamp.txt
```diff
@@ -1 +1 @@
-2021-05-21T07:12:03.731197Z
+2021-05-26T01:34:20.962183Z
```
isoos, line 1: Note: end2end tests should be able to handle new versions now, if you don't change the output, PRs should not affect these goldens. We are using this timestamp to cut out newer versions.

pubspec.yaml
jonasfj: Do we have a choice?

isoos:
Note: with the recent updates in pub-dev, I've tried to use this branch of pana, and with a moderate amount of additional dependency override we can make it work... So if @kevmoo doesn't have the time to fix the conflicting files and CI, I will do it tomorrow.

kevmoo:
Run with it, Istvan! Nate and I worked around what we needed for
repo_manager.

Issue #895, opened 2021-06-03
Pana 0.17.1 creates issues with a set of packages that are published as a group


shyndman:
Hello there,

I manage a set of interrelated packages (https://github.com/nhost/nhost-dart) which are stored in a mono-repo, and published as a group (using https://github.com/invertase/melos). We frequently commit changes that impact more than one package, in order to keep the packages all working together as their public interfaces change.

`pana` is used by our CI (GitHub-Actions-based) in order to ensure that the scores of the packages remain above some threshold. The sources we're scoring are entirely contained in the repo — their hosted versions are not appropriate for use during scoring (because like I said, cross-cutting changes, before publishing takes place)

The way we had this working was by having CI modify the packages' pubspecs in order to use path dependencies, just prior to running pana.

pana 0.17.1 throws a wrench into this approach, as the package being scored is now copied into a temp dir before performing analysis. Which means, a `path: ../my_local_dependency` fails, as the parent directory is elsewhere.

Do you have any suggestions as to how I might structure things as to score a single package in a multi-package repo?

Thank you.

Related to https://github.com/nhost/nhost-dart/issues/18

isoos:
> Do you have any suggestions as to how I might structure things as to score a single package in a multi-package repo?

I can see multiple ways to solve this:
- we could start the copy with the repo-root directory
- we could keep an option around to override the package directory
- we could rewrite e.g. `path` references to point to the original location

Is it only the `path` dependencies, or do you reference assets outside of the current package directory too?

shyndman:
Thanks for the quick reply.

I am only referencing `path` deps.

isoos:
What I'm interested in is the following: do you reference across your package in the monorepo with `path: ../my_local_dependency` dependencies in `pubspec.yaml`, or do you load files/configs/assets too? If the earlier, we could just do the path rewrite, if the later, the repo-root directory copy is the likely way (when copying and not in-place).

shyndman:
I only reference packages using `path: ../my_local_dependency_`. I do not reference files/configs/assets in other packages.

shyndman:
Does that answer your question? Path rewrite seems like the way to go.

shyndman:
Ugh, just realized that _I_ can rewrite to an absolute path when modifying the pubspecs. 🤦 

You're free to keep this open if you think it's valuable, but my problems are sorted out. Thank you.

shyndman:
[...]

PR #906, opened 2021-06-14
Detect git monorepo root and copy everything from that root for local analysis

isoos:
- #895
- I think this is a better solution than `path` rewrites, as it doesn't need to change the `pubspec.yaml` at all.


jonasfj (approved):
Hmm, wouldn't it be preferable to use relative paths instead?

On the other hand, this solves the problem, I'm okay with it :D

PR #925, opened 2021-07-17
Only require flutter if package uses flutter

kentcb:
If a package is Dart-only, it should not receive a 10 point deduction when Flutter isn't installed, but that is currently what occurs. This PR fixes the logic in question.


isoos (approved):

isoos:
Please fix formatting with dartfmt.

kentcb:
@isoos Formatting is fixed, but CI seems to be failing e2e test due to difference in line ending even though my change shouldn't impact that...any thoughts?

kentcb:
Ah, actually, it would make sense if the golden files include messages/output related to Flutter for non-Flutter packages because they would no longer be there. I'll take a look after work.

isoos:
> Ah, actually, it would make sense if the golden files include messages/output related to Flutter for non-Flutter packages because they would no longer be there. I'll take a look after work.

Delete the files in the `test/goldens/end2end` directory and run test again. They will be failing, but regenerate the files with the current output, and it will be easy to compare the impact.

kentcb:
Yeah, I tried that but it's not exactly Windows-friendly. i.e. generating a bunch of platform-specific junk. Am trying to fix somewhat manually...

kentcb:
@isoos OK, good to go

isoos (approved):
Thanks, this looks great!

PR #930, opened 2021-08-05
Terminate the timeouted processes with sigkill.

isoos:
- I think one of the core reasons we have incomplete dartdoc reports may be the lack of kill signal.
- https://github.com/dart-lang/pub-dev/issues/4750


sigurdm (approved):
LGTM

PR #931, opened 2021-08-05
runProc's deduplicate parameter is no longer used and will be ignored.

isoos:
The SDK issue has been fixed a long time ago.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -26,8 +26,13 @@ Future<ProcessResult> runProc(
   String? workingDirectory,
   Map<String, String>? environment,
   Duration? timeout,
+  @Deprecated('The parameter has no effect and will be removed.')
   bool deduplicate = false,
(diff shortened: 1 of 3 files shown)


lib/src/utils.dart
```diff
@@ -26,8 +26,13 @@ Future<ProcessResult> runProc(
   String? workingDirectory,
   Map<String, String>? environment,
   Duration? timeout,
-  bool deduplicate = false,
```
sigurdm, line 29: Is this considered publicly exposed? Otherwise I would just remove it straight away.

isoos: With prior breaking changes we were honoring the semantic versioning, and if we want to remove this, I'd rather bump the version. I'm fine either way.

sigurdm (approved):
Ah - the runProc function is indeed exposed.
Then let us deprecate.

Issue #824, opened 2021-01-15
dart run pana --hosted googleapis_auth fails to find any platforms


jonasfj:
> ## ✗ Support multiple platforms (0 / 20)
> ### [x] 0/20 points: Supports 0 of 2 possible platforms (native, js)

I also checked the JSON output, there is no tags.

jonasfj:
Probably same issue with https://pub.dev/packages/webdriver

sigurdm:
Adding some debug prints reveals the issue:
```
Package not compatible with sdk dart using runtime native-aot, Because:
* `package:googleapis_auth/auth_browser.dart` that imports:
* `package:googleapis_auth/src/oauth2_flows/implicit.dart` that imports:
* `dart:js`)
Package not compatible with sdk dart using runtime native, Because:
* `package:googleapis_auth/auth_browser.dart` that imports:
* `package:googleapis_auth/src/oauth2_flows/implicit.dart` that imports:
* `dart:js`)
Package not compatible with sdk dart using runtime js, Because:
* `package:googleapis_auth/auth_io.dart` that imports:
* `package:googleapis_auth/src/oauth2_flows/metadata_server.dart` that imports:
* `dart:io`)
```

We should find a way to have that explanation bubble up to the output. There is a comment
```
        if (violation != null) {
          // This is reported elsewhere
        } else {
```

It seems that is lying.

isoos:
https://github.com/dart-lang/pana/pull/929 adds a bit more information to the report. Notes from the code review to be addressed later:
- > `Consider supporting these prerelease platforms:`
  > This line reads weirdly...
- > `Package not compatible with SDK dart`
  > We should end with a full stop `.`.
- > `Because it is not compatible with any of the supported runtimes: native-aot, native, js`
  > We should end with a full stop `.`.
  > We really should document somewhere what this means, and link. As the package author there is not much to go on.

isoos:
In the case of `webdriver 3.0.0` we don't report any platform compatibility because:
- it doesn't have a single top-level library (`webdriver.dart`), but multiple ones
- `*_io.dart` imports `dart:io`, `*_html` imports `dart:html`, and as we iterate through the platforms, each of them is blocked by some of the libraries

sigurdm:
> it doesn't have a single top-level library (webdriver.dart), but multiple ones
*_io.dart imports dart:io, *_html imports dart:html, and as we iterate through the platforms, each of them is blocked by some of the libraries

[...]

PR #932, opened 2021-08-05
Expose explanations when top-level libraries are not compatible with any runtime

isoos:
- #824 
- also updates the message formats


lib/src/tag/_specs.dart
```diff
@@ -122,16 +122,17 @@ class FlutterPlatform {
 
 class Sdk {
   final String name;
+  final String formattedName;
   final List<String> allowedSdks;
   final List<Runtime> allowedRuntimes;
-  Sdk(this.name, this.allowedSdks, this.allowedRuntimes);
+  Sdk(this.name, this.formattedName, this.allowedSdks, this.allowedRuntimes);
```
sigurdm, line 128: Consider using named params to make the use-site easier to read.

lib/src/tag/_violations.dart
```diff
@@ -118,45 +118,47 @@ class SdkViolationFinder {
             return nonAllowedSdks.isEmpty
                 ? null
                 : (path) => Explanation(
-                      'Package not compatible with SDK ${sdk.name}',
+                      'Package is not compatible with ${sdk.formattedName} SDK.',
```
sigurdm, line 121: Maybe?
```suggestion
                      'Package is not compatible with the ${sdk.formattedName} SDK.',
```

lib/src/tag/_violations.dart
```diff
@@ -118,45 +118,47 @@ class SdkViolationFinder {
...
+          explanations.add(runtimeResult);
           supports = false;
           break;
         }
       }
       if (supports) return null;
     }
     return Explanation(
-      'Package not compatible with SDK ${sdk.name}',
+      'Package not compatible with ${sdk.formattedName} SDK.',
```
sigurdm, line 158: Same here

sigurdm (approved):
Nice!

PR #934, opened 2021-08-09
Rewrite of UrlChecker and UrlStatus.

isoos:
I'd like to have redis-based URL status caching on pub.dev, and the current methods do not provide a way to cache the missing status, instead it will be retried every time. 

Also: added `flutter.dev` in the default list of internal urls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
+## 0.20.0
+
+* **BREAKING CHANGES**
+  * `UrlStatus` is converted to a class with fields.
+  * `UrlChecker` internal cache is removed (incl. `maxCacheSize`, `existsInCache`, `markExistsInCache`).
+    Caching implementation should wrap the `UrlChecker.checkUrlExists` method.
+
 ## 0.19.1
 
 * Upgraded dependencies: `analyzer` and `json_serializable`.
 * Report a reason when no platform is detected - still lacking more details.
 * Terminate the timeouted processes with `sigkill`.
 * The `deduplicate` parameter in `runProc` is no longer used and will be ignored.
+* Updated cache handlers for `UrlChecker`.
 
 ## 0.19.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * **BREAKING CHANGES**
   * `UrlStatus` is converted to a class with fields.
   * `UrlChecker` internal cache is removed (incl. `maxCacheSize`, `existsInCache`, `markExistsInCache`).
-    Caching implementation should wrap the `UrlChecker.checkUrlExists` method.
+    Caching implementations should wrap the `UrlChecker.checkUrlExists` method.
 
 ## 0.19.1
 
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -107,25 +107,44 @@ const _repoReplacePrefixes = {
   'https://www.gitlab.com': 'https://gitlab.com',
 };
 
-enum UrlStatus {
-  invalid,
-  internal,
-  missing,
-  exists,
+/// The URL's parsed and queried status.
+class UrlStatus {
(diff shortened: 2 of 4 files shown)


isoos:
Hm, actually, it may be worth to split the two function:
- we want to have URL validity check and internal URL detection in one part,
- safe url check + status caching in the other.

The split would allow us to (a) allow internal URLs for internal packages, while also (b) check their existence and report it if they are broken.

lib/src/download_utils.dart
```diff
@@ -164,29 +188,39 @@ class UrlChecker {
...
     }
+    return exists ? UrlStatus.exists : UrlStatus.missing;
   }
 
   /// Checks if the [uri] exists in cache. The cache contains only
   /// the valid URLs, failures (which may be short-lived, transient)
   /// are not stored, rather retried.
   ///
   /// Returns true if [uri] has been valid in the cache, false otherwise.
+  @Deprecated('Do not use this method, it will be removed in a future release.')
```
sigurdm, line 217: Please point to the alternative (if there is one)

lib/src/download_utils.dart
```diff
@@ -114,18 +114,42 @@ enum UrlStatus {
...
+    @Deprecated('Do not use built-in cache, it will be removed in a future release.')
+        int? maxCacheSize,
+    GetCachedUrlStatus? getCachedUrlStatus,
+    SetCachedUrlStatus? setCachedUrlStatus,
+  })  : _maxCacheSize = maxCacheSize ?? 10000,
+        _getCachedUrlStatus = getCachedUrlStatus,
+        _setCachedUrlStatus = setCachedUrlStatus {
+    // Making sure both are (un)set at the same time.
+    if (_getCachedUrlStatus != null) {
+      assert(_setCachedUrlStatus != null);
```
sigurdm, line 144: I would prefer using  `ArgumentError` over asserts here.

lib/src/download_utils.dart
```diff
@@ -114,18 +114,42 @@ enum UrlStatus {
   exists,
 }
 
+/// The method returns the cached existence status of the [url]
+/// - returns `true` if the [url] exists,
+/// - returns `false` if the [url] is missing or has issues serving,
+/// - returns `null` if the [url] status is not in the cache.
+typedef GetCachedUrlStatus = Future<bool?> Function(String url);
```
sigurdm, line 121: If these always come as a pair, perhaps it would be better to have an interface with two methods?

isoos:
PTAL, it is a bit different design, leaving the cache implementation to the class that overrides this.

lib/src/report/template.dart
```diff
@@ -48,26 +47,30 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+    } else if (!status.exists) {
       issues.add(
         Issue(
           "$name doesn't exist.",
           span: tryGetSpanFromYamlMap(pubspec.originalYaml, key),
           suggestion: 'At the time of the analysis `$url` was unreachable.',
         ),
       );
-    } else if (status == UrlStatus.exists && !url.startsWith('https://')) {
+    } else if (status.exists && !status.isSecure) {
```
sigurdm, line 73: I think in this case you can assume status.exists
```suggestion
    } else if (!status.isSecure) {
```

CHANGELOG.md
```diff
@@ -1,9 +1,17 @@
+## 0.20.0
+
+* **BREAKING CHANGES**
+  * `UrlStatus` is converted to a class with fields.
+  * `UrlChecker` internal cache is removed (incl. `maxCacheSize`, `existsInCache`, `markExistsInCache`).
+    Caching implementation should wrap the `UrlChecker.checkUrlExists` method.
```
sigurdm, line 6:
```suggestion
    Caching implementations should wrap the `UrlChecker.checkUrlExists` method.
```

sigurdm (approved):
LGTM

PR #938, opened 2021-08-11
Track and store URL problems in the Summary.

isoos:
- https://github.com/dart-lang/pub-dev/issues/4902
- Only URLs with problems are stored (assuming it will be the rare case).
- URLs in `README.md` and `CHANGELOG.md` could also go into this list (in a future change).
- Tested with `random_names`, I wouldn't put it to the end2end list for now, I'd rather look for a package that has a problem in their readme.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -133,6 +133,18 @@ class UrlStatus {
         isInternal = false,
         isSecure = false,
         exists = false;
+
+  /// Returns a brief problem code that can be displayed when linking to it.
+  /// Returns `null` when URL has no problem.
+  String? getProblemCode({
+    required bool packageIsKnownInternal,
+  }) {
+    if (isInvalid) return 'invalid';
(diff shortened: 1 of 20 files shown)


sigurdm (commented):
Could we make a non-end2end test for this?

lib/src/download_utils.dart
```diff
@@ -133,6 +133,18 @@ class UrlStatus {
         isInternal = false,
         isSecure = false,
         exists = false;
+
+  /// Returns a brief problem code that can be displayed when linking to it.
+  /// Returns `null` when URL has no problem.
+  String? getProblemCode({
+    required bool packageIsKnownInternal,
+  }) {
+    if (isInvalid) return 'invalid';
```
sigurdm, line 142: Do these codes belong in an enum in the model?

isoos:
> Could we make a non-end2end test for this?

added them, PTAL.

sigurdm (approved):
LGTM

PR #939, opened 2021-08-13
Fix retry of package download.

### LINKED ISSUE #940: Improve retry-logic for downloading
https://github.com/dart-lang/pana/blob/master/lib/src/download_utils.dart#L40-L55 throws an Error if we don't get 200. This won't be retried.

We should improve this logic to retry at least 500's.

isoos:
Fixes #940.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -49,7 +49,8 @@ Future<void> downloadPackage(
         }
         if (res.statusCode != 200) {
           // throwing error stops retrying (e.g. on 404)
-          throw AssertionError('Unable to access URL: "$packageUri".');
+          throw AssertionError(
+              'Unable to access URL: "$packageUri" (status code: ${res.statusCode}).');
         }
         await _extractTarGz(res, destination);
       } finally {
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -43,7 +43,12 @@ Future<void> downloadPackage(
       try {
         final req = await c.getUrl(packageUri);
         final res = await req.close();
+        if (res.statusCode >= 500) {
+          // may be a temporary failure, let's retry...
+          throw Exception('Unable to access URL: "$packageUri".');
+        }
         if (res.statusCode != 200) {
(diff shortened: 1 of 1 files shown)


lib/src/download_utils.dart
```diff
@@ -43,7 +43,12 @@ Future<void> downloadPackage(
       try {
         final req = await c.getUrl(packageUri);
         final res = await req.close();
+        if (res.statusCode >= 500) {
+          // may be a temporary failure, let's retry...
+          throw Exception('Unable to access URL: "$packageUri".');
+        }
         if (res.statusCode != 200) {
+          // throwing error stops retrying (e.g. on 404)
           throw AssertionError('Unable to access URL: "$packageUri".');
```
sigurdm, line 52: I think we should still find something else to throw here. AssertionError is not really meant for this purpose. This might require making a custom exception class and a passing a `retryIf` parameter.

lib/src/download_utils.dart
```diff
@@ -43,7 +43,12 @@ Future<void> downloadPackage(
       try {
         final req = await c.getUrl(packageUri);
         final res = await req.close();
+        if (res.statusCode >= 500) {
+          // may be a temporary failure, let's retry...
+          throw Exception('Unable to access URL: "$packageUri".');
```
sigurdm, line 48: We might as well add more information here, such as the status-code.

lib/src/download_utils.dart
isoos: As long as it is an `Error`, we don't need to pass it to `retryIf`.

sigurdm: My point is: it is not right to throw an error here. Errors are for unrecoverable situations, and failing to download a file should be recoverable.

AssertionError has a special meaning, it is thrown when an Assertion fails - I would never expect that in production (that is running without asserts).

isoos: Maybe we should just depend on `package:http_retry` (which we already do in pub-dev), for such logic?

isoos:
Closing in favor of #943.

Issue #941, opened 2021-08-13
unable to run `pana` locally (failing to parse Dart SDK info)


pq:
```
☁  pana [master] pub global run pana
INFO       Running `/Applications/Tools/dart-sdk/bin/dart --version`...
Unhandled exception:
Null check operator used on a null value
#0      new DartSdkInfo.parse (package:pana/src/sdk_env.dart:505:38)
#1      ToolEnvironment._init (package:pana/src/sdk_env.dart:87:37)
<asynchronous suspension>
#2      ToolEnvironment.create (package:pana/src/sdk_env.dart:160:5)
<asynchronous suspension>
#3      PackageAnalyzer.create (package:pana/src/package_analyzer.dart:60:28)
<asynchronous suspension>
#4      main (file:///Users/pquitslund/.pub-cache/hosted/pub.dartlang.org/pana-0.19.1/bin/pana.dart:134:22)
<asynchronous suspension>
```

Here's my SDK info:

```
☁  linter [master] dart --version
Dart SDK version: 2.15.0-edge.e8ddc0219f1e8f1ad784143fec693890e2b81954 (be) (Fri Aug 13 13:27:41 2021 +0000) on "macos_x64"
```

See also: https://github.com/dart-lang/linter/issues/2504

isoos:
@pq Can you also include the output of `/Applications/Tools/dart-sdk/bin/dart --version`? I've added the version you've provided in the test in #942, but it was parsed correctly already. I suspect it is a very different string that the code gets.

pq:
Sure thing.

```
☁  sdk [master] ⚡  /Applications/Tools/dart-sdk/bin/dart --version
Dart SDK version: 2.15.0-edge.e8ddc0219f1e8f1ad784143fec693890e2b81954 (be) (Fri Aug 13 13:27:41 2021 +0000) on "macos_x64"
```

Also note this is failing on CI builds too:

https://github.com/dart-lang/linter/pull/2871/checks?check_run_id=3326046782

isoos:
@pq: could you do a quick check for me: is this `stdout` or `stderr`? Current stable is `stderr`, and `pana` is checking for that, but maybe newer SDKs switch to `stdout`?

pq:
It looks like `stdout` to me.

/fyi @bkonyi

pq:
Thanks for taking a look @isoos!

isoos:
@pq: 0.20.0 has been published with the fix that should unblock you

[...]

PR #942, opened 2021-08-16
Fix SDK version parsing issue.

isoos:
- #941
- Throw `FormatException` when SDK version parsing fails.
- Read both `stdout` and `stderr`, as newer SDKs switched to `stdout`.


sigurdm (approved):
LGTM

sigurdm:
That's a good one :)

PR #943, opened 2021-08-17
Use package:http's retry in download_utils.dart

### LINKED ISSUE #940: Improve retry-logic for downloading

isoos:
Fixes #940.


sigurdm (approved):
LGTM

PR #950, opened 2021-08-23
Add license detection

bharat-biradar:
Replaces the current license detection mechanism of PANA with a license detector based on google's v2 classifier.


pubspec.yaml
```diff
@@ -8,6 +8,7 @@ environment:
 
 dependencies:
   analyzer: ^1.0.0
+  archive: ^3.1.2
```
isoos, line 11: I think this should be a dev_dependency, as it is only used outside of lib/.

lib/src/third_party/diff_match_patch/README.md
```diff
@@ -0,0 +1,12 @@
+# Diff-Match-Patch
```
isoos, line 1: Any reason not to use https://pub.dev/packages/diff_match_patch ?

lib/src/license_detection/crc32.dart
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'license_detector.dart';
```
isoos, line 5: I think this could be an independent library, doesn't need to be part of the license_detector.dart

lib/src/license_detection/license_detector.dart
```diff
@@ -0,0 +1,244 @@
...
+  for (var i = 0; i < matches.length; i++) {
+    if (retain[i]) {
+      retainedmatches.add(matches[i]);
+    }
+  }
+
+  return retainedmatches;
+}
+
+double unclaimedTokenPercentage(
```
isoos, line 213: This method is a bit confusing and without documentation: it takes matches with ranges, but it divides their sum with a separate count. Can it be more than 1.0? Please expand on it.
It also matches a field name, so maybe rename this to `calculateUnclaimedTokenPercentage`.

lib/src/third_party/diff_match_patch/README.md
bharat-biradar: We just needed the diff functionality and thought it was  better not to depend on third party library.

isoos: It is really strange to showcase its license and tests in the middle of the pana package. I'd rather have it as a dependency. wdyt @jonasfj ?

.gitignore
```diff
@@ -3,3 +3,6 @@
 .packages
 .pub/
 pubspec.lock
+test/license_test_assets/license_scenarios
+.vscode
+third_party/spdx/licenses
```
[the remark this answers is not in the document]
bharat-biradar: `third_party/spdx/licenses` and scenarios are excluded because they contain around 500 text files which will flood the whole PR and make it difficult to review I'll add them in a different PR.

lib/src/license_detection/crc32.dart
bharat-biradar: Yes but we did not want to expose the crc32 function and similar other functions outside of license_detection and hence made it part of the `license_detector`.

isoos: Making this a library doesn't expose it as public API. (that happens only with exports)

jonasfj (approved):
This is really awesome, thanks for all the hard work!

r+

I think we should land this and iterate on it further in follow-up PRs.

PR #952, opened 2021-09-08
Add spdx license assets files in lib/src/third_party/spdx/licenses

isoos:
- updated script that downloads the assets
- also updated initialization to load from new location
- added 400+ license files


jonasfj (approved):
Please update `lib/src/third_party/spdx/licenses` with a `README.md` that says where these licenses are from... and also a `LICENSE` file that at-least says something about licensing -- maybe just refer to the sources as is done in:
https://github.com/spdx/license-list-data#licensing-information

PR #953, opened 2021-09-08
Bump tested and used SDK to 2.14 – use Object.hashAll and drop pkg:quiver

kevmoo:


isoos (approved):

isoos:
Merging as-is, as tests are broken for independent reasons (the new license detection).

PR #954, opened 2021-09-09
Use lints/core.yaml as default options override.

isoos:


lib/src/analysis_options.dart
```diff
@@ -61,35 +61,24 @@ Future<String> _getFlutterAnalysisOptions(String? flutterSdkDir) async {
...
-          }
-        }
-      }
+    final rs = await _httpGetWithRetry(Uri.parse(
+        'https://raw.githubusercontent.com/dart-lang/lints/main/lib/core.yaml'));
+    if (rs.statusCode == 200 && rs.body.contains('rules:')) {
+      _cachedLintsCoreOptionsOnGithub = rs.body;
+      return _cachedLintsCoreOptionsOnGithub!;
     }
   } catch (_) {
```
sigurdm, line 76: Can we turn this into `} on Exception catch (_) {` now we're here?

sigurdm (approved):
LGTM

jonasfj (approved):

PR #957, opened 2021-09-10
Fix GitHub Action runs...

isoos:


isoos:
For some reason the setup script did not find the 2.5.0 release. Using a simple setup script from flutter plugins to setup Flutter.

jonasfj (approved):
but would it not be better to fix `subosito/flutter-action@v1.5.0` ?

Maybe it's just a temporary bug...

isoos:
Maybe it is temporary, we can revisit that later. Until then let's get our CI working :)

PR #963, opened 2021-09-13
Avoid `flutter pub get` failing on example directory.

isoos:


lib/src/utils.dart
```diff
@@ -244,4 +244,23 @@ extension ProcessResultExt on ProcessResult {
...
+  /// (both converted to [String]), with limits on individual line
+  /// lengths and total lines. Total length should not be more than 4KiB.
+  String get asTrimmedOutput {
+    Iterable<String> firstFewLines(String type, String output) sync* {
+      if (output.isEmpty) return;
+      yield '$type:';
+      yield* const LineSplitter()
+          .convert(output)
+          .take(10)
+          .map((e) => e.length > 200 ? e.substring(0, 200) : e);
```
sigurdm, line 258: Could we make the logic here, such that it is clear that the output was truncated (eg. [...200 more lines])

sigurdm (approved):
Nice
LGTM

sigurdm:
There are a few new test failures - we should probably look into them

jonasfj (approved):

isoos:
> There are a few new test failures - we should probably look into them

I'm on it. I'll leave the license test failures, but fix the end2end and version test.

isoos:
Not the only failing test is the known issue with the license detection. Merging...

PR #968, opened 2021-09-16
Better handling of JSON parsing issues in pub outdated.

### LINKED ISSUE #949: Handle failing "dart pub outdated" failing non-zero
When `dart pub outdated` we should not try to parse: `"It looks like pub.dartlang.org is having some trouble."`.
Instead when we get exit code non-zero we should probably not try to parse the output.
```
Error: FormatException: Unexpected character (at character 1) It looks like pub.dartlang.org is having some trouble. ^
at _ChunkedJsonParser.fail (native)
at _ChunkedJsonParser.parseNumber (native)
at _ChunkedJsonParser.parse (native)
at _parseJson (native)
at JsonDecoder.convert (native)
at JsonCodec.decode (native)
at ToolEnvironment.runPubOutdated (package:pana/src/sdk_env.dart:364)
at StackZoneSpecification._registerUnaryCallback.<fn> (unknown location)
```

isoos:
- Fixes #949.
- also updated flutter version processing to use the same method

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -297,15 +297,17 @@ class ToolEnvironment {
   Future<Map<String, dynamic>> getFlutterVersion() async {
     final result = _handleProcessErrors(
         await runProc([..._flutterCmd, '--version', '--machine']));
-    var content = result.asJoinedOutput;
     final waitingForString = 'Waiting for another flutter';
-    if (content.contains(waitingForString)) {
-      content = content
-          .split('\n')
-          .where((e) => !e.contains(waitingForString))
-          .join('\n');
-    }
-    return json.decode(content) as Map<String, dynamic>;
+    return result.parseJson(transform: (content) {
+      if (content.contains(waitingForString)) {
+        return content
+            .split('\n')
+            .where((e) => !e.contains(waitingForString))
+            .join('\n');
+      } else {
+        return content;
+      }
+    });
   }
 
   Future<bool> detectFlutterUse(String packageDir) async {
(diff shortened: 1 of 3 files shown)


jonasfj (approved):

lib/src/sdk_env.dart
```diff
@@ -297,15 +297,17 @@ class ToolEnvironment {
   Future<Map<String, dynamic>> getFlutterVersion() async {
     final result = _handleProcessErrors(
         await runProc([..._flutterCmd, '--version', '--machine']));
-    var content = result.asJoinedOutput;
     final waitingForString = 'Waiting for another flutter';
```
jonasfj, line 300: Can't we avoid this if we use `FLUTTER_ROOT=/path/to/flutter dart pub outdated` instead?

isoos: I think this happened sporadically in CI, and while we've changed the flutter setup since, we've always had `FLUTTER_ROOT` set.

jonasfj: My point is: I don't think `dart pub outdated` will ever print this line. But it's possible that `flutter pub outdated` will, so there is no need for this special case if we use `dart pub outdated` with `FLUTTER_ROOT` instead.

isoos: Yeah, we've always run `dart pub outdated` (there was a PR to change that, but got reverted), however, this is `flutter --version --machine`, which did write this on the output, on CI. Let's keep it for a bit, and maybe do a more systemic review of the code later: we should be able to identify similar patterns in multiple places, which may have become obsolete over time.

PR #969, opened 2021-09-17
Fail gracefully on source code parsing bugs.

isoos:
This helps to unblock analysis on bugs like #914: the URI of the source code is exposed in the report, the exception is logged with stacktrace and warning, and we will no longer display a forever-awaiting status on packages like that (e.g. `wialon 1.0.1+1`).


sigurdm (approved):
LGTM

PR #973, opened 2021-09-24
Tag Flutter plugins with is:plugin.

isoos:
https://github.com/dart-lang/pub-dev/issues/5133

### DIFF
--- a/lib/src/tag/tagger.dart
+++ b/lib/src/tag/tagger.dart
@@ -81,7 +81,6 @@ import 'package:path/path.dart' as path;
 import 'package:pub_semver/pub_semver.dart';
 
 import '../null_safety.dart';
-import '../pubspec.dart';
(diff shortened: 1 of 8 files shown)


lib/src/tag/tagger.dart
```diff
@@ -295,6 +299,14 @@ class Tagger {
     }
   }
 
+  /// Adds tags for Flutter plugins.
+  void flutterPluginTags(List<String> tags, List<Explanation> explanations) {
+    const _pluginTag = 'is:plugin';
+    if (_pubspec.hasFlutterPluginKey) {
```
sigurdm, line 305: We have access to the pubspec already:
```suggestion
    final pubspec = _pubspecCache.pubspecOfPackage(packageName);
    if (pubspec.hasFlutterPluginKey) {
```

sigurdm (approved):

PR #999, opened 2022-01-13
Upgrade package:analyzer dependency constraint.

isoos:
Also upgrading `json_annotation`, as the build message seemed to insist on a new range.


sigurdm (approved):
LGTM

PR #1000, opened 2022-01-13
screenshot generation

szakarias:


lib/src/model.dart
```diff
@@ -13,6 +13,22 @@ import 'pubspec.dart';
...
 part 'model.g.dart';
 
+@JsonSerializable(includeIfNull: false)
+
+/// A processed screenshot contains paths with processed images as stored in GCS.
+class ProcessedScreenshot {
+  final String originalImage; // full/<imagePath>
+  final String webpImage; // full/gen/<imagePath>.webp
+  final String webpThumbnail; // full/gen/100x100/<imagePath>.webp
+  final String pngThumbnail; // full/gen/100x100/<imagePath>.png
```
jonasfj, line 23: `full/` ?

lib/src/screenshots.dart
```diff
@@ -0,0 +1,170 @@
...
+Map<String, List<String>>? _screenshotValidationProblems;
+
+Map<String, List<String>> validateScreenshots(
+    List<p.Screenshot> screenshots, String pkgDir) {
+  if (_screenshotValidationProblems != null) {
+    return _screenshotValidationProblems!;
+  }
+
+  final problems = <String, List<String>>{};
+  for (final e in screenshots) {
```
jonasfj, line 21: Let's also limit the size and number of screenshots?

sigurdm:
I think we need to teach /usr/local/google/home/sigurdm/projects/pana/.github/workflows/test-package.yml how to find the webp tools: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners

test/goldens/end2end/url_launcher-6.0.3.json_report.md
```diff
@@ -13,7 +13,6 @@
 
 ### [*] 10/10 points: Package has an example
 
-* Found example at: `example/lib/main.dart`
```
sigurdm, line 16: Should this line have stayed?

test/screenshot_test.dart
```diff
@@ -0,0 +1,241 @@
...
+    final result = await processAllScreenshots(declared, pkgDir);
+
+    expect(result.first.processedScreenshot, isNull);
+    expect(result.first.problems, isNotEmpty);
+    expect(result.first.problems.first,
+        contains('No file found at specified screenshot `path`'));
+  });
+
+  test('success - process WebP, PNG and GIFs', () async {
+    if ((await runProc(['cwebp'])).exitCode != 0) {
```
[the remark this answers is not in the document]
szakarias: Done

test/goldens/end2end/url_launcher-6.0.3.json_report.md
szakarias: No, as discussed. Since we are adding more issue types it doesn't make sense to have a bullet point with the positive case in between the other issues.

lib/src/package_context.dart
```diff
@@ -132,6 +136,13 @@ class PackageContext {
     }
   }
 
+  Future<List<ScreenshotResult>> processScreenshots(
+      List<p.Screenshot> screenshots, String pkgDir) async {
+    if (_screenshotResults != null) return _screenshotResults!;
+    _screenshotResults = await processAllScreenshots(screenshots, pkgDir);
+    return _screenshotResults!;
```
jonasfj, line 143: nit: Can't we avoid two return statements with;
```suggestion
    if (_screenshotResults == null) {
       _screenshotResults = await processAllScreenshots(screenshots, pkgDir);
    }
    return _screenshotResults;
```

or maybe just:


```suggestion
    return _screenshotResults ??= await processAllScreenshots(screenshots, pkgDir);
```

lib/src/pubspec.dart
```diff
@@ -43,6 +43,8 @@ class Pubspec {
 
   Map get environment => (_content['environment'] as Map?) ?? {};
 
+  List<pubspek.Screenshot>? get screenshots => _inner.screenshots;
```
jonasfj, line 46: Then you can just do: `screenshots.isNotEmpty` going forward.
```suggestion
  List<pubspek.Screenshot> get screenshots => _inner.screenshots ?? [];
```

lib/src/package_context.dart
```diff
@@ -132,6 +136,13 @@ class PackageContext {
     }
   }
 
+  Future<List<ScreenshotResult>> processScreenshots(
```
jonasfj, line 139: This is a memorized method that takes two parameters... but don't return a new result of the parameters change.

That's probably a bad idea. We could document this behavior in a documentation string, so that people know that this method will forever return results for whatever it was called with first.

Or we could contemplate making it a getter for `screenshotResults` that lazily processes screenshots.
It might be nice to document the laziness in a documentation string as it affects performance characteristics, or maybe it's better to make it a method without parameters.

It's unclear to me how `pkgDir` is different from `PackageContext.packageDir`, similarly why can't we get `screenshots` from `PackageContext.pubspec`.

lib/src/model.dart
```diff
@@ -13,6 +13,22 @@ import 'pubspec.dart';
 
 part 'model.g.dart';
 
+@JsonSerializable(includeIfNull: false)
+
+/// A processed screenshot contains paths with processed images as stored in GCS.
+class ProcessedScreenshot {
+  final String originalImage; // <imagePath>
+  final String webpImage; // gen/<imagePath>.webp
```
jonasfj, line 21: So if the original file is `myfolder/myfile.png` this becomes `gen/myfolder/myfile.png.webp` is that correct?
Or are we removing the `.png` so that it becomes `gen/myfolder/myfile.webp`? :D

szakarias: We are not removing the orignal. We make a gen folder for all the images we need.

szakarias: So the answer to the first question is yes.

lib/src/package_context.dart
```diff
@@ -132,6 +135,15 @@ class PackageContext {
     }
   }
 
+  Future<List<ScreenshotResult>> processScreenshots() async {
+    if (_screenshotResults != null) return _screenshotResults!;
+    if (pubspec.screenshots != null) {
+      _screenshotResults =
+          await processAllScreenshots(pubspec.screenshots!, packageDir);
+    }
+    return _screenshotResults!;
```
jonasfj, line 144: So if `pubspec.screenshots` is not null, but an empty list, you'll get null assertion fail.
```suggestion
    if (_screenshotResults == null) {
      _screenshotResults =
          await processAllScreenshots(pubspec.screenshots ?? [], packageDir);
    }
    return _screenshotResults!;
```
or just:
```suggestion
    _screenshotResults ??=
          await processAllScreenshots(pubspec.screenshots ?? [], packageDir);
    return _screenshotResults!;
```

or just:
```suggestion
    return _screenshotResults ??=
          await processAllScreenshots(pubspec.screenshots ?? [], packageDir);
```

lib/src/screenshots.dart
```diff
@@ -0,0 +1,327 @@
...
+  /// The path to where the generated WebP screenshot is temporarily stored.
+  /// Should only be used if [processedScreenshot] is not null.
+  String get webpScreenshotPath =>
+      path.join(_outDir!, processedScreenshot!.webpImage);
+}
+
+Future<List<ScreenshotResult>> processAllScreenshots(
+    List<p.Screenshot> screenshots, String pkgDir) async {
+  final results = <ScreenshotResult>[];
+  final tempDir = Directory.systemTemp.createTempSync().path;
```
jonasfj, line 55: Are we going to delete this temporary folder? When?
Could we do this in a ` finally {...}` block?

lib/src/screenshots.dart
```diff
@@ -0,0 +1,327 @@
...
+/// Otherwise, a conversion is tried, using the `cwebp` tool which can handle
+/// PNG, JPEG, and TIFF as input. For GIF files the `gif2webp` conversion tool
+/// is used.
+///
+/// For any other input format the screenshot generation will fail. It will also
+/// fail if any of the mentioned tools are not available on the command line and
+/// an error message will be written to stderr.
+Future<List<String>> _generateWebpScreenshot(
+  String originalPath,
+  String webpPath,
```
jonasfj, line 170:
```suggestion
  String targetWebpPath,
```

lib/src/model.dart
```diff
@@ -13,6 +13,22 @@ import 'pubspec.dart';
 
 part 'model.g.dart';
 
+@JsonSerializable(includeIfNull: false)
+
+/// A processed screenshot contains paths with processed images as stored in GCS.
```
jonasfj, line 18:
```suggestion
/// A processed screenshot contains paths with processed images as stored using `storeResource` as passed to `PackageAnalyzer.inspectPackage`.
```
This is correct, right? This isn't necessarily the path in GCS, this is the path that was passed to `storeResource`.

What `storeResource` does with the path name is irrelevant to pana. When pub.dev calls pana it probably chooses to store the result in some sub-folder on GCS. But that detail does not concern pana.

lib/src/screenshots.dart
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  factory ScreenshotResult.failed(List<String> problems) {
+    return ScreenshotResult._(null, problems, null);
+  }
+
+  /// The path to where the generated WebP thumbnail is temporarily stored.
+  /// Should only be used if [processedScreenshot] is not null.
+  String get webpThumbnailPath =>
+      path.join(_outDir!, processedScreenshot!.webpThumbnail);
```
jonasfj, line 39: Why not make this a `Future<Uint8List> readWebpThumbnail() async` method?

The actual path within `_outDIr` is an implementation detail. The only thing the public pana interface is that `storeResource` is called with the path that appears in `processedScreenshot`.

jonasfj: Or we should make this a property `final Uint8List webpThumbnail;` that way we won't need to manage the life-cycle of the `_outDir` which is currently leaked?

lib/src/screenshots.dart
jonasfj: Maybe, we should just make `ScreenshotResult.webpThumbnail` a `Uint8List` as proposed above... That way we can delete this temporary directory at the end of `processAllScreenshots` in a finally-block.

This assumes we can reasonably keep all screenshots in memory? But don't we have sensible size limits that ensures this won't be a problem?

Note. I'm proposing that after generating the image, we read the image file and keep it memory as `Uint8List`.
That way we can delete the temporary folder when this function returns.

lib/src/screenshots.dart
```diff
@@ -0,0 +1,327 @@
...
+
+Future<List<String>> _convertWithCWebp(
+    String originalPath, String webpPath) async {
+  final cwebpResult =
+      await _checkedRunProc(['cwebp', originalPath, '-o', webpPath]);
+  if (cwebpResult.exitCode == 0) {
+    return [];
+  }
+  return [
+    '$originalPath: Converting screenshot with `cwebp` failed with exitcode ${cwebpResult.exitCode}.'
```
jonasfj, line 216:
```suggestion
    '`$originalPath`: Converting screenshot with `cwebp "$originalPath" -o result.webp` failed with _exit code_ `${cwebpResult.exitCode}`.'
```

If we use some backticks we get nice code formatting in markdown.

And if we provide the command, then advanced users can hopefully reproduce locally.

This is markdown, right?

Formatting advice regarding use of backticks applies to all these problems.

I think "exit code" is two words, or maybe "exit-code"? Alternatively we can write "... failed, exiting ${cwebpResult.exitCode}" -- but some people probably won't understand that.

lib/src/screenshots.dart
```diff
@@ -0,0 +1,327 @@
...
+  final infoResult = await runProc(['webpinfo', webpPath]);
+  if (infoResult.exitCode != 0) {
+    return [
+      "$originalPath: Running 'webpinfo' on $webpPath failed with exitcode ${infoResult.exitCode}."
+    ];
+  }
+
+  if ((infoResult.stdout as String).contains('Animation: 1')) {
+    staticWebpPath = path.join(Directory.systemTemp.path,
+        '${path.basenameWithoutExtension(webpPath)}_static.webp');
```
jonasfj, line 261: Another temporary folder that is leaked.

jonasfj (changes requested):

lib/src/package_context.dart
szakarias: done

lib/src/package_context.dart
szakarias: Updated.

lib/src/screenshots.dart
szakarias: done

lib/src/model.dart
szakarias: Done.

lib/src/package_analyzer.dart
```diff
@@ -92,7 +93,9 @@ class PackageAnalyzer {
     });
   }
 
-  Future<Summary> _inspect(String pkgDir, InspectOptions options) async {
+  Future<Summary> _inspect(String pkgDir, InspectOptions options,
+      {Future<void> Function(String filename, List<int> data)?
+          storeResource}) async {
```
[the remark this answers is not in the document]
szakarias: Done.

lib/src/package_context.dart
szakarias: Done,

lib/src/screenshots.dart
szakarias: Done.

lib/src/screenshots.dart
szakarias: Done.

lib/src/screenshots.dart
szakarias: Done.

lib/src/pubspec.dart
szakarias: done.

lib/src/report/documentation.dart
```diff
@@ -15,33 +16,46 @@ import '../pubspec.dart';
...
-    String packageDir, Pubspec pubspec) async {
+    PackageContext context, Pubspec pubspec) async {
+  final packageDir = context.packageDir;
   // TODO: run dartdoc for coverage
 
   final candidates = exampleFileCandidates(pubspec.name, caseSensitive: true);
   final examplePath = candidates
       .firstWhereOrNull((c) => File(p.join(packageDir, c)).existsSync());
-  final issues = <Issue>[
+  var issues = <Issue>[
```
sigurdm, line 26: nit: I think this can stay final

lib/src/screenshots.dart
```diff
@@ -0,0 +1,338 @@
...
+    problems.add(
+        '${screenshot.path}: Screenshot `description` is too long. Description length must be less '
+        'than 160 characters.');
+  }
+  if (screenshot.path.isEmpty) {
+    problems.add('${screenshot.path}: Screenshot `path` property is missing.');
+  }
+
+  final fullPath = path.join(pkgDir, screenshot.path);
+  if (!path.normalize(fullPath).startsWith(pkgDir)) {
```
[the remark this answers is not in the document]
szakarias: Done.

PR #1001, opened 2022-01-18
Formalized repository URL parsing and resolving.

isoos:
- reusing and repurposing the tests for the now-defunct `getRepositoryUrl`

### DIFF
--- a/lib/src/repository_url.dart
+++ b/lib/src/repository_url.dart
@@ -126,17 +126,16 @@ RepositoryUrl? _tryParseRepositoryUrl(String input) {
   if (repoSegmentIndex < 0) return null;
 
   String? separator;
-  final extraSegments = segments.skip(repoSegmentIndex + 1).toList();
-  if (extraSegments.isNotEmpty &&
-      _githubSegmentSeparators.contains(extraSegments.first)) {
-    separator = extraSegments.removeAt(0);
-  }
-
   String? branch;
-  if (RepositoryProvider.isGitHubCompatible(provider) &&
-      separator != null &&
-      extraSegments.isNotEmpty) {
-    branch = extraSegments.removeAt(0);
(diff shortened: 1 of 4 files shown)


lib/src/repository_url.dart
```diff
@@ -0,0 +1,180 @@
...
+  static const github = 'github';
+  static const gitlab = 'gitlab';
+  static const unknown = 'unknown';
+
+  static bool isGitHubCompatible(String provider) =>
+      provider == github || provider == gitlab;
+}
+
+const _repoReplacePrefixes = {
+  'http://github.com': 'https://github.com',
```
sigurdm, line 98: Perhaps this replacement is better done by changing the scheme of the parsed url.

```
url = url.replace(scheme:'https')
``

File: lib/src/repository_url.dart [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+      .pathSegments
+      .where((s) => s.isNotEmpty)
+      .toList();
+  final repoSegmentIndex = _repoSegmentIndex(provider, segments);
+  if (repoSegmentIndex < 0) return null;
+
+  String? separator;
+  final extraSegments = segments.skip(repoSegmentIndex + 1).toList();
+  if (extraSegments.isNotEmpty &&
+      _githubSegmentSeparators.contains(extraSegments.first)) {
```
sigurdm, line 131: Should this also be conditioned on the provider being github?

sigurdm (approved):
LGTM

lib/src/repository_url.dart
isoos: Done.

lib/src/repository_url.dart
isoos: Done.

sigurdm (approved):
LGTM

PR #1007, opened 2022-01-25
Fix retry on version listing HTTP requests during getVersionListing.

### LINKED ISSUE #1006: Retry 5xx errors when listing versions
```
 Error: Request to https://pub.dartlang.org/api/packages/http_parser failed with status 503: Service Unavailable.
    at BaseClient._checkResponseSuccess (package:http/src/base_client.dart:103:5)
    at BaseClient.read (package:http/src/base_client.dart:52:5)
    at StackZoneSpecification._registerUnaryCallback.<fn> (unknown location) 
```

Similar to https://github.com/dart-lang/pana/pull/1005 we need to retry if it throws a `ClientException` with 5xx, or not use `Client.read` and `Client.readBytes`, and then manually check the response code.

isoos:
Fixes #1006

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -242,7 +242,7 @@ Future<String> getVersionListing(String package, {Uri? pubHostedUrl}) async {
   log.fine('Downloading: $url');
 
   return await retry(() => http.read(url),
-      retryIf: (e) => e is SocketException || e is TimeoutException);
+      retryIf: (e) => e is IOException || e is http.ClientException);
 }
 
 class PanaProcessResult extends ProcessResult {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - `package:analyzer` dependency upgraded to `^3.1.0`.
 - Fix dart analysis. Code problems are now propagatet to the pana report.
+- Fix retry on version listing HTTP requests during `getVersionListing`.
 
 ## 0.21.5
 


lib/src/utils.dart
```diff
@@ -242,7 +242,7 @@ Future<String> getVersionListing(String package, {Uri? pubHostedUrl}) async {
   log.fine('Downloading: $url');
 
   return await retry(() => http.read(url),
-      retryIf: (e) => e is SocketException || e is TimeoutException);
+      retryIf: (e) => e is IOException || e is http.ClientException);
```
sigurdm, line 245: Are we sure we want to retry all ClientExceptions? Would it not be better to use http.get and retry depending on the error-code?

isoos: Added explicit status code check (and also timeout), PTAL.
We should have a `RetryException` in `package:retry`, wdyt @jonasfj?

lib/src/utils.dart
```diff
@@ -241,8 +241,19 @@ Future<String> getVersionListing(String package, {Uri? pubHostedUrl}) async {
...
-      retryIf: (e) => e is SocketException || e is TimeoutException);
+  return await retry(() async {
+    final rs = await http.get(url).timeout(const Duration(seconds: 20));
+    if (rs.statusCode == 200) {
+      return rs.body;
+    }
+    final message = '"$url" returned with status code ${rs.statusCode}.';
+    if (rs.statusCode >= 400 && rs.statusCode < 500) {
+      // does not retry on errors
+      throw AssertionError(message);
```
sigurdm, line 252: I would not use AssertionError here, that is specific to assertions (you would never expect it to be thrown when assertions are disabled.

I would still have a retryIf. Perhaps you can define a RetryException locally here?

isoos: Done. PTAL.

sigurdm (approved):
LGTM

PR #1018, opened 2022-02-15
Move repository_url.dart to src/repository

isoos:


isoos:
I'm planning to have more utils in `src/repository`, already working on the default branch name detection.

jonasfj (approved):

PR #1019, opened 2022-02-15
Detect default git branch with 'git remote' commands.

isoos:


isoos:
This is not used yet, and I'm planning to enable it in the near future only for internal packages, so that we can push out the need for strict isolation.

lib/src/repository/default_git_branch.dart
```diff
@@ -0,0 +1,55 @@
...
+    final remoteAddPr = await runProc(
+      ['git', 'remote', 'add', 'origin', baseUrl],
+      workingDirectory: dir,
+    );
+    if (remoteAddPr.exitCode != 0) {
+      log.warning(
+          'Failed to run `git remote add`.\n${remoteAddPr.asJoinedOutput}');
+    }
+    final remoteShowPr = await runProc(
+      ['git', 'remote', 'show', 'origin'],
```
jonasfj, line 34: Maybe, wrap with some retry, assuming this make network connections.

Or check in git manual that it does retry, and then leave a comment with a reference to documentation.

jonasfj (approved):

sigurdm (approved):

PR #1027, opened 2022-03-02
Git repository verification.

isoos:


isoos:
Updated to use git show instead of git checkout.

lib/src/repository/git_local_repository.dart
```diff
@@ -0,0 +1,159 @@
...
+}
+
+class GitLocalRepository {
+  final String localPath;
+  final String origin;
+
+  GitLocalRepository({
+    required this.localPath,
+    required this.origin,
+  });
```
jonasfj, line 41: Unless you are documenting all the invariants you rely on, this should be private.

lib/src/repository/git_local_repository.dart
```diff
@@ -0,0 +1,159 @@
...
+    } on GitToolException catch (e, st) {
+      log.warning(e.toString(), st);
+      return null;
+    } finally {
+      await repo?.delete();
+    }
+  });
+}
+
+class GitLocalRepository {
```
jonasfj, line 34: What is this, can we give a documentation comment? should we annotated it `@sealed`?

```suggestion
/// Interface for reading a remote git repository.
///
/// This objects uses a local temporary folder for interfacing with remote repository.
/// Hence, it is important to call [delete] or temporary files will be leaked.
class GitLocalRepository {
```

lib/src/repository/git_local_repository.dart
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+    }
+    throw GitToolException('Could not find HEAD branch.', output);
+  }
+
+  /// Return the String content of the file in [branch] and [path].
+  ///
+  /// Throws [GitToolException] if the git command fails.
+  Future<String> showStringContent(String branch, String path) async {
```
jonasfj, line 122: You should wrap `branch` and `path` some sanity checks...
 * Check they are not empty
 * Check that `branch` doesn't contain `/`
 * Sanity checks for special characters?

Throw `ArgumentError` if something is wrong, and document any non-obvious constraints.
A branch like `foo/bar` could certainly cause bugs to happen.

lib/src/repository/git_local_repository.dart
```diff
@@ -0,0 +1,159 @@
...
+    );
+  }
+
+  Future<void> _init() async {
+    await _runGit(['init']);
+    await _runGit(['remote', 'add', 'origin', origin]);
+  }
+
+  /// Detects the default branch name by querying the origin and
+  /// matching `HEAD branch` in the output.
```
jonasfj, line 96:
```suggestion
  /// Detects the default branch name (typically `master` or `main`)
  /// by querying the origin and matching `HEAD branch` in the output.
```

lib/src/repository/check_repository.dart
```diff
@@ -2,29 +2,77 @@
...
+
+  void failVerification(String message, [error, StackTrace? st]) {
+    verificationFailure = message;
+    log.info(message, error, st);
+  }
+
+  if (context.options.checkRemoteRepository) {
+    GitLocalRepository? repo;
+    try {
+      repo = await GitLocalRepository.createLocalRepository(url.baseUrl);
```
jonasfj, line 37:
```suggestion
    final repo = await GitLocalRepository.createLocalRepository(url.baseUrl);
    try {
```

jonasfj: it can't return a `repo` that should be deleted if it throws an exception -- so no need to enter try/finally before creating `repo`.

jonasfj (approved):

lib/src/repository/git_local_repository.dart
jonasfj: The public interface for `GitLocalRepository` does not enforce any patterns, or tell the users what strings can safely be used as `branch`.

It would be safer if we just limit branch names to something sane.

You have a good point that all callers of `GitLocalRepository.showStringContent` in the current code, does have some patterns protecting against branch names that contain `:` or `/`. But the method is public, and I assume we made an abstraction called `GitLocalRepository` because we imagine more things will use this in the future.

So adding an extra check inside `GitLocalRepository.showStringContent` protects us against such mistakes in the future.

lib/src/repository/check_repository.dart
jonasfj: Why does the declaration need to be outside the try-block?

isoos: the finally block still applies...
anyway, this method is now only used in tests, so we may just remove it and do similar wrapping in tests...

Issue #1031, opened 2022-03-09
If the package depend private packages on the self-hosted server, pana couldn't get the dependencies.


talisk:
I run the pana just like this.
```
pana --hosted-url <hosted-url>  --dart-sdk <dart-sdk> --flutter-sdk <flutter-sdk> -j --hosted <package_a> <version>
```

The package_a and its some dependencies are both hosted on the private server at `hosted-url`. Pana couldn't get the dependencies. Because pana will request `/api/packages` from `pub.dartlang.org`.

```json
{
    "level": "SHOUT",
    "message": "Problem analyzing <package_a> <version>",
    "error": "Exception: \"https://pub.dartlang.org/api/packages/<dependency_a>\" returned with status code 404.",
    "stackTrace": "#0      getVersionListing.<anonymous closure> (package:pana/src/utils.dart:253:9)\n<asynchronous suspension>\n#1      RetryOptions.retry (package:retry/retry.dart:131:16)\n<asynchronous suspension>\n#2      getVersionListing (package:pana/src/utils.dart:244:10)\n<asynchronous suspension>\n#3      computeOutdatedVersions (package:pana/src/report/dependencies.dart:295:37)\n<asynchronous suspension>\n#4      trustworthyDependency.dependencies (package:pana/src/report/dependencies.dart:51:15)\n<asynchronous suspension>\n#5      trustworthyDependency (package:pana/src/report/dependencies.dart:235:29)\n<asynchronous suspension>\n#6      createReport (package:pana/src/report/create_report.dart:46:5)\n<asynchronous suspension>\n#7      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:229:15)\n<asynchronous suspension>\n#8      PackageAnalyzer.inspectPackage.<anonymous closure>.<anonymous closure> (package:pana/src/package_analyzer.dart:87:16)\n<asynchronous suspension>\n#9      withTempDir (package:pana/src/utils.dart:233:12)\n<asynchronous suspension>\n#10     main (file:///Users/username/.pub-cache/hosted/<hosted-url>/pana-0.21.8/bin/pana.dart:160:19)\n<asynchronous suspension>\n"
}
```

PR #1032, opened 2022-03-09
Pass the `hosted-url` to PackageAnalyzer, read `PUB_HOSTED_URL` env f…

talisk:
…rom environment in `computeOutdatedVersions` for the self-hosted pub dependencies. (#1031)

### DIFF
--- a/lib/src/report/dependencies.dart
+++ b/lib/src/report/dependencies.dart
@@ -292,10 +292,18 @@ Future<List<OutdatedVersionDescription>> computeOutdatedVersions(
   }
   final result = <OutdatedVersionDescription>[];
 
-  final versionListing = jsonDecode(await getVersionListing(
-    name,
-    pubHostedUrl: hostedDependency.hosted?.url,
-  ));
+  Uri? environmentUri;
+  try {
+    // If 'PUB_HOSTED_URL' is not null, use it first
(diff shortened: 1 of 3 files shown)


lib/src/report/dependencies.dart
```diff
@@ -292,10 +292,18 @@ Future<List<OutdatedVersionDescription>> computeOutdatedVersions(
...
-  final versionListing = jsonDecode(await getVersionListing(
-    name,
-    pubHostedUrl: hostedDependency.hosted?.url,
-  ));
+  Uri? environmentUri;
+  try {
+    // If 'PUB_HOSTED_URL' is not null, use it first
+    final hostUrlString = context.toolEnvironment.environment['PUB_HOSTED_URL'];
+    if (hostUrlString != null) {
+      environmentUri = Uri.parse(hostUrlString);
```
isoos, line 300: `environmentUri = Uri.tryParse(hostUrlString);`
and let's drop the try-catch, as we in general don't like generic catch blocks, and the `environmentUri = null` is not needed either.

isoos (changes requested):

talisk:
@isoos I was wondering if you could review again. :)

jonasfj (approved):

PR #1038, opened 2022-03-17
Update SPDX corpus + license tagging.

isoos:
- updated SPDX lincese corpus
- emit `license:<license>` tag on all the detected licenses, not just the first one
- emit `license:osi-approved` and `license:fsf-libre` (when applicable, code generated from the `update_license.dart` tool

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -193,14 +193,16 @@ class PackageAnalyzer {
 
     final licenses = await detectLicenseInDir(pkgDir);
     if (licenses.isNotEmpty) {
-      tags.addAll(licenses.map((l) => 'license:${l.spdxIdentifier.toLowerCase()}').toSet());
-      if (licenses.all((l) => fsfLibreLicenses.contains(l.spdxIdentifier))) {
+      tags.addAll(licenses
+          .map((l) => 'license:${l.spdxIdentifier.toLowerCase()}')
+          .toSet());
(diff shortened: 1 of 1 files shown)


lib/src/package_analyzer.dart
```diff
@@ -192,7 +193,18 @@ class PackageAnalyzer {
...
+        licenses
+            .map((l) => l.spdxIdentifier)
+            .expand(
+              (id) => [
+                'license:${id.toLowerCase()}',
+                if (fsfLibreLicenses.contains(id)) 'license:fsf-libre',
+                if (osiApprovedLicenses.contains(id)) 'license:osi-approved',
+              ],
+            )
+            .toSet(),
```
jonasfj, line 206: To `expand + toSet` seems a bit extreme... maybe it's simpler to do:
```suggestion
      tags.addAll(licenses.map((l) => 'license:${l.spdxIdentifier.toLowerCase()}').toSet());
      if (licenses.all((l) => fsfLibreLicenses.contains(l.spdxIdentifier))) {
        tags.add('license:fsf-libre');
      }
      if (licenses.all((l) => osiApprovedLicenses.contains(l.spdxIdentifier))) {
        tags.add('license:osi-approved');
      }
```

Also shouldn't we require all licenses to be OSI approved before we grant:`'license:osi-approved'`

jonasfj (approved):

PR #1045, opened 2022-03-24
Improved runProc method + output size limit for pubspec.yaml from git

isoos:
- lazy `stdout` and `stderr` parsing for process output
- explicit `asString` and `asBytes` fields (still need to figure out how to handle errors)
- updated git content extraction to use output size limits


lib/src/repository/git_local_repository.dart
```diff
@@ -110,15 +114,22 @@ class GitLocalRepository {
   /// Return the String content of the file in [branch] and [path].
   ///
   /// Throws [GitToolException] if the git command fails.
-  Future<String> showStringContent(String branch, String path) async {
+  Future<String> showStringContent(
+    String branch,
+    String path, {
+    int? maxOutputBytes,
```
jonasfj, line 120: Where is `maxOutputBytes` used?

lib/src/repository/git_local_repository.dart
```diff
@@ -110,15 +114,22 @@ class GitLocalRepository {
   /// Return the String content of the file in [branch] and [path].
   ///
   /// Throws [GitToolException] if the git command fails.
```
jonasfj, line 116: What does it do if `maxOutputBytes` is exceeded, throw what exception or truncate the result.

jonasfj (approved):

lib/src/repository/git_local_repository.dart
isoos: Right now `runProc` truncates the results, prepends the `stderr` with `Output exceeded $maxOutputBytes bytes.` and also sets the `wasOutputExceeded` flag. As a result, the `git` operation will fail with a `GitToolException`.

Issue #1047, opened 2022-03-26
Dependency versions are not properly parsed


axel-op:
The versions of the dependencies are not correctly parsed during package resolution.

The reason is that in `pkg_resolution.dart`, the regular expression used to parse the PubEntry content is outdated, resulting in an empty `pkgVersions` map.

Stacktrace:

```console
SEVERE     Problem with `dart pub upgrade`
           Bad state: For [xxx], the parsed version null did not match the locked version 3.0.0+2.
           #0      _validateLockedVersions.<anonymous closure> (package:pana/src/pkg_resolution.dart:224:11)
           #1      SplayTreeMap.forEach (dart:collection/splay_tree.dart:604:8)
           #2      _validateLockedVersions (package:pana/src/pkg_resolution.dart:220:12)
           #3      createPkgResolution (package:pana/src/pkg_resolution.dart:206:5)
           #4      PackageContext.resolveDependencies (package:pana/src/package_context.dart:70:26)
           <asynchronous suspension>
           #5      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:149:27)
           <asynchronous suspension>
           #6      PackageAnalyzer.inspectDir.<anonymous closure> (package:pana/src/package_analyzer.dart:98:14)
           <asynchronous suspension>
           #7      withTempDir (package:pana/src/utils.dart:235:12)
           <asynchronous suspension>
           #8      main (file:///C:/Users/axelp/Documents/Git%20Repositories/pana/bin/pana.dart:173:19)
           <asynchronous suspension>
```

Pana version: 0.21.10
Dart SDK version: 2.17.0-229.0.dev (dev)

isoos:
Could you please provide a reproducible example (either a public package or a detailed pubspec with the problem)?

axel-op:
The problem is reproducible on packages with a pubspec containing a `dependencies` section.

For example, when running pana on [this package](https://github.com/SpinlockLabs/github.dart), I have the following output:

```console
INFO       Running `C:\tools\dart-sdk\bin\dart.exe --version`...
INFO       Running `flutter.bat --no-version-check --version --machine`...
INFO       Running `git rev-parse --show-toplevel`...
INFO       Running `C:\tools\dart-sdk\bin\dart.exe pub upgrade --verbose`...
SEVERE     Problem with `dart pub upgrade`
           Bad state: For async, the parsed version null did not match the locked version 2.8.2.
[...]

PR #1048, opened 2022-03-26
Update package resolution RegEx

axel-op:
Fixes #1047.

PR #1052, opened 2022-04-01
Use `pub outdated` for dependency resolution

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/1047
Supersedes: https://github.com/dart-lang/pana/pull/1048


test/goldens/end2end/sdp_transform-0.2.0.json
```diff
@@ -108,5 +94,6 @@
     "branch": "master",
     "isVerified": true
   },
-  "urlProblems": []
+  "urlProblems": [],
+  "errorMessage": "`dart pub get` failed:\n\n```\nOUT:\nResolving dependencies...\nERR:\npubspec.yaml has no lower-bound SDK constraint.\nYou should edit pubspec.yaml to contain an SDK constraint:\n\nenvironment:\n  sdk: '>=2.10.0 <3.0.0'\n\nSee https://dart.dev/go/sdk-constraint\n```"
```
isoos, line 98: Any idea why this has changed? It looks like something that shouldn't change if we keep the same SDK version (2.16.1 on GitHub Actions).

jonasfj: There is a distinction between whether or not we consume a package directly as root package, or as a dependency. Dependencies don't have to specify SDK constraints, but instead get an inferred constraint.

This is because some old packages were published without SDK constraints.

@sigurdm, this might be something we should do something about.. any ideas?

jonasfj (approved):

test/goldens/end2end/sdp_transform-0.2.0.json
sigurdm: Yeah not sure what is best here...

We could add the inferred constraint before running `pub outdated` if we need...

jonasfj: Or we create a dummy package in a temporary directory with a dependency on this package.
But then we have to depend on it from pub.dev, not using a path-dependency, because that could cause conflicts with cycles.

sigurdm: > Or we create a dummy package in a temporary directory with a dependency on this package.

I guess that is what we already do?

In that case it seems we are living with the cycle case...

jonasfj: If we depend on it locally we need to rewrite the pubspec.yaml, that's also fine...

sigurdm: Ah - no we do this:
```
  /// Removes the dev_dependencies from the pubspec.yaml
  /// Adds lower-bound minimal SDK constraint - if missing.
  /// Returns the backup file with the original content.
  Future<File> _stripAndAugmentPubspecYaml(String packageDir) async {
  ```
  
  I guess if we already are doing this we could just continue doing it (though it is ugly as muck).

isoos:
Closing as #1093 makes this obsolete.

Issue #1055, opened 2022-04-04
Score penalty for using self-deprecated members


rydmike:
# Score penalty for using self-deprecated members

It seems packages on pub.dev get a penalty for being nice and providing back fill/compatibility for self deprecated members and providing references to them in document comments.

Discussion reference: https://twitter.com/RydMike/status/1510958228154331138

## Rationale against this rule

Please consider if it would not be preferable if packages would be able to provide backwards compatibility for self-deprecated members when possible, instead of breaking APIs. In order to do so you have to use them internally while supporting backwards compatibility for an agreed period of time. Typically until next breaking release, only then is it completely removed. Providing package users ample time and grace period to migrate.

Also using the self deprecated members in document comments should be allowed so you can provide proper migration guidance.

## Current status

With current status a package gets a better score, if it is being less nice and break things and don't document it properly. This seems a bit counter intuitive to what there score should be doing.

## Possible solution

Would it be possible for `pana` to respect the analysis option.

```yaml
errors:
  deprecated_member_use_from_same_package: ignore
````

That Flutter SDK itself uses too. I think that would solve it nicely for package that are prudent and nice enough to offer back-fill and time scoped back-fill of deprecated members when possible.

mit-mit:
This appears to be coming from the analyzer as a diagnostic (and not from https://github.com/dart-lang/lints as first assumed):

https://dart.dev/tools/diagnostic-messages#deprecated_member_use_from_same_package

mit-mit:
Sorry it's taken so long to offer an update. This triggered a rather large discussion of errors, warnings, lints and hints, and we're working thorough some options to make some simplifications and improvements in that area.

About the original issue: we agree that the `[deprecated_member_use_from_same_package](https://dart.dev/tools/diagnostic-messages#deprecated_member_use_from_same_package)` rule is an "opinion" and this something that should either be entirely off, or in set of [recommended lints](https://pub.dev/packages/lints) (which aren't scored on).

We're working through that change. As mentioned above this is heppening today as `deprecated_member_use_from_same_package` is coming from the [analyzer diagnostics](https://dart.dev/tools/diagnostic-messages#diagnostics).

And then finally my questions: are there other rules form that list of [analyzer diagnostics](https://dart.dev/tools/diagnostic-messages#diagnostics) that you feel shouldn't be 1) on by default, and 2) scored on?

rydmike:
@mit-mit thanks! This looks good. 

[...]

PR #1057, opened 2022-04-05
Default analysis options: ignore deprecated from same package.

isoos:
#1055

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.21.13
+
+- `deprecated_member_use_from_same_package: ignore` set to default analysis options
+
 ## 0.21.12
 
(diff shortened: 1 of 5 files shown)


jonasfj (commented):
I'm not sure we should take an opinion on this.

jonasfj:
@mit-mit did you reach a conclusion after digging into diagnostics vs lints?

mit-mit:
Not yet. This has evolved into a potentially very large restructuring of how we do this, so it's taking a little while.

But I'm very supportive of temporarily solving this by ignoring this in scoring.

jonasfj (approved):

isoos:
As discussed today, not merging this for now.

PR #1058, opened 2022-04-08
Check all pubspec.yaml during repository verification.

isoos:
- detects if multiple `pubspec.yaml` has the same package name, which may be a copy-paste error or a fraud attempt
- exposes the matched package path in the result data

### DIFF
--- a/lib/src/repository/check_repository.dart
+++ b/lib/src/repository/check_repository.dart
@@ -52,16 +52,17 @@ Future<Repository?> checkRepository(PackageContext context) async {
     late GitLocalRepository repo;
     try {
       repo = await GitLocalRepository.createLocalRepository(url.baseUrl);
-      branch ??= await repo.detectDefaultBranch();
+      branch = await repo.detectDefaultBranch();
 
       // list all pubspec.yaml files
       final files = await repo.listFiles(branch);
-      // TODO: verify file name patterns
+      // TODO: verify all file name patterns
 
       final pubspecFiles =
           files.where((path) => p.basename(path) == 'pubspec.yaml').toList();
       if (pubspecFiles.isEmpty) {
-        failVerification('Could not find any `pubspec.yaml` in the repository.');
+        failVerification(
+            'Could not find any `pubspec.yaml` in the repository.');
         return result();
       }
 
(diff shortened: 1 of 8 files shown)


test/goldens/end2end/url_launcher-6.0.3.json
```diff
@@ -164,7 +164,7 @@
     "branch": "master",
     "packagePath": "packages/url_launcher/url_launcher",
     "isVerified": false,
-    "verificationFailure": "Unable to parse `pubspec.yaml` from git repository."
+    "verificationFailure": "Repository `packages/url_launcher/url_launcher/pubspec.yaml` URL missmatch: `https://github.com/flutter/plugins/tree/main/packages/url_launcher/url_launcher` != `https://github.com/flutter/plugins/tree/master/packages/url_launcher/url_launcher`."
```
jonasfj, line 167: Is this really a problem?

test/goldens/end2end/audio_service-0.17.0.json
```diff
@@ -173,8 +173,9 @@
   "repository": {
     "baseUrl": "https://github.com/ryanheise/audio_service",
     "branch": "minor",
+    "packagePath": "audio_service",
     "isVerified": false,
-    "verificationFailure": "Repository has no `pubspec.yaml` in location: `/`."
+    "verificationFailure": "Repository `audio_service/pubspec.yaml` URL missmatch: `https://github.com/ryanheise/audio_service/tree/master/audio_service` != `https://github.com/ryanheise/audio_service`."
```
jonasfj, line 178: Do we want to expect deep links?

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,81 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+              false,
+              'Repository `$path` name missmatch: '
+              '`${gitPubspec.name}` != `${context.pubspec.name}`.');
+        } else if (gitPubspec.repositoryOrHomepage !=
+            context.pubspec.repositoryOrHomepage) {
+          return _PubspecMatch(
+              path,
+              true,
+              'Repository `$path` URL missmatch: '
+              '`${gitPubspec.repositoryOrHomepage}` != `${context.pubspec.repositoryOrHomepage}`.');
```
jonasfj, line 89: This message is not good. It's unclear we found and what we expected.

jonasfj (changes requested):

lib/src/repository/check_repository.dart
```diff
@@ -43,7 +49,7 @@ Future<Repository?> checkRepository(PackageContext context) async {
 
   if (context.options.checkRemoteRepository) {
     isVerified = false;
-    GitLocalRepository? repo;
+    late GitLocalRepository repo;
     try {
       repo = await GitLocalRepository.createLocalRepository(url.baseUrl);
       branch ??= await repo.detectDefaultBranch();
```
jonasfj, line 55: Shouldn't we do `branch = await repo.detectDefaultBranch();` ?

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,86 @@ Future<Repository?> checkRepository(PackageContext context) async {
         failVerification('Repository has no `pubspec.yaml`.');
```
jonasfj, line 64:
```suggestion
        failVerification('Could not find any `pubspec.yaml` in the repository.');
```

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,86 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+          maxOutputBytes: _maxPubspecBytes,
+        );
+        final gitPubspec = Pubspec.parseYaml(content);
+        // verification steps
+        if (gitPubspec.name != context.pubspec.name) {
+          return _PubspecMatch(
+              path,
+              false,
+              'Repository `$path` name missmatch: '
+              '`${gitPubspec.name}` != `${context.pubspec.name}`.');
```
jonasfj, line 82: What did you find, what did you expect?

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,86 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+        final gitRepoUrl = RepositoryUrl.tryParse(gitRepoOrHomepage);
+        if (gitRepoUrl?.baseUrl != url.baseUrl) {
+          return _PubspecMatch(path, true,
+              'Repository `$path` URL missmatch: expected `${url.baseUrl}` but got `${gitRepoUrl?.baseUrl}`.');
+        }
+        if (gitPubspec.version == null) {
+          return _PubspecMatch(
+              path, true, 'Repository `$path` has no version.');
+        }
+        if (gitPubspec.toJson().containsKey('publish_to ')) {
```
jonasfj, line 98: Whitespace is probably an issue here :D
```suggestion
        if (gitPubspec.toJson().containsKey('publish_to')) {
```

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,86 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+          return _PubspecMatch(
+              path, true, 'Repository `$path` has `publish_to`.');
```
jonasfj, line 100:
```suggestion
              path, true, '`$path/pubspec.yaml` from the repository defines `publish_to`, thus, we are unable to verify the package is published from here.');
```

jonasfj: generally, these message probably need more explanations.

lib/src/repository/check_repository.dart
```diff
@@ -58,47 +64,86 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
-        return result();
+
+      Future<_PubspecMatch> matchRepoPubspecYaml(String path) async {
+        // checkout the pubspec.yaml at the assumed location
+        final content = await repo.showStringContent(
+          branch!,
+          path,
+          maxOutputBytes: _maxPubspecBytes,
+        );
+        final gitPubspec = Pubspec.parseYaml(content);
```
jonasfj, line 75: We should probably handle `FormatException` here, so that we don't fail because some invalid pubspec.yaml exist in the repository.

isoos: I'd start with the current strict approach, and see how it goes. I'll add a TODO to revisit it later.

jonasfj (approved):

PR #1060, opened 2022-04-11
Expose repository verification failures in the report (and score).

isoos:
- The score of valid `pubspec.yaml` (10 points) is awarded only if there is no repository verification failure (either because it was not attempted, or it failed).
- also simplified `PackageContext` a bit..


test/goldens/end2end/_dummy_pkg-1.0.0-null-safety.1.json_report.md
```diff
@@ -2,6 +2,15 @@
...
 
+<details>
+<summary>
+Failed to verify repository URL.
+</summary>
+
+Please provide a valid repository URL.
+
+`pkg/pub_integration/test_data/_dummy_pkg/pubspec.yaml` from the repository has no `version`.
+</details>
```
jonasfj, line 13: Look at this, I think we should reorder it, so things like bad SDK constraints comes before repository verification.

I also think we need to improve the message a bit.

We could say things like:
 * Found repository URL `https://github.com/dart-lang/pub-dev` in `homepage` property of `pubspec.yaml`.
 * Cloned repository and found `pkg/pub_integration/test_data/_dummy_pkg/pubspec.yaml` with the correct package name, but no `version` property, this it is unlikely this package was published from this repository.

lib/src/report/template.dart
```diff
@@ -171,6 +171,14 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
         isRequired: pubspec.homepage == null));
     issues.addAll(await findUrlIssues('documentation', 'Documentation URL'));
     issues.addAll(await findUrlIssues('issue_tracker', 'Issue tracker URL'));
+
+    final repository = await context.repository;
+    if (repository?.verificationFailure != null) {
+      issues.add(Issue('Failed to verify repository URL.',
+          suggestion:
+              'Please provide a valid repository URL.\n\n${repository!.verificationFailure}'));
```
jonasfj, line 179:
```suggestion
              'Please provide a valid `repository` URL in `pubspec.yaml`, such that:\n\n'
              ' * `repository` can be cloned,\n'
              ' * a clone of the repository contains a `pubspec.yaml`, which:,\n'
              '    * contains `name: ${packageName}`,\n'
              '    * contains a `version` property, and,\n'
              '    * does not contain a `publish_to` property.\n\n'
              '${repository!.verificationFailure}'));
```

jonasfj: Maybe, we should also link to:
https://dart.dev/tools/pub/pubspec#repository

Or we could create a `dart.dev/go/pubspec-repository` in:
https://github.com/dart-lang/site-www/blob/main/firebase.json#L137-L156
(if we want a perma-link we can fix in the future)

jonasfj (approved):

lib/src/report/template.dart
isoos: Updated message to have an inline link to the `pubspec#repository` for now.

PR #1062, opened 2022-04-20
Fix repository verification exceptions.

isoos:


lib/src/repository/check_repository.dart
```diff
@@ -73,8 +74,34 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+        //       unrelated, but badly formatted pubspec.yaml in the repository.
+        // ignore: prefer_typing_uninitialized_variables
+        var yamlDoc;
+        try {
+          yamlDoc = yaml.loadYaml(content);
+        } on yaml.YamlException catch (e, st) {
+          log.info('Invalid yaml file: $path', e, st);
+        } on ArgumentError catch (e, st) {
+          log.info('Invalid yaml file: $path', e, st);
+        }
```
jonasfj, line 91: Don't you need to handle `FormatException` too?

lib/src/repository/check_repository.dart
```diff
@@ -73,8 +74,34 @@ Future<Repository?> checkRepository(PackageContext context) async {
...
+        if (yamlDoc == null) {
+          return _PubspecMatch(path, false,
+              '`$path` from the repository is not a valid YAML document.');
+        }
+
+        late final Pubspec gitPubspec;
+        try {
+          gitPubspec = Pubspec.parseYaml(content);
+        } catch (e, st) {
```
jonasfj, line 100: Maybe only handle `FormatException` what can this really throw?

jonasfj (approved):

lib/src/repository/check_repository.dart
isoos: Done.

lib/src/repository/check_repository.dart
isoos: Done.

jonasfj (approved):

PR #1067, opened 2022-05-30
Produce basic JSON summary of a specified package

CicadaCinema:
This will include the names of public symbols and their kind (function, variable, class, etc)

### DIFF
--- a/lib/src/package_analysis.dart
+++ b/lib/src/package_analysis.dart
@@ -1,0 +1,60 @@
+import 'dart:io';
+
+import 'package:analyzer/dart/analysis/context_builder.dart';
+import 'package:analyzer/dart/analysis/context_locator.dart';
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:args/args.dart';
+import 'package:path/path.dart' as path;
+
+Future<void> main(List<String> arguments) async {
+  // parse command-line arguments
+  final argResults = ArgParser().parse(arguments);
+  final paths = argResults.rest;
+
+  if (paths.length != 1) {
+    _exitFatal('Please specify a exactly one directory for analysis.', 1);
+  }
+
+  final packageLocation = paths.first;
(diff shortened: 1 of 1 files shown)


lib/src/package_analysis.dart
```diff
@@ -0,0 +1,60 @@
...
+
+import 'package:analyzer/dart/analysis/context_builder.dart';
+import 'package:analyzer/dart/analysis/context_locator.dart';
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:args/args.dart';
+import 'package:path/path.dart' as path;
+
+Future<void> main(List<String> arguments) async {
+  // parse command-line arguments
```
CicadaCinema, line 11: The analyser will have a number of distinct functions (creating a JSON summary, comparing two JSON summaries, comparing two package versions directly), so it probably makes sense to introduce subcommands for each of these functions

lib/src/package_analysis.dart
```diff
@@ -0,0 +1,60 @@
...
+      // use canonicalize to ensure we get an absolute path
+      includedPaths: [path.canonicalize(packageLocation)],
+    ).first,
+  );
+
+  for (final filePath in context.contextRoot.analyzedFiles()) {
+    final session = context.currentSession;
+    final someResolvedLibrary = await session.getResolvedLibrary(filePath);
+
+    // match [packageLocation]/lib/*.dart
```
CicadaCinema, line 36: There are 3 kinds of libraries in a package which can be imported:
1. `lib/my_package.dart` (we usually import this library when we want to use the package)
2. `lib/*.dart` (could be in a sub-folder of `lib`, eg `lib/feature/*.dart`)
3. `lib/src/*.dart` (not in a sub-folder)

Even type 3 libraries can be imported for use, but it might make sense to exclude these libraries from analysis (by default, adding a switch to disable this).

CicadaCinema:
WHOOPS, wrong repo

PR #1068, opened 2022-06-03
Verified URLs in dedicated AnalysisResult object.

isoos:


lib/src/model.dart
```diff
@@ -337,6 +340,32 @@ class ReportSection {
   Map<String, dynamic> toJson() => _$ReportSectionToJson(this);
 }
 
+/// NOTE: the content of the class is experimental, clients should not rely on it yet.
+/// Contains the verified and normalized outside references:
```
jonasfj, line 344:
```suggestion
/// Contains the verified and normalized external references:
```

lib/src/model.dart
```diff
@@ -337,6 +340,32 @@ class ReportSection {
...
+/// - when a value is present, it has been tried and verified to be working,
+/// - when a value is absent, it was either absent or its verification failed.
+///
+/// Failed verification reasons are presented in the [Report].
```
jonasfj, line 348: Please outline what "verified" means for the different URLs

jonasfj (approved):

PR #1072, opened 2022-06-20
Use the stripped/augmented pubspec.yaml for 'pub outdated'

### LINKED ISSUE #1069: Support "up to date dependencies" fails on dev only path dependencies
See the score page for [build_runner](https://pub.dev/packages/build_runner/score), which gives the following error:

```
OUT:
Resolving dependencies...
ERR:
Because build_runner depends on _test_common from path which doesn't exist (could not find package _test_common at "../_test_common"), version solving failed.
```

The _test_common package is a local-only path dependency, only visible in `dev_dependencies`. It should not be affecting the score negatively?

isoos:
- Fixes #1069.
- Re-implements part of #1052, without the actual reuse (which should be a separate PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.21.13-dev
 
 - Populate the pubspec's `repository` field.
+- Fix: Use the stripped/augmented `pubspec.yaml` for `pub outdated`.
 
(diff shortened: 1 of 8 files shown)


sigurdm (approved):
LGTM

PR #1073, opened 2022-06-22
Remove most of the parsing of 'pub upgrade'.

isoos:
- Creates the `allDependencies` list in cases where pub upgrade failed. (It can be argued both ways whether we should this, but I think it is better to have the list of listed dependencies there).
- Follows the idea from #1052 - we don't need the output of `pub upgrade`, we can use the less error-prone output of `pub outdated`.
- Keeping the `pub upgrade` for now, merging the two in a follow-up PR, as it will alter some of the error messages.


sigurdm (approved):
LGTM

jonasfj (approved):

PR #1078, opened 2022-06-24
Fix: internal hostname check using full-length match.

### LINKED ISSUE #1076: Score message: 'Documentation URL isn't helpful.' is unhelpful :)
I'm  trying to clean up my score on pub.dev for https://pub.dev/packages/onepub/versions/1.2.2

I'm getting  a message on the score page that reads:

```
0/10 points: Provide a valid pubspec.yaml
Documentation URL isn't helpful.
```
My documentation URL is: https://docs.onepub.dev

The message gives no clue as to what the problem is and my link goes to the onepub manual which would seem to be appropriate, so I have no idea how to resolve the issue.

isoos:
- Fixes #1076.
- Adds `flutter.io` to the list of internal hosts.
- Simplify setup to expand host names to also include subdomains.
- Checks the entire string to match, instead of the prior substring comparison.


lib/src/download_utils.dart
```diff
@@ -114,28 +114,37 @@ class UrlChecker {
...
+  ///
+  /// When a host is [String], we will expand the rule to include:
+  /// - exact hostname match (`^$host$`), and
+  /// - subdomain hostname match (`^.+\.$host$`).
   void addInternalHosts(Iterable<Pattern> hosts) {
-    _internalHosts.addAll(hosts);
+    _internalHosts.addAll(hosts.expand((host) {
+      if (host is String) {
+        return [
+          RegExp('^$host\$'),
```
sigurdm, line 136: I think you need to use `https://api.dart.dev/stable/2.17.5/dart-core/RegExp/escape.html` to escape the string before inserting it in a RegExp - otherwise 'pub.dev' would also match 'pubAdev

Also in the line below.

isoos: Good catch, fixed it + added test.

isoos:
> Should we also improve the message "is not helpful" is indeed not helpful (could be another PR)

I've changed the message to `is pointing to a domain reserved for internal packages only.`

sigurdm (approved):
LGTM

PR #1079, opened 2022-06-24
Predefined platform assignment for binary-only packages.

isoos:
- #1077
- the assigned platforms are the ones with Dart SDK release
- submitting with two commits so that we can see the diff of the change
- I'm not yet sure if we should reduce the score if the `platforms:` in `pubspec.yaml` is in conflict with these predefined platforms, postponing to a later PR to follow-up.


lib/src/tag/tagger.dart
```diff
@@ -184,8 +186,8 @@ class Tagger {
       if (_isBinaryOnly) {
         tags.add(Sdk.dart.tag);
         explanations.add(Explanation('Binary only',
-            'Cannot assign flutter SDK tag because it is binary only',
-            tag: null));
+            'Cannot assign Flutter SDK tag because it is binary only.',
```
sigurdm, line 189: We should probably also reconsider this...

A binary-only package can usually be used as a tool with the flutter sdk...

isoos: I'm all for this! It seems to be better to be optimistic with the scope / capabilities of such packages, and if we get reports that something should be otherwise, we should add checks for conflicts later on...

sigurdm (approved):
LGTM

PR #1080, opened 2022-06-24
Code cleanup in dependency resolution.

isoos:
- Follow-up to #1073.
- The try-catch block is a no-op.
- There is no point of checking `exitCode` > 0, we know it is != 0 already.


sigurdm (approved):
LGTM

PR #1083, opened 2022-07-01
Expose 'RepositoryProvider' in the analysis result ('Repository.provider')

isoos:
This will be needed for URL rewrite rules (we shouldn't even try to rewrite URLs that are not known providers).

### DIFF
--- a/lib/src/model.dart
+++ b/lib/src/model.dart
@@ -377,20 +377,38 @@ class AnalysisResult {
 ///   - must not have a `publish_to` key.
 @JsonSerializable()
 class Repository {
+  final String provider;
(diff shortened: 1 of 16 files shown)


lib/src/model.dart
```diff
@@ -377,22 +377,40 @@ class AnalysisResult {
 ///   - must not have a `publish_to` key.
 @JsonSerializable()
 class Repository {
+  final String provider;
```
sigurdm, line 380: Would it make sense to represent this as an enum?

https://pub.dev/packages/json_serializable#enums

sigurdm (approved):
LGTM

PR #1084, opened 2022-07-04
Merge RepositoryUrl into Repository, expand URL resolver.

isoos:
- Refactored `Repository` fields to better separate the parts of a repository.
- Merged `RepositoryUrl` into `Repository`, there is very little reason to keep the two around.
- Expanded the URL resolution function: it should handle all kinds of links, but also prevents invalid links to be created.


sigurdm (approved):
LGTM

PR #1091, opened 2022-07-19
Remove 'pubspec_overrides.yaml' when running 'pub upgrade' or 'pub outdated'.

### LINKED ISSUE #1089: pana does not handle `pubspec_overrides.yaml`
`freezed` uses a **pubspec_overrides.yaml** file, that overrides to the local `freezed_annotation` during development
```yaml
dependency_overrides:
  freezed_annotation:
    path: ../freezed_annotation
```

But this results in `pana --hosted freezed` to not be able to resolve those dependencies during **✗ Pass static analysis**.

### How to reproduce:
```sh
dart pub global activate pana
pana --hosted freezed
```
Points: 40/120.
```sh
git clone https://github.com/rrousselGit/freezed.git
pana freezed/packages/freezed
```
Points: 100/120.

isoos:
Fixes #1089

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -468,18 +468,36 @@ class ToolEnvironment {
     return DartdocResult(pr, pr.exitCode == 15, hasIndexHtml, hasIndexJson);
   }
 
-  /// Removes the dev_dependencies from the pubspec.yaml
+  /// Removes the `dev_dependencies` from the `pubspec.yaml`,
+  /// and also removes `pubspec_overrides.yaml`.
+  ///
   /// Adds lower-bound minimal SDK constraint - if missing.
-  /// Returns the backup file with the original content.
+  ///
+  /// Returns the result of the inner function, and restores the
+  /// original content upon return.
   Future<R> _withStripAndAugmentPubspecYaml<R>(
     String packageDir,
     FutureOr<R> Function() fn,
   ) async {
-    final backup = await _stripAndAugmentPubspecYaml(packageDir);
+    final now = DateTime.now();
+    final pubspecBackup = await _stripAndAugmentPubspecYaml(packageDir);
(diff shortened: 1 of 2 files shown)


lib/src/sdk_env.dart
```diff
@@ -468,18 +468,36 @@ class ToolEnvironment {
...
     FutureOr<R> Function() fn,
   ) async {
-    final backup = await _stripAndAugmentPubspecYaml(packageDir);
+    final now = DateTime.now();
+    final pubspecBackup = await _stripAndAugmentPubspecYaml(packageDir);
+
+    // backup of pubspec_overrides.yaml
+    final poyamlFile = File(p.join(packageDir, 'pubspec_overrides.yaml'));
+    File? poyamlBackup;
+    if (await poyamlFile.exists()) {
```
sigurdm, line 488: Usually the "-sync" methods are both faster and more convenient
```suggestion
    if (poyamlFile.existsSync()) {
```

sigurdm (approved):
Ok for now - but I don't like all this messing around with the files of the package. It seems fragile to me.

Moving forward we should really either:
* Work on a copy of the package
* Make a dummy-package depending on the package.

isoos:
> Work on a copy of the package

We already do the analysis on a copy:
https://github.com/dart-lang/pana/blob/master/lib/src/package_analyzer.dart#L96-L98

> Make a dummy-package depending on the package

Yeah, that would also be nice.

lib/src/sdk_env.dart
isoos: Keeping it for consistency, and also we don't want to block the analysis VMs (yet, until sandboxing at least).

sigurdm:
> We already do the analysis on a copy: [...]

Nice - I had forgotten!

PR #1096, opened 2022-07-21
Grant 10 points for using only OSI-approved licenses.

### LINKED ISSUE #1037: Grant 10 points for successful license detection
I propose that we grant 10 points to packages with a `LICENSE` file from which we can detect a license.

We are currently to detect the license of most packages, see:
https://github.com/dart-lang/pub-dev/issues/4373#issuecomment-1042778403

Investigation of which license we fail to detect is filed as https://github.com/dart-lang/pana/issues/1015

Packages that don't use one of the licenses from https://github.com/dart-lang/pana/tree/master/lib/src/third_party/spdx/licenses won't get these 10 points. That is unfortunate for packages with a commercial license. But we could say that if the `LICENSE` file starts with the string:
> `**THIS PACKAGE REQUIRES A COMMERCIAL LICENSE**\n` (followed by whatever license the package wants to use)

Then we detect the license as `license:commercial` or `license:proprietary` and grant the 10 points. So we simply have a string that indicates the package is commercial and we shouldn't do license detection for it.

When license detection fails, we can link to a page on dart.dev that explains:
 * How license detection on `LICENSE` works, and how to run pana's license detection locally.
 * How opt-out of license detection by marking a package a commercial (using the magic string).

9TSBF86jmQEBGSa (1)

-----------------------

**Questions:**
 * What string should we use in `LICENSE` to indicate a proprietary / commercial package?
    * Option: `**THIS PACKAGE IS COMMERCIALLY LICENSED**\n`
    * Option: `**THIS PACKAGE REQUIRES A COMMERCIAL LICENSE**\n`
 * Should we use the term `license:commercial` in the UI?
    * Option: `license:proprietary`
    * Option: `license:commercial`

isoos:
- Fixes #1037.
- Grants 10 points for package with only OSI-approved licenses.
- Grants 5 points for packages with detected SPDX license but having non-OSI-approved ones.
- Grants 0 points if no SPDX license was detected or license is missing.


test/goldens/end2end/http-0.13.0.json_report.md
```diff
@@ -1,4 +1,4 @@
-## 20/20 Follow Dart file conventions
+## 30/20 Follow Dart file conventions
```
sigurdm, line 1: 30/20 looks too good to be true. Should this not be 30/30? :)

lib/src/report/template.dart
```diff
@@ -261,3 +267,42 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+  if (tags.isOnlyOsiApproved) {
+    paragraphs = [detected];
+    grantedPoints = maxPoints;
+    reportStatus = ReportStatus.passed;
+  } else if (tags.wasDetected) {
+    paragraphs = [
+      detected,
+      Issue(
+        'The following ${licensePluralized(tags.nonOsiApprovedLicenses)} are not OSI-approved: ${licenseList(tags.nonOsiApprovedLicenses)}.',
+        suggestion: 'Use only OSI-approved license in the `LICENSE` file.',
```
sigurdm, line 294: This suggestion somehow comes out a bit harsh - the licensing of a package is the authors. Not sure how to make it better, maybe:
```suggestion
        suggestion: 'Consider using an OSI-approved license in the `LICENSE` file to share with the community.',
```

lib/src/report/template.dart
```diff
@@ -261,3 +267,42 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+      ),
+    ];
+    grantedPoints = maxPoints ~/ 2;
+    reportStatus = ReportStatus.partial;
+  } else {
+    paragraphs = [
+      Issue(
+        'No license was recognized.',
+        suggestion: 'Use an OSI-approved license in the `LICENSE` file.',
```
sigurdm, line 303: Same thing:
```suggestion
        suggestion: 'Consider using an OSI-approved license in the `LICENSE` file to share with the community.',
```

sigurdm (commented):
I'm not really opposed to this - but I think we will get some backlash for rewarding fewer points to commercial packages. Maybe a badge would be better?

test/goldens/end2end/http-0.13.0.json_report.md
isoos: I'm not sure why `makeSection` has a fixed `maxPoints` parameter, shouldn't we rather just use the sum of subsection.maxPoints instead?

sigurdm: I vaguely remember there being some reason for this - but now I cannot remember why...

lib/src/report/template.dart
```diff
@@ -251,13 +252,59 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
-    maxPoints: 20,
+    maxPoints: 30,
     subsections: subsections,
     basePath: packageDir,
     maxIssues: 10,
   );
 }
+
+Subsection _licenseSection(LicenseTags tags) {
+  const title = 'Uses OSI-approved license';
```
sigurdm, line 272: To be aligned with the other headings this should be in the imperative:
```suggestion
  const title = 'Use an OSI-approved license';
```

lib/src/report/template.dart
```diff
@@ -251,13 +252,59 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+    paragraphs = [detected];
+    grantedPoints = maxPoints;
+    reportStatus = ReportStatus.passed;
+  } else if (tags.wasDetected) {
+    paragraphs = [
+      detected,
+      Issue(
+        'The following ${licensePluralized(tags.nonOsiApprovedLicenses)} are not OSI-approved: ${licenseList(tags.nonOsiApprovedLicenses)}.',
+        suggestion:
+            'Consider using an OSI-approved license in the `LICENSE` file to share with the community.',
```
sigurdm, line 295: Let's link to documentation about this.

Is https://opensource.org/licenses-draft the right place to look for a list of licenses?

isoos: https://opensource.org/licenses may be better for it

lib/src/tag/license_tags.dart
```diff
@@ -0,0 +1,47 @@
...
+  late final isOnlyOsiApproved = tags.contains(_licenseOsiApprovedTag);
+  late final wasDetected = !tags.contains(_licenseUnknownTag);
+
+  late final osiApprovedLicenses =
+      licenses.where((l) => l.isOsiApproved).toList();
+  late final nonOsiApprovedLicenses =
+      licenses.where((l) => !l.isOsiApproved).toList();
+}
+
+LicenseTags createLicenseTags(List<License> licenses) {
```
jonasfj, line 26: Can this be a constructor or static method on `LicenseTags`, like `LicenseTags.fromLicenses`

lib/src/report/template.dart
```diff
@@ -251,13 +252,61 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+    ];
+    grantedPoints = maxPoints ~/ 2;
+    reportStatus = ReportStatus.partial;
+  } else {
+    paragraphs = [
+      Issue(
+        'No license was recognized.',
+        suggestion:
+            'Consider using an [OSI-approved license](https://opensource.org/licenses)'
+            ' in the `LICENSE` file to share with the community.',
```
jonasfj, line 307: Shouldn't this hint that we can recognize most SPDX licenses?

jonasfj (approved):

lib/src/report/template.dart
jonasfj: maybe it's fine... saying SPDX is also a bit of problem if you are using this as advice for finding a license.

PR #1101, opened 2022-08-01
fix: Non-default branch

dnys1:
Fixes the case where `repositoryUrl` points to a non-default branch. Before, `checkRepository` would overwrite this value with the default branch of the repo, causing some checks to fail if the package was not being tracked in the repo's main branch. This can be the case when publishing packages off a feature/next branch before merging to main.

This can be seen in the updated e2e golden for [audio_service](https://pub.dev/packages/audio_service) who's default branch is `minor`, but who's `repositoryUrl` points to `master`.

### DIFF
--- a/lib/src/repository/check_repository.dart
+++ b/lib/src/repository/check_repository.dart
@@ -70,7 +70,7 @@ Future<VerifiedRepository?> checkRepository(PackageContext context) async {
     late GitLocalRepository repo;
     try {
       repo = await GitLocalRepository.createLocalRepository(url.cloneUrl);
-      branch = await repo.detectDefaultBranch();
+      branch ??= await repo.detectDefaultBranch();
 
       // list all pubspec.yaml files
       final files = await repo.listFiles(branch);
--- a/test/goldens/end2end/audio_service-0.18.4.json
+++ b/test/goldens/end2end/audio_service-0.18.4.json
@@ -148,7 +148,7 @@
       "provider": "github",
       "host": "github.com",
       "repository": "ryanheise/audio_service",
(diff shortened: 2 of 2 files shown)


isoos:
I think hosting a published package in a non-default branch is confusing to a user that is not familiar with the project (e.g. it is just a few clicks away to lose the branch context). I'd assume that development of such package may happen in a development branch, but at the point where it is ready to be published, it will be merged into the default branch.

We are also detecting the path inside the repository, in order to verify if the package has no two `pubspec.yaml` inside, in order to prevent a kind of malicious copying attempts. Detecting the default branch and using it without any override aligns better with this approach.

With the two points above, I think we shouldn't do this. /cc @jonasfj @sigurdm

sigurdm:
I agree wih @isoos that the default flow should be publication from the default branch.

I can see a use-case where a fork of the package is maintained in a non-default branch in the same repo.
Such that `package:foo` is in the `main` branch and `package:foo_fork` is in the `fork` branch. But I'm not sure if that is a scenario we are eager to support...

dnys1:
Thanks for the quick reply. I tend to agree. And, for what it's worth, my use case is a little unique. We are publishing our preview versions off a non-default branch. Some of these preview versions require shared packages which are not required for our stable/main track packages. These shared packages were published for the first time as part of the preview release and we had not intended to merge them to our stable (default) branch until we considered them such.

The reasons you outline make sense and, especially if there are security concerns, I would say that it's not worth the impact to pana scores. I just wanted to share the context.

isoos:
Closing this for now, as the special case described here is not something we want to support right away.

PR #1104, opened 2022-08-03
Explain HEAD request for URL checks.

### LINKED ISSUE #1099: Explain what homepage not available means
- https://github.com/dart-lang/pub-dev/issues/5927
- need to explain that we only issue a `HEAD` request

isoos:
Fixes #1099.

### DIFF
--- a/lib/src/references/pubspec_urls.dart
+++ b/lib/src/references/pubspec_urls.dart
@@ -148,7 +148,8 @@ Future<UrlWithIssue> _checkUrl(
       Issue(
         "$name doesn't exist.",
         span: tryGetSpanFromYamlMap(pubspec.originalYaml, key),
-        suggestion: 'At the time of the analysis `$url` was unreachable.',
+        suggestion: 'At the time of the analysis `$url` was unreachable. '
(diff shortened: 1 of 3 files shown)


lib/src/references/pubspec_urls.dart
```diff
@@ -148,7 +148,8 @@ Future<UrlWithIssue> _checkUrl(
       Issue(
         "$name doesn't exist.",
         span: tryGetSpanFromYamlMap(pubspec.originalYaml, key),
-        suggestion: 'At the time of the analysis `$url` was unreachable.',
+        suggestion: 'At the time of the analysis `$url` was unreachable. '
+            'Make sure that the website is reachable via `HEAD` requests.',
```
sigurdm, line 152: Can we point to documentation, perhaps:
```suggestion
            'Make sure that the website is reachable via [`HEAD`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) requests.',
```

sigurdm (approved):
LGTM with a small suggestion

PR #1107, opened 2022-08-05
License text preprocessing to detect ISC license.

### LINKED ISSUE #1071: ISC License
Please add support for ISC licenses.
Currently they are shown as unknown, for example here: https://pub.dev/packages/weak_cache/license

isoos:
- Fixes #1071.
- The current code produced a high score, but not enough to cross the threshold. Removing the known patterns from all the licenses (both the patterns and the input in question) should increase the confidence for all kind of licenses.
- Also did small cleanups as I was debugging the code.

### DIFF
--- a/lib/src/license_detection/license.dart
+++ b/lib/src/license_detection/license.dart
@@ -4,6 +4,10 @@
 
 part of 'license_detector.dart';
 
+/// Regular expression to match the common pattern of `Copyright (c) yyyy[-yyyy] <name>.`
+final _copyrightRegExp =
+    RegExp(r'^copyright.+\d{4}.{0,100}$', caseSensitive: false);
+
 @sealed
 class License {
   /// SPDX identifier of the license, is empty in case of unknown license.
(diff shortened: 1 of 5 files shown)


lib/src/license_detection/license.dart
```diff
@@ -4,6 +4,10 @@
 
 part of 'license_detector.dart';
 
+/// Regular expression to match the common pattern of `Copyright (c) yyyy[-yyyy] <name>.`
+final _copyrightRegExp =
+    RegExp(r'^copyright.+\d{4}.{0,100}$', caseSensitive: false);
```
jonasfj, line 9:
```dart
    RegExp(r'^copyright.{0,5}\d{4}.{0,100}$', caseSensitive: false);
```
isn't `.+` a bit risky? how about limiting it to 5 characters

or maybe:
```dart
    RegExp(r'^copyright\s+(?:(c)|©)\s+\d{4}.{0,100}$', caseSensitive: false);
```

jonasfj (approved):

lib/src/license_detection/license.dart
isoos: Updated with the later, using a `\(c\)` to fix parens.

PR #1109, opened 2022-08-05
Expand broadNative runtime with nativeJit.

### LINKED ISSUE #1028: Include _onNativeJit in broadNative?
Include `_onNativeJit` in `broadNative`?
https://github.com/dart-lang/pana/blob/master/lib/src/tag/_specs.dart#L103-L108

Otherwise, we won't detect platforms tags for anything importing things like `dart:cli`, see:
https://pub.dev/packages/winmd/score

I think `_onNativeJit` shouldn't be part of `broadFlutter`, so we won't detect `sdk:flutter` when `dart:mirrors` is used.

isoos:
Fixes #1028.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.21.16
 
 - Explain `HEAD` request for URL checks.
+- Expand `broadNative` runtime with `nativeJit`.
 
 ## 0.21.15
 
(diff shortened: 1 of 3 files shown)


sigurdm (approved):
LGTM

sigurdm:
One can say it is surprising a package can get platform:android when it uses dart:mirrors. I don't think there is a way do deploy on android without flutter (where mirrors are not allowed)... but I'm not sure how important that is.

isoos:
> One can say it is surprising a package can get platform:android when it uses dart:mirrors. I don't think there is a way do deploy on android without flutter (where mirrors are not allowed)... but I'm not sure how important that is.

I also think that this is strange, and I think it is a sign that we should do this differently.

isoos:
Closing as this seems to get us unintended side effects.

PR #1127, opened 2022-09-14
Prepare release of pana

jonasfj:


isoos (approved):
also need to update version (pub run build_runner build)

PR #1139, opened 2022-10-14
Fix how download URLs are calculated.

### LINKED ISSUE #1135: downloadPackage uses deprecated API, and makes false assumptions about PUB_HOSTED_URL
Fetching the list of versions and reading `archive_url` is the non-deprecated way to fetch packages, see:
https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md

https://github.com/dart-lang/pana/blob/2cf9032869666276b1d4100bef16a16aa85a4e40/lib/src/download_utils.dart#L34-L36

Also it's in general not possible to replace the path from `PUB_HOSTED_URL`.

You must always append to the URL, it is a URL, not a host. The base-path is a legitimate part of the URL.

Example: `PUB_HOSTED_URL=https://pub.example.com/user-repos/foo/` then version-listing api is `https://pub.example.com/user-repos/foo/api/packages/<package>`. Hence, it's always wrong to use `.replace(path: ...`

isoos:
Fixes #1135.

### DIFF
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -22,33 +22,47 @@ Future<void> downloadPackage(
   required String destination,
   String? pubHostedUrl,
 }) async {
-  // Find URI for the package tar-ball
-  final pubHostedUri = Uri.parse(pubHostedUrl ?? 'https://pub.dartlang.org');
-  if (version == null) {
-    final versionsUri = pubHostedUri.replace(path: '/api/packages/$package');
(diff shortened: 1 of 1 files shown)


jonasfj (approved):
Great, you still need to wrap this with retries, if you want reliable downloads :D

isoos:
> Great, you still need to wrap this with retries, if you want reliable downloads :D

It is using RetryClient on line 27. Do we need different retries?

Issue #1134, opened 2022-09-26
Explain that "Unknown Platforms" is because static analysis of package failed.


ashitaprasad:
Various packages like [awesome_select](https://pub.dev/packages/awesome_select/score), [progress_indicator_button](https://pub.dev/packages/progress_indicator_button/score) and others have been tagged as "Unknown Platforms" although the Platform support section shows no error (Supports 6 of 6 possible platforms).

A thing to note is that both these packages failed to pass the static analysis. Is this the cause of this issue?

isoos:
The linked packages has serious failures in their static analysis, and in such cases we don't assign platform tags to the package.

isoos:
We've since updated the label that indicates static analysis issue.

PR #1144, opened 2022-10-19
'has:errors' tag to indicate analysis failures.

isoos:
#1134

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -104,6 +104,9 @@ class PackageAnalyzer {
     InspectOptions options, {
     Future<void> Function(String filename, Uint8List data)? storeResource,
   }) async {
+    const hasToolFailureTag = 'has:tool-failure';
+    const hasAnalysisFailureTag = 'has:analysis-failure';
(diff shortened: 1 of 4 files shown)


lib/src/package_analyzer.dart
```diff
@@ -104,6 +104,9 @@ class PackageAnalyzer {
     InspectOptions options, {
     Future<void> Function(String filename, Uint8List data)? storeResource,
   }) async {
+    const hasToolFailureTag = 'has:tool-failure';
+    const hasAnalysisFailureTag = 'has:analysis-failure';
```
jonasfj, line 108: Aren't these the same to end-users?

Can we just expose them as one tag?

isoos: For the end user we could present both as "Platform classification failed because `pana` encountered issues during package analysis." However, for us, it has a small value to separate the two, as `has:tool-failures` can be used to check failures that we are not aware of, and should be fixed.

isoos: Done.

jonasfj (approved):

Issue #1092, opened 2022-07-20
dart:mirrors shouldn't disqualify platform:linux


isoos:
From `package:pigeon`:

<img width="334" alt="image" src="https://user-images.githubusercontent.com/4778111/179915130-2ec77f99-f7e1-4943-9a32-42449900e2bb.png">

mraleph:
It would be nice to get this fixed. Currently a package like https://pub.dev/packages/steward has no platform compatibility for no good reason.

sigurdm:
This would be a good candidate for: https://github.com/dart-lang/pub-dev/issues/6400

sigurdm:
We believe this is fixed!

PR #1146, opened 2022-10-21
Fix platform tagging with dart:mirrors.

isoos:
- Fixes the Dart SDK part of #1092.
- Expanded `Platform` specification with `allowsNativeJit`, which will allow additional Dart libraries to be used when the package depends only on the Dart SDK.
- Note: inferring the SDK status relies on the current implementation of the `tags` and `explanations` passed in, and run after the sdk tagging was called. A follow-up refactoring + cleanup is needed to better handle this status, and until that I've added `package:steward` to the end2end tests, so we don't break this feature.


lib/src/tag/_specs.dart
```diff
@@ -160,11 +166,13 @@ class Platform {
     'Linux',
     Runtime.broadNative,
     tag: 'platform:linux',
+    allowsNativeJit: true,
```
jonasfj, line 169: Unclear why we need this hack?

Maybe, `broadNative` should include `nativeJit`?

Or is that going to get the wrong detection in some cases?

I suppose cases that works on web and native dart, would get detected as supported as working under flutter web? and there for get `sdk:flutter` and `sdk:dart` with `platform:linux` even if it uses mirrors on linux (which won't work with flutter).

I'm not sure I see a mechanism to solve that. Maybe, we need to detect sdk and then use that to detect platforms... or use the set of supported platforms to limit the set of supported sdks, saying that if it can't have an SDK tag if it doesn't satisfy all detect platforms for the sdk.

jonasfj (changes requested):

jonasfj (approved):

PR #1149, opened 2022-10-27
Tag packages that have screenshots

jonasfj:


lib/src/package_analyzer.dart
```diff
@@ -221,6 +221,9 @@ class PackageAnalyzer {
         }
       }
     }
+    if (processedScreenshots.isNotEmpty) {
+      tags.add('has:screenshot');
```
jonasfj, line 225: @isoos should we be plural or singular?

szakarias (commented):
LGTM

szakarias (approved):

PR #1152, opened 2022-10-28
New platform tagging with separate Dart and Flutter runtimes.

isoos:
The top-level tagging methods with `List<String> tags, List<Explanation> explanations` could be moved to use `TaggingResult` as the next step.

### DIFF
--- a/lib/src/tag/_common.dart
+++ b/lib/src/tag/_common.dart
@@ -16,6 +16,26 @@ typedef Explainer<N> = Explanation? Function(List<N> path);
 /// Explainer should give an explanation of the problem.
 typedef Predicate<T> = Explainer<T>? Function(T node);
 
+/// The accumulated values of a tagging step.
+class TaggingResult {
+  final Set<String> tags;
+  final List<Explanation> explanations;
+
+  TaggingResult(this.tags, this.explanations);
+
+  factory TaggingResult.merge(Iterable<TaggingResult> results) {
+    final tags = <String>{};
+    final explanations = <Explanation>[];
+    for (final result in results) {
+      tags.addAll(result.tags);
+      explanations.addAll(result.explanations);
+    }
+    return TaggingResult(tags, explanations);
+  }
+
+  bool get isSuccess => tags.isNotEmpty && explanations.isEmpty;
+}
+
 /// Indicates an issue.
 class Explanation {
   final String finding;
--- a/lib/src/tag/_specs.dart
+++ b/lib/src/tag/_specs.dart
@@ -121,15 +121,15 @@ class Runtime {
 /// A platform where Dart and Flutter can be deployed.
 class Platform {
   final String name;
-  final Runtime runtime;
+  final Runtime dartRuntime;
+  final Runtime flutterRuntime;
   final String tag;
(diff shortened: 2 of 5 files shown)


lib/src/tag/_specs.dart
```diff
@@ -154,36 +154,39 @@ class Platform {
...
     'Android',
-    Runtime.broadNative,
+    dartRuntime: Runtime.nativeAot,
+    flutterRuntime: Runtime.flutterNative,
     tag: 'platform:android',
   );
   static final ios = Platform(
     'iOS',
-    Runtime.broadNative,
+    dartRuntime: Runtime.nativeAot,
```
jonasfj, line 163: There is no `dartRuntime` on this platform, so what does this mean?

lib/src/tag/tagger.dart
```diff
@@ -221,22 +221,24 @@ class Tagger {
...
+      explanations.addAll(Platform.binaryOnlyNotAssignedPlatforms.map((p) =>
+          Explanation(p.name,
+              'Cannot assign ${p.name} automatically to a binary only package.',
+              tag: p.tag)));
+      return;
+    }
+
+    final platformResults = <TaggingResult>[];
+    for (final platform in Platform.recognizedPlatforms) {
+      TaggingResult checkRuntime(Runtime runtime) {
```
jonasfj, line 236: Could we move this `checkRuntime` to be a method next to `platformTags`?

lib/src/tag/_common.dart
```diff
@@ -16,6 +16,26 @@ typedef Explainer<N> = Explanation? Function(List<N> path);
...
+  final List<Explanation> explanations;
+
+  TaggingResult(this.tags, this.explanations);
+
+  factory TaggingResult.merge(Iterable<TaggingResult> results) {
+    final tags = <String>{};
+    final explanations = <Explanation>[];
+    for (final result in results) {
+      tags.addAll(result.tags);
+      explanations.addAll(result.explanations);
```
jonasfj, line 31: Won't this give conflicting tags and explanations?

Like an explanation saying "Doesn't support linux because", while it has the tag `platform:linux`

lib/src/tag/_specs.dart
isoos: Hm, good point, maybe we make it nullable and leave it out here?

lib/src/tag/_common.dart
isoos: The outer loop iterates over the platforms, and for each of them, we put the dart and flutter runtime tagging result into a list (if they are available, see tagger.dart#307). If any of them is a success, we award the tag, and don't expose any explanation (if the other runtime would fail). Otherwise we return the first result (which like has a failure explanation).

lib/src/tag/tagger.dart
isoos: Done.

lib/src/tag/_common.dart
isoos: Moved the inner method to class-level, it makes this logic more apparent, PTAL.

jonasfj (approved):

Issue #1159, opened 2022-11-14
pubspec.yaml with homepage: to github.com should show "Repository (GitHub)" instead of "Homepage" link


eseidel:
e.g. https://pub.dev/packages/process_run has homepage: github.com..., so it shows:

```
[Homepage](https://github.com/tekartik/process_run.dart)

Process run helpers for Linux/Win/Mac and which like feature for finding executables.

Metadata
```

Where as:
https://pub.dev/packages/analyzer
(correctly) uses the newer "repository: github.com..." and thus shows:

```
Metadata
This package provides a library that performs static analysis of Dart code.

[Repository (GitHub)](https://github.com/dart-lang/sdk/tree/main/pkg/analyzer)
[View/report issues](https://github.com/dart-lang/sdk/issues)
```

isoos:
This non-verification patch has been deployed a few weeks ago, and the repository verification is planned to roll out in early January. Closing this as we don't have much else to do with the issue.

PR #1160, opened 2022-11-16
Lenient repository/homepage URL switch for packages that have no repository verification enabled.

isoos:
joFixes #1159.

### DIFF
--- a/lib/src/references/pubspec_urls.dart
+++ b/lib/src/references/pubspec_urls.dart
@@ -7,6 +7,7 @@ import 'package:path/path.dart' as p;
 import '../model.dart';
 import '../package_context.dart';
 import '../report/_common.dart';
+import '../repository/repository_url_parser.dart';
 
 class PubspecUrlsWithIssues {
   final UrlWithIssue homepage;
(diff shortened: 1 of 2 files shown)


lib/src/references/pubspec_urls.dart
```diff
@@ -58,12 +59,24 @@ Future<PubspecUrlsWithIssues> checkPubspecUrls(PackageContext context) async {
...
-      pubspec.repository == null &&
-      isVerifiedRepository) {
-    var r = repository;
-    repository = homepage;
-    homepage = r;
+  if (pubspec.homepage != null && pubspec.repository == null) {
+    // We may switch these values if the repository has been verified, or
+    // if the verification was not enabled, but the URL parsing recognizes
+    // it as a valid repository.
+    var maySwitch = isVerifiedRepository;
```
sigurdm, line 66: Is switching not the wrong term here? Could we call this something like
```suggestion
    var canUseHomepageAsRepository = isVerifiedRepository;
```

sigurdm (approved):
LGTM

should we have a test covering this logic?

isoos:
> should we have a test covering this logic?

In the long term this will be covered with the repository verification enabled for all packages. I can add a test or two, but if this breaks, no big deal for now.

sigurdm:
> In the long term this will be covered with the repository verification enabled for all packages. I can add a test or two, but if this breaks, no big deal for now.

Yeah I guess it is not worth the time now - just was surprised that this broke in the first place.

Let's just ship it.

jonasfj (approved):

PR #1163, opened 2022-11-23
Summary.createdAt + extending the change method.

### LINKED ISSUE #1162: pana summary should include createdAt
PanaReport in pub-dev has a timestamp.

We use this to determine how old a pana report is.

But there is really no reason we can't just store this in summary. Whenever, pana  is running  can probably trust it to report the correct date-time of when it started running as a `createdAt`.

isoos:
Fixes #1162.

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -138,6 +138,7 @@ class PackageAnalyzer {
       log.info('Unable to read pubspec.yaml', e, st);
       tags.add(PanaTags.hasError);
       return Summary(
+        createdAt: DateTime.now().toUtc(),
         runtimeInfo: _toolEnv.runtimeInfo,
         packageName: null,
         packageVersion: null,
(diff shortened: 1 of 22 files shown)


test/goldens/end2end/http-0.13.0.json
```diff
@@ -1,4 +1,5 @@
 {
+  "createdAt": "2022-11-23T11:09:00.000",
```
jonasfj, line 2: Lets make it UTC

lib/src/package_analyzer.dart
```diff
@@ -255,6 +256,7 @@ class PackageAnalyzer {
       errorMessage = context.errors.map((e) => e.trim()).join('\n\n');
     }
     return Summary(
+      createdAt: DateTime.now(),
```
jonasfj, line 259:
```suggestion
      createdAt: DateTime.now().toUtc(),
```

otherwise it won't end with `Z` when render toJson.

test/end2end_test.dart
```diff
@@ -67,11 +67,13 @@ void main() {
         final flutterDartVersion =
             summary.runtimeInfo.flutterInternalDartSdkVersion;
         summary = summary.change(
-            runtimeInfo: PanaRuntimeInfo(
-          panaVersion: '{{pana-version}}',
-          sdkVersion: '{{sdk-version}}',
-          flutterVersions: {},
-        ));
+          createdAt: DateTime(2022, 11, 23, 11, 09),
```
jonasfj, line 70: Consider using `package:clock` instead

jonasfj (approved):

PR #1166, opened 2022-12-01
Isolate git process from host config through environment variables.

isoos:


lib/src/repository/git_local_repository.dart
```diff
@@ -64,6 +78,11 @@ class GitLocalRepository {
       environment: {
         'LANG': 'C', // default English locale that is always present
         'LC_ALL': 'en_US',
+        // prevent git from reading host configuration files
+        'HOME': homePath,
+        'GIT_CONFIG': p.join(homePath, '.gitconfig'),
+        'GIT_CONFIG_GLOBAL': p.join(homePath, '.global-gitconfig'),
+        'GIT_CONFIG_NOSYSTEM': '1',
```
jonasfj, line 85: Isn't the combination of these overkill?

lib/src/repository/git_local_repository.dart
```diff
@@ -64,6 +78,11 @@ class GitLocalRepository {
       environment: {
         'LANG': 'C', // default English locale that is always present
         'LC_ALL': 'en_US',
+        // prevent git from reading host configuration files
+        'HOME': homePath,
```
jonasfj, line 82: Lets set `GIT_TERMINAL_PROMPT: '0'`

See: https://git-scm.com/docs/git

That's stop it from prompting for anything.

lib/src/repository/git_local_repository.dart
jonasfj: According to: https://git-scm.com/docs/git
You can just set `GIT_CONFIG_GLOBAL` to `/dev/null`.

It might be better to use an empty temporary file for windows compatibility.

But setting `GIT_CONFIG` should be overkill, if you quote:
> GIT_CONFIG_SYSTEM
 Take the configuration from the given files instead from global or system-level configuration files. If GIT_CONFIG_SYSTEM is set, the system config file defined at build time (usually /etc/gitconfig) will not be read. Likewise, if GIT_CONFIG_GLOBAL is set, neither $HOME/.gitconfig nor $XDG_CONFIG_HOME/git/config will be read. Can be set to /dev/null to skip reading configuration files of the respective level.

See https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGSYSTEMcode

isoos: > Isn't the combination of these overkill?

I don't see much drawback doing it with the strict values. Is there?

jonasfj (approved):

Issue #1170, opened 2022-12-08
Pana should use `dart pub` to download and extract packages. 


sigurdm:

sigurdm:
Maybe use `pub unpack` when available.

sigurdm:
This is still relevant

Issue #1165, opened 2022-11-28
ListMixin.firstWhere at .downloadPackage 


jonasfj:
```
Error: Bad state: No element

at .ListMixin.firstWhere
at .downloadPackage ( package:pana/src/download_utils.dart:48 )
at StackZoneSpecification._registerUnaryCallback.<fn> ( package:stack_trace/src/stack_zone_specification.dart:125 )
```

fraca98:
Hi sorry for the question. I published a package on pub.dev and from the Pana analysis i get this error. Is there anything I can do to fix it?
```
0/10 points: Failed to run dartdoc

    Running dartdoc failed with the following error: Bad state: No element

#0      ListMixin.firstWhere (dart:collection/list.dart:167:5)
#1      downloadPackage (package:pana/src/download_utils.dart:48:10)
<asynchronous suspension>
#2      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:124:15)
<asynchronous suspension>
```
I'm a little bit confused cause it never happened to me before 😢

isoos:
@fraca98: what was the package/version?

fraca98:
> @fraca98: what was the package/version?

This is the package https://pub.dev/packages/withings_flutter and the issue is related to the last version (1.0.3). You can check the error in the section 'score'

isoos:
Based on the pana version, and the observable data, this seems to be a cache issue in combination with partial traffic migration of a new release we are rolling out (as I see only the "current" site version with earlier pana). The new release picked up the newly published version earlier than its version list cache expired, and failed to analyze it in the first attempt.

@fraca98: you don't need to do anything, there will be a repeated attempt to analyze the package and it will pick up the correct version.

@jonasfj: we shall extend the download util to use explicit `downloadUrl` when it is available, it seems to be easier to do than figuring out cache invalidation across runtime versions.

PR #1174, opened 2022-12-09
Use archiveUrl when specified.

isoos:
- #1165
- I also think we should not implement #1170, as it also relies on the potentially cached version list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Fix: tar extraction issue with directory entries.
 - Fix: revert the extra line in <summary> blocks.
+- Use 'archiveUrl' when specified.
 
 ## 0.21.22
 
--- a/lib/src/download_utils.dart
+++ b/lib/src/download_utils.dart
@@ -21,6 +21,7 @@ Future<void> downloadPackage(
   String? version, {
   required String destination,
   String? pubHostedUrl,
+  String? archiveUrl,
 }) async {
   pubHostedUrl ??= 'https://pub.dartlang.org';
   final pubHostedUri = Uri.parse(pubHostedUrl);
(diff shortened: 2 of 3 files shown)


sigurdm:
> I also think we should not implement Pana should use dart pub to download and extract packages.  #1170, as it also relies on the potentially cached version list.

Not sure this is true if I read it right. pub doesn't rely on the cached version listings - only as a heuristic to speed up requesting version listings for transitive dependencies.

sigurdm (approved):
LGTM

sigurdm:
> Not sure this is true if I read it right. pub doesn't rely on the cached version listings - only as a heuristic to speed up requesting version listings for transitive dependencies.

Ah you were talking about the cached version listings on pub.dev - just forget what I wrote then.

isoos:
> @isoos I don't understand this concern. The dart pub client will rely on the same APIs as the downloadPackage method uses.

Exactly, and that is causing analysis failures right now. This PR allows us to feed the `archiveUrl` to pana without going through the cached API response.

The root cause of this issue is the cache: we receive a new version, but the cache may be stale for 10 minutes (either because we have split traffic or failed to purge it), during which any analysis that uses the API will fail for the new version.

sigurdm:
Could we make some retry-logic in pana (or perhaps pub-worker) to take these possible 10 minutes into account?

* pub get
  - (wait 1 minute)
* pub get
  - (wait 2 minutes)
* pub get
  - (wait 3 minutes)
* pub get
  - (wait 4 minutes)
* pub get
  - give up.

...or something even more clever

isoos:
I don't think `pub get` should have any minute+ retry. A few seconds may be okay for transient issues, but the process shouldn't hang for minutes just because there may be a cache invalidation down the line...

Coming from that angle, maybe this change is not needed either, instead, we should have a better retry logic on pub.dev, e.g. invalidating the cache if the first attempt of a latest version fails. (And do the second attempt after the cache has been invalidated.)

isoos:
Closing this as we will land the cache purge instead.

PR #1183, opened 2022-12-23
Allow dart 2 compatible packages to be future compatible

sigurdm:
Fix of https://github.com/dart-lang/pub-dev/issues/6305


isoos (approved):

isoos:
I think we should also expose the dart3Incompatible field as well, may be in a different PR.

sigurdm:
> I think we should also expose the dart3Incompatible field as well, may be in a different PR.

Yeah - let us do that in a different PR.

PR #1184, opened 2022-12-29
Batch analyze licenses + compare changes.

isoos:


third_party/pub/batch_analyze_licenses.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:pana/src/license.dart';
+
+Future<void> main() async {
```
jonasfj, line 10: Write a comment that says what this tool does....
```suggestion
Future<void> main(List<String> args) async {

  if (args.length > 1 && args[1] == '--help') {
    print('''
Run license detection on all files matching `.dart_tool/pana/license-cache/LICENSE-<package>`.

Creates a file named: `detected-licenses.<date-time>.json`, on the form:
{
  "<package>": {
    "spdxIds": ["<detected SPDX ID>", ...]
  },
  ...
},
''');
  }
```

Supposing we change where it writes files...

third_party/pub/batch_compare_licenses.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+Future<void> main(List<String> args) async {
```
jonasfj, line 8: Can we add a help comment that says what it does and prints? or maybe a `--help` handler?

jonasfj:
Note: `third_party/...` is for things imported/vendored from a third-party source.

See: https://opensource.google/documentation/reference/thirdparty/non-google3

third_party/pub/README.md
```diff
@@ -0,0 +1,8 @@
+# pub.dev licenses
+The `licenses/` directory could contain the licenses of the latest versions of packages on `pub.dev`.
```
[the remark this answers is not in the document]
isoos: Yeah, true, if we don't want to commit e.g. a subset of license files into the repository, this makes more sense. I'll update it.

isoos:
> We're not embedding them in this repository, we're just temporarily storing them in some cache... they are not distributed as part of pana.

My initial plan was to also include it here as extended tests, but it may not provide much value after all...

.gitignore
```diff
@@ -6,3 +6,4 @@ pubspec.lock
 test/license_test_assets/license_scenarios
 .vscode
 third_party/spdx/licenses
+third_party/pub/licenses
```
jonasfj, line 9:
```suggestion
```

jonasfj (approved):

PR #1188, opened 2023-01-03
Create publishing workflow

sigurdm:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  publish:
+    permissions:
+      id-token: write # This is required for requesting the JWT
+    runs-on: ubuntu-latest
+    environment: pub-dev
+    steps:
+      - uses: actions/checkout@v3
+      - uses: isoos/setup-dart@oidc-token
```
sigurdm, line 16: Not sure I like this specific line :)

jonasfj: hehe, maybe it's time to merge it into upstream...

.github/workflows/publish.yml
```diff
@@ -0,0 +1,22 @@
+name: Publish to pub.dev
+
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+**'
```
jonasfj, line 6: Do we need two stars? is this actually a regex pattern?

It's close I guess:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Maybe, we just need `*` at the end...

jonasfj (approved):

.github/workflows/publish.yml
sigurdm: Yeah - one star is probably better

sigurdm, line 6:
```suggestion
      - '[0-9]+.[0-9]+.[0-9]+*'
```

PR #1192, opened 2023-01-18
Strict names in package layout checks.

isoos:
A matching change to https://github.com/dart-lang/pub-dev/pull/6345

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.21.25
+
+- Strict names in package layout checks.
+
 ## 0.21.24
 
 - Fix: allow <4.0.0 for Dart 3 packages.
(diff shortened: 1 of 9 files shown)


sigurdm (approved):

kuhnroyal:
The removal of the public `textFileNameCandidates` in a patch release is kinda breaking.
I have seen several tools using this and out of a sudden I have failing pipelines.
For example: https://github.com/rufman/license_checker/blob/trunk/lib/src/dependency_checker.dart#L31

Can we restore and deprecate this function in a new release, or retract the version?

isoos:
I'm sorry for breaking it. I suggest to copy-paste the old code from pana into the project, as it is unlikely that we will get that method back.

Background: Most of pana's non-entrypoint, non-model API is public only to get it reused in `pub.dev`, and we are not really tracking other uses, only preserving data model compatibility most of the time. In the near future is also expected that we will migrate pana to a package that has most of these helper methods removed from public API, as it will be invoked on command-line only. Having said that, if there are reasonable suggestions, we may expose internals in a supported way, but so far it never came up.

PR #1198, opened 2023-02-10
Fix future-compatibility check: run pub get in a separate copy.

isoos:
- Updated `ToolEnvironment` to also have `pub upgrade` support with future SDK (note: these become a bit more complex than we've initially planned for, a refactor is due)
-  Updated the compatibility check: copies the package into a temp directory, deletes `.dart_tool`, runs `pub upgrade` before running the analysis.


sigurdm (approved):
LGTM

PR #1201, opened 2023-03-01
Add tags for topics

szakarias:


lib/src/package_analyzer.dart
```diff
@@ -284,6 +284,13 @@ class PackageAnalyzer {
     if (processedScreenshots.isNotEmpty) {
       tags.add(PanaTags.hasScreenshot);
     }
+
+    if (pubspec.topics.isNotEmpty) {
```
isoos, line 288: This check seems to be not necessary.

szakarias: done

isoos (approved):
Please also add a changelog entry.

PR #1202, opened 2023-03-02
Make example-detection work on case-preserving file systems

sigurdm:
This makes the build green again.

We should probably list all the files once, and do the same for all our file checks (readme, changelog...)


isoos (approved):
Thanks!

PR #1213, opened 2023-04-19
New report section for Dart 3 compatiblity.

### LINKED ISSUE #1211: In scoring report, null-safety section should be replaced by dart-3-compatible.

isoos:
Fixes #1211.


sigurdm:
A bit of a shame we don't have any test here. I realize it might not be easy to set up on CI. Could we perhaps use the same SDK for current and future sdk for the sake of the tests?

jonasfj (approved):

jonasfj:
@sigurdm, yeah, test would be nice -- but the future sdk stuff will go away after Dart 3, right?

sigurdm:
Yeah I guess so...

PR #1226, opened 2023-05-12
Upgrade CI to Dart 3.0

isoos:
- Updated the minimum SDK requirement to 3.0, as we want to change some analysis features now that 3.0 is out.
- Also updated end2end version of `dnd` (a follow-up may also update other items in the list).
- Breaking change in clients that may use `PanaProcessResult` or `DartdocResult`.
- Updated dependency test to match the output.


test/goldens/end2end/__timestamp.txt
```diff
@@ -1 +1 @@
-2022-10-20T07:30:43.697198Z
+2023-05-12T12:55:25.249916Z
```
jonasfj, line 1: Should we move this to a separate PR?

isoos: We want these changes either in this PR or the very next one, because the old version set of the packages and their dependencies doesn't make much sense now with 3.0 SDK.

sigurdm (approved):

PR #1232, opened 2023-06-05
Do not block repository check when `pubspec.yaml` is in path that is not accepted.

### LINKED ISSUE #1200: Pana should not fail to verify repo URL if it detects a malformed `pubspec.yaml`
## Description

If you have a mono repository that contains multiple packages with pubspecs, pana is able to detect those packages correctly.

But if you have pubspec files in the repo that are used as templates for code generation, and those pubspecs contain mustache code than pana fails with the following error:
```markdown
### [x] 0/10 points: Provide a valid `pubspec.yaml`

<details>
<summary>
Failed to verify repository URL.
</summary>

Please provide a valid [`repository`](https://dart.dev/tools/pub/pubspec#repository) URL in `pubspec.yaml`, such that:

 * `repository` can be cloned,
 * a clone of the repository contains a `pubspec.yaml`, which:,
    * contains `name: fluttium_protocol`,
    * contains a `version` property, and,
    * does not contain a `publish_to` property.

Unable to parse `pubspec.yaml` from git repository. Invalid argument(s): Path "bricks/fluttium_action/__brick__/{{name.snakeCase()}}/pubspec.yaml" is not accepted.
</details>
```

The repo does have a correct pubpsec.yaml for the `fluttium_protocol` package but because Pana goes through other directories first, it finds a malformed pubspec and immediately stops looking further. I would have assumed it would continue to find a correct pubspec.yaml and if it never did, then it should complain about the reasons why it didn't find a correct pubspec.

## Steps to reproduce

1. `git clone https://github.com/wolfenrain/fluttium.git`
2. `cd fluttium/packages/fluttium_protocol`
3. `pana . --no-warning`

isoos:
- This is one option to fix #1200, by allowing it to continue when the path is not accepted.
- (Please review the alternative too: #1233 .)

### DIFF
--- a/lib/src/repository/check_repository.dart
+++ b/lib/src/repository/check_repository.dart
@@ -95,7 +95,7 @@ Future<VerifiedRepository?> checkRepository({
             path,
             maxOutputBytes: _maxPubspecBytes,
           );
-        } on ArgumentError catch (e) {
+        } on GitToolException catch (e) {
           log.info('Unable to read "$path".', e);
           return _PubspecMatch(path, false, 'Unable to read `$path`: $e.');
         }
--- a/lib/src/repository/check_repository.dart
+++ b/lib/src/repository/check_repository.dart
@@ -87,12 +87,18 @@ Future<VerifiedRepository?> checkRepository({
       }
 
       Future<_PubspecMatch> matchRepoPubspecYaml(String path) async {
+        late String content;
         // checkout the pubspec.yaml at the assumed location
-        final content = await repo.showStringContent(
-          branch!,
-          path,
-          maxOutputBytes: _maxPubspecBytes,
-        );
+        try {
+          content = await repo.showStringContent(
+            branch!,
+            path,
+            maxOutputBytes: _maxPubspecBytes,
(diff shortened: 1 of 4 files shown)


lib/src/repository/check_repository.dart
```diff
@@ -87,12 +87,18 @@ Future<VerifiedRepository?> checkRepository({
...
-          path,
-          maxOutputBytes: _maxPubspecBytes,
-        );
+        try {
+          content = await repo.showStringContent(
+            branch!,
+            path,
+            maxOutputBytes: _maxPubspecBytes,
+          );
+        } on ArgumentError catch (e) {
```
sigurdm, line 98: Catching an (Argument) Error is almost always wrong.

Seems to me there is no good way for the caller to verify the path up front, so passing a bad path should not be an error.

Maybe `repo.showStringContent` should be throwing an `Exception` instead in that case or maybe return null to signify a failed attempt?

isoos: Done (using GitToolException).

PR #1233, opened 2023-06-05
Allow more characters in repository path segments.

### LINKED ISSUE #1200: Pana should not fail to verify repo URL if it detects a malformed `pubspec.yaml`

isoos:
- This is one option to fix #1200, by allowing more characters to be part of the path segments.
- (Please review the alternative too: #1232.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.21.33
 
+- Allow more characters in repository path segments.
 - Upgraded to Dart 3.0.
 
 **Breaking changes**:
--- a/lib/src/repository/git_local_repository.dart
+++ b/lib/src/repository/git_local_repository.dart
@@ -11,7 +11,7 @@ import '../utils.dart' show runProc, PanaProcessResult;
 
 final _acceptedBranchNameRegExp = RegExp(r'^[a-z0-9]+$');
 final _acceptedPathSegmentsRegExp =
-    RegExp(r'^[a-z0-9_\-\.]+$', caseSensitive: false);
+    RegExp(r'^[a-z0-9_\-\.\{\}\(\)]+$', caseSensitive: false);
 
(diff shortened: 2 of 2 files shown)


sigurdm:
Let's have both?

isoos:
> Let's have both?

I'm not convinced that we should have either, because one could argue that the check should be very strict. But in this case it feels like we should allow the check to go through in some way or another...

isoos:
Closing this in favor of #1232. Let's keep it restricted for now.

PR #1253, opened 2023-09-21
Fix failure message when the repository verification is unable to parse the URL from the file in the remote repository.

### LINKED ISSUE #1252: repository tag in pubspec.yaml incorrectly interpreted by pub.dev
pubspec.yaml had a correctly set `repository` tag, but pub.dev says:
 >pubspec.yaml from the repository URL missmatch: expected https://github.com/Alexqwesa/locale_switcher but got null.

[https://pub.dev/packages/locale_switcher](https://pub.dev/packages/locale_switcher)

pubspec.yaml contains:
```yaml
name: locale_switcher
description: A widget for switching the locale of your application with 2 lines of code.
repository: https://github.com/Alexqwesa/locale_switcher
homepage: https://github.com/Alexqwesa/locale_switcher
# ....
```

I also tried to only set `repository` or only set `homepage` - didn't work,

Note: when i first time published package, i made a mistake and set repository wrong...

isoos:
Fixes #1252.


lib/src/repository/check_repository.dart
```diff
@@ -143,9 +143,16 @@ Future<VerifiedRepository?> checkRepository({
               '`$path` from the repository has no `repository` or `homepage` URL.');
         }
         final gitRepoUrl = Repository.tryParseUrl(gitRepoOrHomepage);
-        if (gitRepoUrl?.cloneUrl != parsedSourceUrl.cloneUrl) {
+        if (gitRepoUrl == null) {
+          return _PubspecMatch(
+              path,
+              true,
```
sigurdm, line 149: Could we make this a named (but required argument) that is almost always nice for boolean parameters

lib/src/repository/check_repository.dart
```diff
@@ -143,9 +143,16 @@ Future<VerifiedRepository?> checkRepository({
...
+              '`$path` from the repository has a `repository` or `homepage` field '
+              '"`$gitRepoOrHomepage`" but cannot be parsed as a repository URL.');
+        }
+        if (gitRepoUrl.cloneUrl != parsedSourceUrl.cloneUrl) {
           return _PubspecMatch(path, true,
-              '`$path` from the repository URL missmatch: expected `${parsedSourceUrl.cloneUrl}` but got `${gitRepoUrl?.cloneUrl}`.');
+              '`$path` from the repository URL missmatch: expected `${parsedSourceUrl.cloneUrl}` but got `${gitRepoUrl.cloneUrl}`.');
```
sigurdm, line 155: Can we fix this also?:
```suggestion
              '`$path` from the repository URL mismatch: expected `${parsedSourceUrl.cloneUrl}` but got `${gitRepoUrl.cloneUrl}`.');
```

lib/src/repository/check_repository.dart
```diff
@@ -143,9 +143,16 @@ Future<VerifiedRepository?> checkRepository({
               '`$path` from the repository has no `repository` or `homepage` URL.');
         }
         final gitRepoUrl = Repository.tryParseUrl(gitRepoOrHomepage);
-        if (gitRepoUrl?.cloneUrl != parsedSourceUrl.cloneUrl) {
+        if (gitRepoUrl == null) {
+          return _PubspecMatch(
+              path,
+              true,
+              '`$path` from the repository has a `repository` or `homepage` field '
+              '"`$gitRepoOrHomepage`" but cannot be parsed as a repository URL.');
```
sigurdm, line 151: We should be explicit about "how" it cannot be parsed. Otherwise I would find this message confusing

lib/src/repository/repository_url_parser.dart
```diff
@@ -40,18 +43,19 @@ Repository? tryParseRepositoryUrl(String input) {
...
+        'URL path is not normalized: `${uri.path}` != `${normalizedUri.path}`');
   }
   // detect repo vs path segments
   final segments =
       normalizedUri.pathSegments.where((s) => s.isNotEmpty).toList();
   final repoSegmentIndex = _repoSegmentIndex(provider, segments);
-  if (repoSegmentIndex < 0) return null;
+  if (repoSegmentIndex < 0) {
+    throw FormatException(
+        'Could not find repository segment separator in `${normalizedUri.path}`.');
```
sigurdm, line 57: I somehow still think this message is hard to grog. I wouldn't know what a "repository segment separator" is.

We should probably document how we expect the repository link be formattet.

Maybe we should link to https://dart.dev/tools/pub/pubspec#repository and extend that with an explanation of how to deep link into the repo.

sigurdm (approved):

lib/src/repository/repository_url_parser.dart
isoos: I've updated with the message to report the issue at `pana`'s issue tracker. I think ultimately we want to have unrecognized examples reported, and keep the officially supported description vague. At this point in the parsing it will probably not fail at github deep links, but may fail on niche git repository structures like self-hosted git frontends.

PR #1262, opened 2023-10-02
Fix: static analysis is done only when the package resolution was successful.

isoos:
Fixes issues like the following score page:
image


sigurdm (approved):
LGTM

PR #1267, opened 2023-10-18
Upgrade lints to 3.0.0

isoos:
Upgrading lints affects both pana (issues fixed) and the analysis of packages.


lib/src/tag/tagger.dart
```diff
@@ -70,6 +70,7 @@
 /// * platform:windows
 ///
 /// A package has the same platform tags as the primary library.
+library tagger;
```
parlough, line 73: Consider dropping the library names (here and elsewhere):

```suggestion
library;
```

lib/src/tag/_violations.dart
```diff
@@ -81,6 +81,8 @@
 /// If both "platforms" and "flutter.platforms" are declared "platforms" takes
 /// precedence. (There might be rare cases where you want to declare a more
 /// narrow set of platforms than you have plugin support for).
+library _violations;
```
parlough, line 84:
```suggestion
library;
```

parlough (commented):
Exciting!

szakarias (approved):
LGTM

Issue #1229, opened 2023-06-01
[meta] pana standalone should be the same as pub.dev output


isoos:
- the customizations will be either removed from pub.dev or moved into pana
- `dartdoc` needs to have the the same output artifacts as `pub_dartdoc` on `pub.dev` or we should get them in a different way
- `pub.dev`-provided inferred data should be integrated in pana

sigurdm:
We are done here!

Issue #979, opened 2021-10-14
Indicate which elements are not documented in pana report


jonasfj:
Indicate which elements are not documented in pana report:
```
10/10 points: 20% or more of the public API has dartdoc comments
   * 65 out of 66 API elements (98.5 %) have documentation comments.
```
Couldn't we do something, maybe `<details><summary>` so that the user can see the first 3-5 elements without documentation comments? Maybe we only need to do this when we give zero points.

(I don't recall if this section is computed in pub-dev, there is a small risk it is)

Issue #837, opened 2021-02-22
Pana shows reduced score while running locally


dharanidharandharmasivam:
While running the pana locally for [our package](https://pub.dev/packages/syncfusion_flutter_charts), the score got reduced for not supporting multiple platforms.  I have attached the output log below by running pana locally.

<details>
  <summary>Output log of pana for our package</summary>

```

## ✓ Follow Dart file conventions (20 / 20)
### [*] 10/10 points: Provide a valid `pubspec.yaml`


### [*] 5/5 points: Provide a valid `README.md`


### [*] 5/5 points: Provide a valid `CHANGELOG.md`


## ✓ Provide documentation (10 / 10)
### [*] 10/10 points: Package has an example

* Found example at: `example/lib/main.dart`

## ✗ Support multiple platforms (0 / 20)
### [x] 0/20 points: Supports 0 of 3 possible platforms (iOS, Android, Web)

Found 8 issues. Showing the first 2:

Consider supporting multiple platforms:

<details>
<summary>
Package not compatible with runtime flutter-native on Android
</summary>

Because:
* `package:syncfusion_flutter_charts/src%5Ccommon%5Chandcursor%5Cweb.dart` that imports:
* `dart:html`
</details>

## ✓ Pass static analysis (30 / 30)
### [*] 30/30 points: code has no errors, warnings, lints, or formatting issues


## ✓ Support up-to-date dependencies (20 / 20)
### [*] 10/10 points: All of the package dependencies are supported in the latest version

|Package|Constraint|Compatible|Latest|
|:-|:-|:-|:-|
|[`flutter`]|`flutter`|0.0.0|0.0.0|
|[`intl`]|`>=0.15.0 <0.20.0`|0.16.1|0.17.0|
|[`syncfusion_flutter_core`]|`^18.4.43`|18.4.43|18.4.43|
|[`vector_math`]|`^2.0.8`|2.1.0-nullsafety.3|2.1.0|

<details><summary>Transitive dependencies</summary>

|Package|Constraint|Compatible|Latest|
|:-|:-|:-|:-|
|[`characters`]|-|1.1.0-nullsafety.3|1.1.0|
|[`collection`]|-|1.15.0-nullsafety.3|1.15.0|
|[`meta`]|-|1.3.0-nullsafety.3|1.3.0|
|[`path`]|-|1.7.0|1.8.0|
|[`sky_engine`]|-|0.0.99|0.0.99|
[...]

PR #1270, opened 2023-10-27
Run and process dartdoc output to create pub-data.json and the coverage report.

isoos:
- part of #1229, #837 and #979
- copied most files in `lib/src/dartdoc` from `pub-dev` (except `lib/src/dartdoc/dartdoc.dart`)
- enabled to run the analysis by default in `bin/pana.dart`, using the latest `dartdoc` through (a) custom `PUB_CACHE` + `pub global active dartdoc`
- enabled to run on a few selected packages in end2end tests (also updated end2end reference timestamp and output)


isoos:
PTAL. I've removed the retry loop, deprecated its option and refactored the package_context code as we have discussed. I've opted for another small breaking change with `DartdocResult`, as we haven't used that class for anything recently.

sigurdm (approved):
LGTM

PR #1282, opened 2023-11-14
Limit the number of exported documentation entries on pub-data.json.

isoos:
Copy of the code from https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_dartdoc/lib/pub_data_generator.dart#L91-L97


szakarias (approved):
LGTM

PR #1283, opened 2023-11-14
Pre-compute missing documentation symbol before pruning the results.

isoos:
- Fixes the edge case of #1282, where a large but overall under-documented package removes some documentation which may be reported as missing.
- It also enables the separation of entries data (part of a long-term refactoring of pub-data.json to better fit the indexing).

### DIFF
--- a/lib/src/dartdoc/pub_dartdoc_data.dart
+++ b/lib/src/dartdoc/pub_dartdoc_data.dart
@@ -64,9 +64,15 @@ class Coverage {
   /// The number of API elements with accepted documentation.
   final int documented;
 
+  /// Some of the API symbols that are without accepted documentation.
+  ///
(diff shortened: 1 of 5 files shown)


lib/src/dartdoc/pub_dartdoc_data.dart
```diff
@@ -64,9 +64,15 @@ class Coverage {
   /// The number of API elements with accepted documentation.
   final int documented;
 
+  /// Some of the API symbols that are without accepted documentation.
+  ///
+  /// To limit the output size, we only store the a subset of the missing symbols.
```
sigurdm, line 69:
```suggestion
  /// To limit the output size, we only store a subset of the missing symbols.
```

sigurdm (approved):

PR #1288, opened 2023-11-20
Test to detect new libraries added to the Dart SDK.

### LINKED ISSUE #1286: Add test to automatically check if we know about all SDK libraries
https://github.com/dart-lang/sdk/blob/main/sdk/lib/libraries.yaml

isoos:
- Fixes #1286.
- Maybe in the future we should parse and interpret the library hierarchy automatically, but I don't expect too many changes there that warrants such details.

### DIFF
--- a/test/tag/dart_sdk_libraries_test.dart
+++ b/test/tag/dart_sdk_libraries_test.dart
@@ -1,0 +1,45 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:http/http.dart';
+import 'package:test/test.dart';
+
+/// This tests loads the library specification of the Dart SDK and tests if
+/// all of them are mentioned in `lib/src/tag/_specs.dart`.
+void main() {
+  test('Dart SDK libraries exist in _specs.dart', () async {
+    final rs = await get(Uri.parse(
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
LGTM

test/tag/dart_sdk_libraries_test.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'package:http/http.dart';
+import 'package:test/test.dart';
+
+/// This tests loads the library specification of the Dart SDK and tests if
+/// all of them are mentioned in `lib/src/tag/_specs.dart`.
+void main() {
+  test('Dart SDK libraries exist in _specs.dart', () async {
+    final rs = await get(Uri.parse(
+        'https://raw.githubusercontent.com/dart-lang/sdk/main/sdk/lib/libraries.json'));
```
sigurdm, line 16: Actually you might be able to find this in the local sdk. 
```
p.join(p.dirname(p.dirname(Platform.resolvedExecutable)), 'lib', 'libraries.json')
```

isoos: I'm undecided what the better approach here would be: detecting it locally is faster for the test, but looking at the git file gives better early warning...

sigurdm: Yeah... maybe the early warning is good. Are we running CI with a fixed SDK version, or auto-updating?

isoos: Fixed version for stable and for preview tests too, at one point we were autoupdating for preview but IIRC that got broken somehow.

Issue #1273, opened 2023-10-31
Remove reliance on dartdoc kind names (dartdoc_index.json)


isoos:

isoos:
This has been implemented.

PR #1292, opened 2023-11-22
Deprecate the use of 'kind' in pub-data.json.

isoos:
- the last part of #1273 (before complete removal)
- uses the same logic as https://github.com/dart-lang/pana/blob/master/lib/src/dartdoc/dartdoc_index.dart#L99-L100


isoos:
kind ping

sigurdm (approved):

PR #1298, opened 2023-12-14
Don't report hidden constructors as undocumented API symbols.

### LINKED ISSUE #1297: pana can't recognize my enum's documentation
I have fully documented my project, but pana (and pub.dev) says there are undocumented parts.

https://github.com/frknkrc44/easy_utils/blob/main/lib/src/navigation/page_route_type.dart

I already documented this enum (you can see it) but the pana output (same as pub.dev):
```
72 out of 73 API elements (98.6 %) have documentation comments.

Some symbols that are missing documentation: `PageRouteType`.
```

isoos:
Fixes #1297.

### DIFF
--- a/lib/src/dartdoc/index_to_pubdata.dart
+++ b/lib/src/dartdoc/index_to_pubdata.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:collection/collection.dart';
+
 import 'dartdoc_index.dart';
 import 'pub_dartdoc_data.dart';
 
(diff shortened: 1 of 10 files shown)


isoos:
Note: I think this fix could be worth publishing with the last deployment of `pub.dev`.

frknkrc44 (changes requested):
You are getting 5 or less items from the missing document list output, but it may cause an error when calculating the total number. You can remove ".take(5)" to solve this issue.

lib/src/dartdoc/index_to_pubdata.dart
```diff
@@ -55,7 +74,7 @@ PubDartdocData dataFromDartdocIndex(DartdocIndex index) {
   return PubDartdocData(
     coverage: Coverage(
       documented: documented,
-      total: apiElements.length,
+      total: total,
       symbolsMissingDocumentation: symbolsMissingDocumentation,
```
frknkrc44, line 78: Then you can add ".take(5).toList()" to here.

lib/src/dartdoc/index_to_pubdata.dart
```diff
@@ -37,12 +39,29 @@ PubDartdocData dataFromDartdocIndex(DartdocIndex index) {
     ));
   }
 
-  final documented = apiElements.where((e) => e.documentation != null).length;
+  /// Some symbols are repeated, and one instance may have a documentation, while
+  /// the other misses it. E.g. an enum with a default constructor may have dartdoc
+  /// on the enum type, but won't have anything on the (likely absent-from-code)
+  /// constructor.
```
sigurdm, line 45: I'm not sure "symbols" are actually repeated, there **could** be dartdoc on the enum AND the constructor.

Given that the constructor will show up as a separate entity in the dartdoc, the question is if we should actually make this fix.

On the other hand it does not make much sense to document enum constructors separately...

isoos: > On the other hand it does not make much sense to document enum constructors separately...

On top of that: these constructors are not in the code, they are omitted, but present in the dartdoc output. I think it would be a bad user experience to force developers to write constructors + documentation for those which could have been omitted otherwise.

lib/src/dartdoc/index_to_pubdata.dart
isoos: good catch! fixed.

lib/src/dartdoc/index_to_pubdata.dart
isoos: Updated documentation to reflect this better.

frknkrc44 (approved):
LGTM

isoos:
Note: I also think we should link to the dartdoc pages that are missing the documentation, but that requires a bit more changes and I'd rather get this out sooner than block it on the full set of changes.

sigurdm (approved):
I still think there might be more to do here (should default constructors perhaps have a default doc-string?) but let's land this as an improvement.

PR #1301, opened 2023-12-18
Remove --no-warning CLI argument support.

isoos:


sigurdm:
Should we perhaps just keep this forever to avoid breaking people's CI?

isoos:
> Should we perhaps just keep this forever to avoid breaking people's CI?

I don't think `pana` should be like that forever. We will change internals, behavior and output over time (and I have a few of such changes planned right now), and based on how you depend on them, it will be breaking. (On the other hand, if people setup pana with versions constraints, they won't see a breaking change here).

sigurdm:
>  (On the other hand, if people setup pana with versions constraints, they won't see a breaking change here).

I suspect many do something along the lines of:

```
> dart pub global activate pana
> dart pub global run pana . --no-wait
```

Are your planned changes gonna break that flow? Otherwise I think we should leave it (And I think that is a reasonable flow that we should avoid breaking, even when doing a major version bump).

isoos:
Hm, good point. I wish we had a better path to trigger migration, because keeping code like this forever isn't ideal either. On the other hand, searching for `"pub global run pana" site:github.com` and `"dart run pana" site:github.com` has like less than 10 uses with `--no-warning`. Maybe if we migrate them, can this be removed?

isoos:
Thinking about this furtherer, maybe this is no that bad CLI option to keep. It is not too far fetched that the analysis may run something that warrants the warning message again (`dart:ffi` or any non-Dart toolchain for Android apps comes to mind). Keeping it for now.

PR #1302, opened 2023-12-20
Rename runProc -> runConstrained and move related classes into src/tool/run_constrained

isoos:
- This PR does not change the code.
- Further refactor/breaking changes will touch this code, and it become way too many things changing at once - I'm splitting up the PRs, and this is the first part.


sigurdm (approved):
LGTM, but lets also rename the function to `runProcess` (or something even better... `runWithTimeout`...) Doesn't have to be this PR, but maybe the file-name should be changed already now.

runProc doesn't really say much, and is an abbreviation.

isoos:
Renamed to `runWithTimeout` (also `run_with_timeout.dart`).

isoos:
after an offline discussion, renamed to `runConstrained`.

PR #1316, opened 2024-01-16
Remove some methods and method parameters from ToolEnvironment.

isoos:
- `'--no-validate-links'` in `dartdoc()` is always set anyway, no need to keep the optional parameter (similarly the other parameters are never set, we can remove them).
- removing further methods that are never called or restringing the visibility of them (`_getFlutterVersion()`).


sigurdm (approved):
Nice!

PR #1323, opened 2024-01-24
Dynamically reduce dartdoc timeout with total time budget option.

isoos:
Should address https://github.com/dart-lang/pub-dev/issues/7430

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -30,7 +30,7 @@ class InspectOptions {
   final String? pubHostedUrl;
 
   /// The output directory to copy the generated docs. When not specified,
-  /// the documents will be discarded.
+  /// the generated docs will be discarded.
   final String? dartdocOutputDir;
 
   /// The total time budget allocated for the full analysis. `pana` may not be
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -26,17 +26,33 @@ import 'tool/run_constrained.dart';
 import 'utils.dart';
 
 class InspectOptions {
+  /// The PUB_HOSTED_URL to use for the package download and dependency analysis.
   final String? pubHostedUrl;
+
+  /// The output directory to copy the generated docs. When not specified,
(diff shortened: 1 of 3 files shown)


lib/src/package_analyzer.dart
```diff
@@ -28,12 +28,14 @@ import 'utils.dart';
...
   final String? pubHostedUrl;
   final String? dartdocOutputDir;
+  final Duration? totalTimeBudget;
   final Duration? dartdocTimeout;
   final int? lineLength;
 
   InspectOptions({
     this.pubHostedUrl,
     this.dartdocOutputDir,
+    this.totalTimeBudget,
```
sigurdm, line 38: Could we document this a bit.
Can you pass both timeouts? How do they interact? Are there defaults?

isoos: Updated this and other options. PTAL

lib/src/package_analyzer.dart
```diff
@@ -26,15 +26,33 @@ import 'tool/run_constrained.dart';
 import 'utils.dart';
 
 class InspectOptions {
+  /// The PUB_HOSTED_URL to use for the package download and dependency analysis.
   final String? pubHostedUrl;
+
+  /// The output directory to copy the generated docs. When not specified,
+  /// the documents will be discarded.
```
sigurdm, line 33:
```suggestion
  /// the generated docs will be discarded.
```

lib/src/package_analyzer.dart
```diff
@@ -26,15 +26,33 @@ import 'tool/run_constrained.dart';
...
   final String? dartdocOutputDir;
-  final Duration? dartdocTimeout;
+
+  /// The total time budget allocated for the full analysis. `pana` may not be
+  /// able to finish the analysis withing this time, but some parts will be
```
sigurdm, line 37:
```suggestion
  /// able to finish the analysis within this time, but some parts will be
```

sigurdm (approved):

PR #1327, opened 2024-02-02
Additional CLI arguments for pana binary.

isoos:
- `[dart|flutter]-sdk` is extended `[dart|flutter]-config-home` to have a better `SdkConfig` object.
-  `[dartdoc|resources]-output` for the files written by `dartdoc` and `pana` (e.g. `webp` screenshots)
- `dartdoc-timeout` and `total-time-budget` to control how long `pana` tries to run (or abort processing).


isoos:
Note: I think `total-time-budget` may not be the best name, any suggestions?

sigurdm:
> Note: I think total-time-budget may not be the best name, any suggestions?

`total-timeout` ?

isoos:
> total-timeout ?

I'm good with that, although, right now we don't really enforce it yet, only for reducing the dartdoc-timeout if we would be otherwise over the total. (I think later we could add some checks to really consider it as a timeout).

sigurdm (approved):

PR #1329, opened 2024-02-06
Account for removal of `analysis_options_user.yaml` from Flutter

parlough:
This file no longer exists after https://github.com/flutter/flutter/pull/138261 and is generally no longer supported.

Also update nearby code with expanded comments and avoiding the non-null assertions.


isoos (commented):
Needs an update for the golden files:
- `rm test/goldens/end2end/*json*`
- `dart test test/end2end_test.dart`

isoos:
I think you may have deleted all the files and not just `*json*`, which needs a full re-run if you change SDKs (as the dependencies will change).

parlough:
Thanks for the help with the goldens! I think I was only considering my Dart version and needed to use a matching Flutter build.

lib/src/analysis_options.dart
```diff
@@ -8,92 +8,53 @@ import 'dart:isolate';
...
+/// The default analysis options configuration (in its raw yaml format).
+Future<String> getDefaultAnalysisOptionsYaml() async =>
+    await _getLintsCoreAnalysisOptions();
 
 Future<String> _getLintsCoreAnalysisOptions() async {
-  // try to load local lints from the resolved package references
-  if (_cachedLintsCoreInResolvedReferences != null) {
-    return _cachedLintsCoreInResolvedReferences!;
+  // Try to load local lints from the resolved package references.
+  if (_cachedLintsCoreInResolvedReferences case final cachedCoreLints?) {
```
isoos, line 25: I don't find this more readable than the `!= null` version. Could you please change it back (also the one below).

isoos (changes requested):

lib/src/analysis_options.dart
parlough: Fair! The syntax does take a bit of getting used to :)

I reverted this change but then ended up simplifying the implementation a bit further as we ended up only using one of the cached fields anyway. So this private method is now agnostic of the caching. Let me know what you think, thanks!

isoos (approved):

Issue #1324, opened 2024-01-31
Score down folks using legacy web APIs


kevmoo:
See https://api.dart.dev/main

This WILL be a lint, but we've missed the window for the February stable release.

Can we put something together in pana (then pub.dev) to give the following nudges.

kevmoo:
Can link to http://dart.dev/go/package-web which will point to "the right docs" when we're ready

mit-mit:
It seems a bit premature to take points away before the old APIs are clearly marked as deprecated and new APIs are marked stable?

kevmoo:
We've already added notes to the legacy APIs - https://api.dart.dev/main/dart-html/dart-html-library.html

We want a signal on pub first. Deprecations are loud and tend to cause lots of issues for the world. We just want to put (a bit) of pressure on the ecosystem.

We are not taking ANY points away. We're just giving the opportunity to earn more points for "doing the right thing" 😄

kevmoo:
Had a great chat with @isoos and @jonasfj about this. Going to repurpose this request to track packages that are "wasm ready" as a label and then give them some "bling" on the package site.

parlough:
Everyone will want some of the Wasm bling! I definitely like that more than reducing points as they can be a source of frustration.

kevmoo:
Like we could look into using the Wasm logo or something.

![image](https://github.com/dart-lang/pana/assets/17034/69119e94-8414-4be8-9c61-a4d07e23c6a8)

mosuem:
Is this still a P1 issue?

kevmoo:
I'd like some of the false positives/negatives to be prioritized. maybe not p1, though.

sigurdm:
@kevmoo, Can we try to flesh the remaining work out a bit.

What false positives/negatives do we know?

Currently a package gets is:wasm-ready iff:

* all `dart:` libraries transitively reachable in the import graph from the main library belongs to the set
  - 'dart:async',
  - 'dart:collection',
  - 'dart:convert',
  - 'dart:core',
  - 'dart:developer',
  - 'dart:math',
  - 'dart:typed_data',
  - 'dart:_internal',
  - 'dart:ui', 
  - 'dart:ui_web', 
  - 'dart:js_interop', 
  - 'dart:js_interop_unsafe'
 
Is that the right rule?

Do we have any other things we want to affect the decision?

kevmoo:
[...]

PR #1336, opened 2024-02-08
is:wasm-ready tag

sigurdm:
Package will get the is:wasm-ready tag if it uses only the set of libraries

Only adds the is:wasm-ready tag, does not count it in scoring or mention in the report.

towards https://github.com/dart-lang/pana/issues/1324


lib/src/tag/_specs.dart
```diff
@@ -89,6 +89,12 @@ class Runtime {
     tag: PanaTags.runtimeWeb,
   );
 
+  static final wasm = Runtime(
+    'wasm',
+    {..._onAllPlatforms},
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/tag/_specs.dart
```diff
@@ -216,6 +223,12 @@ class Platform {
     flutterRuntime: Runtime.flutterWeb,
     tag: PanaTags.platformWeb,
   );
+  static final wasm = Platform(
+    'Wasm',
+    dartRuntime: Runtime.wasm,
+    flutterRuntime: Runtime.wasm,
+    tag: PanaTags.platformWasm,
+  );
```
jonasfj, line 231: Should we not remove this for now?

lib/src/tag/pana_tags.dart
```diff
@@ -17,13 +17,15 @@ abstract class PanaTags {
...
   static const runtimeWeb = 'runtime:web';
+  static const isWasmReady = 'is:wasm-ready';
 
   // platform tags
   static const platformAndroid = 'platform:android';
   static const platformIos = 'platform:ios';
   static const platformLinux = 'platform:linux';
   static const platformMacos = 'platform:macos';
   static const platformWeb = 'platform:web';
+  static const platformWasm = 'platform:wasm';
```
jonasfj, line 28: And hold off on doing this?

jonasfj (approved):

lib/src/tag/_specs.dart
sigurdm: Ah yes

test/goldens/end2end/audio_service-0.18.10.json_report.md
```diff
@@ -20,11 +20,11 @@ Detected license: `MIT`.
 
 ## 20/20 Platform support
 
-### [*] 20/20 points: Supports 4 of 6 possible platforms (**iOS**, **Android**, **Web**, Windows, **macOS**, Linux)
```
isoos, line 23: I think you need to rebase the PR on HEAD.

sigurdm: Ah thanks! - I was getting very confused

isoos (approved):

lib/src/tag/_specs.dart
```diff
@@ -90,6 +91,12 @@ class Runtime {
     tag: PanaTags.runtimeWeb,
   );
 
+  static final wasm = Runtime(
+    'wasm',
+    {..._onAllPlatforms, 'ui', 'ui_web', 'js_interop', 'js_interop_unsafe'},
```
kevmoo, line 96: BUT!!! 

We need to check the EXCLUSION of `dart:js`, `dart:js_util`, `dart:html` (and) friends, and `pkg;js`!!

kevmoo (changes requested):

lib/src/tag/_specs.dart
isoos: This is an allow-list: if a package imports/exports anything else outside of these (e.g. `dart:js`), it won't get the tag. IIRC `package:js` exports `dart:js_util` and if something depends on it, it will be applied to it too.

kevmoo: Do we test that?

kevmoo: I'd love to have a test that validates each of these independently. just to be sure!

isoos: Not specifically these, but 3 packages in our end2end tests got the tag (`async`, `http`, `lints`), while others e.g. `dnd` and `url_launcher` did not get it, which seems to be about right to me.

isoos (approved):

lib/src/tag/_specs.dart
sigurdm: Added some tests - they revealed problems when the package is a flutter package it would never get the tag because we were special-casing for the dart sdk. I tried to generalize.

jonasfj (approved):

PR #1346, opened 2024-02-27
Fix configuration directory issue.

isoos:
- added missing reference to `flutterSdk.environment`
- handling the `Welcome to Flutter` screen by stripping it away from the output (in case any future regression happens)
- added config dir option to end2end test, it may help to expose future regression

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -295,18 +295,27 @@ class ToolEnvironment {
   Future<Map<String, dynamic>> _getFlutterVersion() async {
     final result = await runConstrained(
       [..._flutterSdk.flutterCmd, '--version', '--machine'],
+      environment: _flutterSdk.environment,
       throwOnError: true,
     );
-    final waitingForString = 'Waiting for another flutter';
-    return result.parseJson(transform: (content) {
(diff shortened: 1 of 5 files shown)


lib/src/sdk_env.dart
```diff
@@ -295,18 +295,27 @@ class ToolEnvironment {
...
             .where((e) => !e.contains(waitingForString))
             .join('\n');
-      } else {
-        return content;
       }
+      // filter for welcome screen
+      if (content.contains('Welcome to Flutter!')) {
+        final lines = content.split('\n');
+        final separator = lines.indexWhere((l) => l.trim().isEmpty);
+        if (separator >= 0) {
```
sigurdm, line 314: This looks ok to me, but we almost need a test for this logic - otherwise it might end up doing more harm than good.

isoos: Added a test for it. Although, I think this is the `unknown unknown` category, where we can't possibly know what will break us, only hope that we can catch it early.

sigurdm: True - but I have made "unknown unknown" fixes before, that turned out to themselves be broken, and thus just make things worse :) .

sigurdm (approved):

PR #1347, opened 2024-02-27
Starting each analysis from scratch in end2end tests.

isoos:
As a follow-up to #1346, this makes sure that all end2end test start with a fresh temp directory, pub cache and configuration, making sure that one temporary error at one analysis does not hide an issue with a later analysis.

### DIFF
--- a/test/end2end_test.dart
+++ b/test/end2end_test.dart
@@ -18,45 +18,6 @@ final _goldenDir = p.join('test', 'goldens', 'end2end');
 final _pubDevUri = Uri.parse('https://pub.dartlang.org/');
 
 void main() {
-  late String tempDir;
-  late PackageAnalyzer analyzer;
-  http.Client? httpClient;
-  late HttpServer httpServer;
-
-  setUpAll(() async {
-    tempDir = Directory.systemTemp
-        .createTempSync('pana-test')
-        .resolveSymbolicLinksSync();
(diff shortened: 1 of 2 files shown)


test/end2end_test.dart
```diff
@@ -18,56 +18,47 @@ final _goldenDir = p.join('test', 'goldens', 'end2end');
...
   }) {
     final filename = '$package-$version.json';
     group('end2end: $package $version', () {
+      late String tempDir;
+      late PackageAnalyzer analyzer;
+      http.Client? httpClient;
+      late HttpServer httpServer;
       late final Map<String, Object?> actualMap;
 
       setUpAll(() async {
```
sigurdm, line 34: Should all this setup code not be in a `setUp` rather than `setUpAll`?

isoos: We have multiple `test()` methods, I'd rather keep it like that.

sigurdm: But doesn't that still create only one analysis context for all tests?

isoos: `setupAll` runs only a single time for all the tests in the group, I think that's what we want here. Not sure: what do you see as a problem here?

sigurdm: Ah - I think I saw it as residing at a higher level - sorry.

sigurdm (approved):

PR #1353, opened 2024-04-16
Log dartdoc output.

isoos:


isoos:
This is to expose `dartdoc` output issues like https://github.com/dart-lang/pub-dev/issues/7644

sigurdm (approved):

PR #1368, opened 2024-05-16
Fix spdx license updater + remove package:archive dependency.

isoos:


sigurdm (approved):
Nice
... this is not a tool end-users need to run, right?
Will there be a problem on windows?
Does this work on macos?

isoos:
> ... this is not a tool end-users need to run, right?

Yes, only pana developers may run this tool.

> Will there be a problem on windows?
> Does this work on macos?

It is likely that it won't work on windows, but may work on macos (if `unzip` is present, I don't recall if it is by default).

PR #1379, opened 2024-06-18
Verify 'executables' in 'pubspec.yaml' and check if the related files exist

isoos:
- fixes bug with `has:executable` tagging
- verifies the `executables` entry in `pubspec.yaml` and reports it in the layout checks if any of the binary files is missing

### DIFF
--- a/lib/src/report/template.dart
+++ b/lib/src/report/template.dart
@@ -183,6 +183,20 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
               '${repository!.verificationFailure}'));
     }
 
+    final executableFiles = await context.executablesInBinDirectory;
+    for (final e in context.pubspec.executables.entries) {
+      final filename = e.value.endsWith('.dart')
+          ? e.value
+          : (e.value.isEmpty ? '${e.key}.dart' : '${e.value}.dart');
+      if (!executableFiles.contains(filename)) {
+        issues.add(Issue(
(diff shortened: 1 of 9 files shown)


lib/src/report/template.dart
```diff
@@ -183,6 +183,20 @@ Future<ReportSection> followsTemplate(PackageContext context) async {
...
+          ? e.value
+          : (e.value.isEmpty ? '${e.key}.dart' : '${e.value}.dart');
+      if (!executableFiles.contains(filename)) {
+        issues.add(Issue(
+          'Missing or invalid executable file: `bin/$filename`.',
+          suggestion: 'Update `executable` field in `pubspec.yaml` or '
+              'fix the missing or invalid file in the `bin/` directory.',
+        ));
+      }
+    }
```
jonasfj, line 198: I think this is correct, but I have to ask? Are we sure there isn't a diagnostic covering this already?
And if not, should we consider adding it to the `analyzer` instead?

jonasfj (approved):

lib/src/report/template.dart
isoos: writing `pana: pana2.dart` in pana's `pubspec.yaml` does not reveal anything during analysis. I agree, it could get inside `analyzer` too, and once it does, we may just remove it from here. But until then, it is a cheap check we can just do.

PR #1380, opened 2024-07-01
Upgraded tar dependency + CI versions

isoos:


sigurdm (approved):

isoos:
Note: apparently the end2end tests behave differently on CI compared to my local environment, and the golden files don't match. Using a Dart SDK that is matching the stable Flutter's Dart SDK seems to solve the issue. In the future we should probably update the end2end test with configurable SDK selection that is downloading the SDK if needed.

Issue #1369, opened 2024-05-21
Analyze packages with `resolution:workspace`.


sigurdm:
Related to https://github.com/dart-lang/pub/issues/4127

sigurdm:
Either we

* Remove `resolution` and `workspace` sections from the pubspec before analyzing.
* Make a "dummy package" that depends on the package for resolution.

jakemac53:
With workspaces effectively shipped in 3.5 stable (for Dart packages at least), should we prioritize work on this? I migrated the build/test repos to use workspaces, but just realized that means they won't get docs published for new versions.

sigurdm:
https://pub.dev/packages/build_runner/score seems good now!

PR #1394, opened 2024-08-08
Remove workspace and resolution from pubspec.yaml before resolving

sigurdm:
Fixes https://github.com/dart-lang/pana/issues/1369


isoos (commented):
Let's also make a note in changelog + add version bump

isoos (approved):

PR #1408, opened 2024-10-10
Pass-through for formatter/page_width in analysis_options.yaml

### LINKED ISSUE #1405: Respect `formatter: page-width` setting in `analysis_options.yaml`
In the context of https://github.com/dart-lang/sdk/issues/56863
We have to make sure that we respect the page-width setting that users have set.

isoos:
Fixes #1405

### DIFF
--- a/lib/src/analysis_options.dart
+++ b/lib/src/analysis_options.dart
@@ -108,5 +108,15 @@ String updatePassthroughOptions({
     }
   }
 
+  final origFormatter = origMap['formatter'];
+  if (origFormatter is Map) {
+    final pageWidth = origFormatter['page_width'];
+    if (pageWidth is int) {
+      final customFormatter =
+          customMap.putIfAbsent('formatter', () => <String, Object?>{}) as Map;
+      customFormatter['page_width'] = pageWidth;
+    }
+  }
+
   return json.encode(customMap);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.22.13
+
+- Pass-through for `formatter/page_width` in `analysis_options.yaml`.
+- Upgraded `lints` to `^5.0.0`
+
 ## 0.22.12
 
 - Updated report text for `wasm` readiness.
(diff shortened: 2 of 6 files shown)


jonasfj (approved):

lib/src/analysis_options.dart
```diff
@@ -108,5 +108,15 @@ String updatePassthroughOptions({
...
 
+  final origFormatter = origMap['formatter'];
+  if (origFormatter is Map) {
+    final pageWidth = origFormatter['page_width'];
+    if (pageWidth is int) {
+      final customFormatter =
+          customMap.putIfAbsent('formatter', () => <String, Object?>{}) as Map;
+      customFormatter['page_width'] = pageWidth;
+    }
+  }
```
munificent, line 119: FYI this could also be:

```dart
  if (origMap['formatter'] case {'page_width': int pageWidth}) {
    final customFormatter =
        customMap.putIfAbsent('formatter', () => <String, Object?>{}) as Map;
    customFormatter['page_width'] = pageWidth;
  }
```

jonasfj: @lrhn only if you use dart format with a page width of at-least 103 chars :rofl: 

Otherwise it becomes (once formatted):
```dart
  if (origMap['formatter'] case {'page_width': int pageWidth}) {
    (customMap.putIfAbsent('formatter', Map<String, Object?>.empty)
        as Map)['page_width'] = pageWidth;
  }
```

Personally, I'd suggest it's more readable (albeit a bit dumb) to do:
```dart
  if (origMap['formatter'] case {'page_width': int pageWidth}) {
    customMap['formatter'] ??= <String, Object?>{};
    customMap['formatter']['page_width'] = pageWidth;
  }
```

or 
```dart
  if (origMap['formatter'] case {'page_width': int pageWidth}) {
    customMap['formatter'] = {
      ...?customMap['formatter'],
      'page_width': pageWidth,
    };
  }
```
Arguably, they aren't as efficient, but they read nicely -- which is sadly almost always the most important thing (at-least in `pana`).

PR #1412, opened 2024-10-24
Detect Swiftpm support

### LINKED ISSUE #1402: Score support for swift package manager in plugins with ios/macos support
Reading:
https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin

As far as I can tell we need to detect the presence of a `<plugin_name>/<platform>/Package.swift` file. Where platform is respectively `ios` and `macos` for the two platforms, or `darwin` that covers both.

I suggest we do an initial zero-scoring report note as we did for wasm support:

* If the package is an ios plugin:
  * if it has `<plugin_name>/ios/Package.swift` or `<plugin_name>/darwin/Package.swift` file, mark platform support as fully passing
   * otherwise write "in the future this will receive a reduced score, please migrate to Swift package manager (link)"
* Similarly for macos

sigurdm:
Fixes #1402


lib/src/tag/pana_tags.dart
```diff
@@ -38,6 +38,7 @@ abstract class PanaTags {
   static const hasScreenshot = 'has:screenshot';
   static const isPlugin = 'is:plugin';
   static const isNullSafe = 'is:null-safe';
+  static const isSwiftPackageManagerReady = 'is:swiftpm-plugin';
```
isoos, line 41: I think we should call the variable as close as the field value is, e.g. `isSwiftPMPlugin` or similar.

isoos (approved):
lgtm with the variable rename

isoos (commented):
Also: please update the CHANGELOG

lib/src/tag/pana_tags.dart
sigurdm: Nice catch - forgot to rename this!

lib/src/report/multi_platform.dart
```diff
@@ -118,13 +118,19 @@ Future<ReportSection> multiPlatform(PackageContext context) async {
...
   return makeSection(
       id: ReportSectionId.platform,
       title: 'Platform support',
       maxPoints: 20,
       basePath: context.packageDir,
-      subsections: [subsection, wasmSubsection],
+      subsections: [
+        subsection,
+        wasmSubsection,
+        if (swiftPackageManagerSubsection != null) swiftPackageManagerSubsection
```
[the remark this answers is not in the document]
sigurdm: I think that is becoming a thing later. https://github.com/dart-lang/language/blob/main/accepted/future-releases/0323-null-aware-elements/feature-specification.md

lib/src/report/multi_platform.dart
```diff
@@ -178,3 +184,49 @@ Future<Subsection> _createWasmSubsection(PackageContext context) async {
...
+    PackageContext context) async {
+  final tr = await context.staticAnalysis;
+  final description = 'Swift Package Manager support';
+
+  if (tr.tags.contains(PanaTags.isSwiftPmPlugin)) {
+    return Subsection(
+      description,
+      [
+        RawParagraph(
+            'This iOS or MacOs plugin supports the Swift Package Manager. '
```
loic-sharma, line 200:
```suggestion
            'This iOS or MacOS plugin supports the Swift Package Manager. '
```

lib/src/report/multi_platform.dart
```diff
@@ -178,3 +184,49 @@ Future<Subsection> _createWasmSubsection(PackageContext context) async {
...
+            'It will be rewarded additional points in a future version of the scoring model.'),
+        RawParagraph(
+            'See https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors for details.'),
```
loic-sharma, line 203: Consider linking to `https://docs.flutter.dev/to/spm` instead. This redirects to the docs we have on SwiftPM. If we ever move the docs location, we'll update that redirect so that we don't break tooling that links to it.

lib/src/report/multi_platform.dart
```diff
@@ -178,3 +184,49 @@ Future<Subsection> _createWasmSubsection(PackageContext context) async {
...
+  final explanation = tr.explanations
+      .where((e) => e.tag == PanaTags.isSwiftPmPlugin)
+      .firstOrNull;
+  if (explanation != null) {
+    return Subsection(
+      description,
+      [
+        explanationToIssue(explanation),
+        RawParagraph(
+          'This package for iOS or MacOs does not support the Swift Package Manager. '
```
loic-sharma, line 219:
```suggestion
          'This package for iOS or MacOS does not support the Swift Package Manager. '
```

lib/src/tag/tagger.dart
```diff
@@ -355,6 +350,53 @@ class Tagger {
...
+    var swiftPmSupport = true;
+
+    for (final darwinOs in ['macos', 'ios']) {
+      if (pathExists(
+          pubspec.originalYaml, ['flutter', 'plugin', 'platforms', darwinOs])) {
+        isDarwinPlugin = true;
+        final specificPackageSwiftFile =
+            path.join(darwinOs, packageName, 'Package.swift');
+        final genericPackageSwiftFile =
+            path.join('darwin', packageName, 'Package.swift');
```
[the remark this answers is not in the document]
loic-sharma: > Does `ffiPlugin` only affect method channels, and not affect the need for a podspec or `Package.swift`? (I'm not sure entirely sure, just guessing)

From my understanding, a plugin can support CocoaPods & SwiftPM and also be an `ffiPlugin`. For example: https://docs.flutter.dev/platform-integration/ios/c-interop#open-source-third-party-library

cc @dcharkes to confirm.

> Is `sharedDarwinSource: true` the only circumstance under which we should look in `darwin/<packageName>/Package.swift` ? (I'm suspecting so)

Yup that's correct

> It would probably be useful with Flutter plugin reference documentation.
> Similar to [pubspec reference docs](https://dart.dev/tools/pub/pubspec) covering the properties under `flutter.plugin` and what files they imply exist.

This is an excellent idea. I filed this issue: https://github.com/flutter/website/issues/11334

lib/src/tag/tagger.dart
```diff
@@ -355,6 +350,53 @@ class Tagger {
...
+        if (!(File(path.join(packageDir, specificPackageSwiftFile))
+                .existsSync() ||
+            File(path.join(packageDir, genericPackageSwiftFile))
+                .existsSync())) {
+          swiftPmSupport = false;
+          explanations.add(Explanation(
+              'Package does not support the Swift Package Manager on $darwinOs',
```
loic-sharma, line 385: Super nitpick: could we use the `iOS` and `macOS` casings here? I don't feel strongly

loic-sharma:
This is awesome, thanks for adding this! 💙

lib/src/report/multi_platform.dart
sigurdm: Done

lib/src/tag/tagger.dart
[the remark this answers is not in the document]
jonasfj: @sigurdm should we take `sharedDarwinSource` into account?

loic-sharma (approved):
Looks great, thanks for adding this!

PR #1413, opened 2024-10-24
Ignore git error on repository root detection.

isoos:
Needed for https://github.com/dart-lang/pub-dev/pull/8173

### DIFF
--- a/lib/src/package_analyzer.dart
+++ b/lib/src/package_analyzer.dart
@@ -251,7 +251,7 @@ Future<String?> _detectGitRoot(String packageDir) async {
       return pr.stdout.asString.trim();
     }
   } on GitToolException catch (_) {
-    // ignore exception
+    // not in a git directory (or git is broken) - ignore exception
   }
   return null;
 }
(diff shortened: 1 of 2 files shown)


lib/src/package_analyzer.dart
```diff
@@ -242,12 +242,16 @@ class PackageAnalyzer {
...
-    workingDirectory: packageDir,
-  );
-  if (pr.exitCode == 0) {
-    return pr.stdout.asString.trim();
+  try {
+    final pr = await runGitIsolated(
+      ['rev-parse', '--show-toplevel'],
+      workingDirectory: packageDir,
+    );
+    if (pr.exitCode == 0) {
```
sigurdm, line 250: Can this ever be false?

lib/src/package_analyzer.dart
```diff
@@ -242,12 +242,16 @@ class PackageAnalyzer {
...
+      return pr.stdout.asString.trim();
+    }
+  } on GitToolException catch (_) {
+    // ignore exception
```
sigurdm, line 254:
```suggestion
    // not in a git directory (or git is broken) - ignore exception
```

sigurdm (approved):
LGTM with suggestions

sigurdm (approved):

lib/src/package_analyzer.dart
isoos: No, it is always true. I've removed that check.

Issue #1419, opened 2024-11-19
iOS/macOS packages are incorrectly labeled as SwiftPM incompatible if they don't have native code.


sigurdm:
For example this federated package: https://pub.dev/packages/shared_preferences/score

This package supports iOS/macOS, but all the native code is in a dependency, which does support SwiftPM: https://pub.dev/packages/shared_preferences_foundation/score

Could we update the pana tagger to flag a package as SwiftPM incompatible only if it has a <platform directory>/<plugin name>.podspec file?

For example, if the shared_preferences plugin has a ios/shared_preferences.podspec file.

stuartmorgan:
> Could we update the pana tagger to flag a package as SwiftPM incompatible only if it has a /.podspec file?
> 
> For example, if the shared_preferences plugin has a ios/shared_preferences.podspec file.

Capturing here in addition to the PR: the authoritative indication of whether or not there should be native code is the presence of the `pluginClass` key ([example](https://github.com/flutter/packages/blob/761812665448c7e5a1bd5810aeabfac1ac027038/packages/shared_preferences/shared_preferences_foundation/pubspec.yaml#L16)).

stuartmorgan-g:
I think "bad-swiftpm" is not the best way to think about this. It's perfectly valid for something to be conceptually a plugin (federated or not) but have an iOS and/or macOS implementation that doesn't use the Flutter native plugin machinery (could be Dart only, could be using build hooks, etc.). The only thing that is a problem is a plugin that is using the non-SwiftPM (i.e., CocoaPods) native plugin machinery on iOS/macOS.

Something like "legacy-native-build" would be a better way to express the thing that is a problem.

PR #1420, opened 2024-11-19
Only mark package swiftpm incompatible when pluginClass is present.

sigurdm:
Fixes: https://github.com/dart-lang/pana/issues/1419


lib/src/tag/tagger.dart
```diff
@@ -378,8 +378,10 @@ class Tagger {
             ? 'darwin'
             : darwinOs;
 
+        final podspecFile = path.join(osDir, '$packageName.podspec');
```
isoos, line 381: where is the osDir coming from?

sigurdm: So if a platform is using `sharedDarwinSource` we look in `darwin`, otherwise we look in the platform's name (eg. `ios`).

isoos: Ah, sorry, I didn't do `git pull upstream...` before checking the surrounding code.

isoos (approved):

loic-sharma (approved):
Thanks for updating this!

stuartmorgan:
This is not a very future-proof test, since eventually we expect *no* plugins to have a podspec. I would recommend skipping if the relevant platform section of the pubspec doesn't contain `pluginClass`, as that's the key that tells Flutter to look for native code.

sigurdm:
That makes a lot more sense. Updated.

loic-sharma (approved):
Re-LGTM!

PR #1424, opened 2024-11-26
Detect library hrefs with or without ending slash.

isoos:
This is assuming that the new href URLs in dartdoc will get an ending slash: https://github.com/dart-lang/dartdoc/issues/3936

### DIFF
--- a/lib/src/dartdoc/dartdoc_internals.dart
+++ b/lib/src/dartdoc/dartdoc_internals.dart
@@ -17,6 +17,7 @@ bool isHrefALibrary(String? href) {
   if (href.endsWith('.html')) {
     return false;
   }
-  // libraries have no slash in their hrefs
-  return !href.contains('/');
+  final segments = href.split('/').where((s) => s.isNotEmpty).toList();
+  // libraries have only a single segment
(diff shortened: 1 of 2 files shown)


lib/src/dartdoc/dartdoc_internals.dart
```diff
@@ -17,6 +17,7 @@ bool isHrefALibrary(String? href) {
   if (href.endsWith('.html')) {
     return false;
   }
-  // libraries have no slash in their hrefs
-  return !href.contains('/');
+  final segments = href.split('/').where((s) => s.isNotEmpty).toList();
```
sigurdm, line 20: This also removes empty segments in the middle `a//b->a/b` is that ever a problem? should we test for an empty segment only at the end?

sigurdm (approved):

lib/src/dartdoc/dartdoc_internals.dart
isoos: In this specific case, it would be considered as a bug in dartdoc, however, it doesn't hurt if we remove only the last one. Done.

PR #1428, opened 2024-12-09
Use 'flutter pub global activate dartdoc' for Flutter packages.

isoos:


sigurdm (commented):
Is this really needed?

sigurdm (approved):
I get it now - we need to activate and run with the same sdk

PR #1430, opened 2024-12-27
Make example's README.md priority over any dart file in the examples directory.

isoos:
- https://github.com/dart-lang/pub-dev/issues/8436
- https://dart.dev/tools/pub/package-layout#examples says that "if your package has many files under its example directory, including a file named README.md, then your package's Example tab displays the contents of example/README.md", which means we should bump the README.md to the top of that list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.22.18-dev
+
+- Make example's `README.md` priority over any dart file in the examples directory.
+
 ## 0.22.17
 
 - Fixed lower dependency constraint for `cli_util`.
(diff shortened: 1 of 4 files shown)


.github/workflows/test-package.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.6.0-334.4.beta]
+        sdk: [3.6.0]
```
sigurdm, line 24: Should we do this in a separate PR?

sigurdm (commented):
Seems reasonable.

Thinking of it - should the 'example' tab have a paragraph for promoting the `dart pub unpack` command for exploring the example locally?

sigurdm (approved):

isoos:
Rebased on #1430.

Issue #1403, opened 2024-10-04
If repository verification finds bad repository link: 0 points


sigurdm:
We want to strongly encourage correct repository links.
A bad link is often indicative of malattribution. And can lend false credibility to a packages.

The repository verification process can have 3 outcomes:

* valid
* invalid
* failed deciding validity

We only want to decrease the score in the second case.

The score for an invalid repository link should be 0 (total, for the full report).
Or we find an alternative way to make them not show up in search (perhaps set the discontinued status of the package).

sigurdm:
Current status: we no longer display bad repository links. If validation fails because of eg. connection issues they are still shown.

They are thus getting a lower score, but still not 0. Do we still want that?

cc @jonasfj

PR #1433, opened 2025-01-08
Expose more precise repository verification status in 'AnalysisResult'.

isoos:
- #1403
- I think we may have a path to introduce stricter checks without breaking the pana output much, this is a first step to it.
- Note: this follows the previous concept of keeping the `AnalysisResult` object compact, while exposing any issue or failure message in the report.

### DIFF
--- a/test/goldens/end2end/lints-1.0.0.json_report.md
+++ b/test/goldens/end2end/lints-1.0.0.json_report.md
@@ -6,6 +6,22 @@
 
 * `pubspec.yaml` doesn't have a `repository` entry.
 
+<details>
+<summary>
+Failed to verify repository URL.
+</summary>
+
+Please provide a valid [`repository`](https://dart.dev/tools/pub/pubspec#repository) URL in `pubspec.yaml`, such that:
+
+ * `repository` can be cloned,
+ * a clone of the repository contains a `pubspec.yaml`, which:,
+    * contains `name: lints`,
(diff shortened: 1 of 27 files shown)


test/goldens/end2end/lints-1.0.0.json_report.md
```diff
@@ -6,6 +6,22 @@
 
 * `pubspec.yaml` doesn't have a `repository` entry.
 
+<details>
+<summary>
+Failed to verify repository URL.
+</summary>
+
+Please provide a valid [`repository`](https://dart.dev/tools/pub/pubspec#repository) URL in `pubspec.yaml`, such that:
```
sigurdm, line 14: Do we want to write this on a missing repository link? Closed source packages might not have any.

isoos: I think if we are not giving points other packages get, we should give a reason and ways to fix it.
The source is available with the package, the license may restrict its use, but I don't think that should be a concern here.

sigurdm: Are we giving less points when a repository is not present?

isoos: As discussed offline, I've updated a code to only emit this if the verification explicitly failed. PTAL.

sigurdm (approved):

PR #1444, opened 2025-03-03
Passthru dartdoc option for showUndocumentedCategories

jonasfj:
We think the `showUndocumentedCategories` might allow topics that don't have any `/// {@category <category>}` annotated elements to also be displayed.

CC @HosseinYousefi

### DIFF
--- a/lib/src/dartdoc/dartdoc_options.dart
+++ b/lib/src/dartdoc/dartdoc_options.dart
@@ -61,5 +61,6 @@ final _passThroughKeys = <String>[
   // 'examplePathPrefix',
   'exclude',
   'include',
+  'showUndocumentedCategories',
   'nodoc',
 ];


HosseinYousefi (approved):

lib/src/dartdoc/dartdoc_options.dart
```diff
@@ -61,5 +61,6 @@ final _passThroughKeys = <String>[
   // 'examplePathPrefix',
   'exclude',
   'include',
+  'showUndocumentedCategories',
```
isoos, line 64: Note: we do have `'showUndocumentedCategories': true` on line 52. Isn't that enough?

jonasfj: Yes, that's should be enough, no idea why it's not working.

jonasfj:
Hmm, I guess this isn't a fix then.

PR #1447, opened 2025-03-04
Lower score for packages with discontinued dependencies

### LINKED ISSUE #1445: Lower score in "up-to-date dependencies" if direct dependency is "discontinued"
Depending on a discontinued package is not healthy for the ecosystem in the long run

It could be seen as equivalent to not allowing the latest version of a dependency in your constraint.

cc @kevmoo @mit-mit

sigurdm:
Fixes #1445


isoos:
Looks good to me, but needs to be rebased.

isoos (approved):

Issue #1448, opened 2025-03-10
Investigate while page_width is not applied for dart format check


isoos:
Original report:
https://github.com/dart-lang/pub-dev/issues/3041#issuecomment-2709061492

package version example to check: `uvccamera 0.0.12`

isoos:
The root cause seems to be the following combination in the analysis context:

- The package uses `flutter_lints` (referenced in `dev_dependencies` and `analyzer_options.yaml` as it should be). The package uses `page_width` as it should be.
- `pana` removes `dev_dependencies` before analyzing the package (no `flutter_lints` after this in the resolved dependencies for this package).
- `pana` updates the `analyzer_options.yaml` for running the `dart analyze` command (removes any custom include file reference, uses a standard set of lints, allows transparent bypass for a specific subset of options), but not for `dart format`.
- Since `dart format` gets the original content, but without the dev dependencies, it is not able to find the referenced `flutter_lints` from `analyzer_options.yaml`. It swallows the error and does not report it, and surprisingly it also ignores the `page_width` property.
- With ignoring the property, it falls back to the default 80 character length lines, causing wrong analysis.

The narrowest seems to be to also update the `analysis_options.yaml` for `dart format`, and we should discuss if other parts of the toolchain needs any change.

isoos:
The fix has been deployed, closing.

PR #1449, opened 2025-03-10
Use restricted analysis_options file also when formatting

sigurdm:
#1448


isoos:
Let's also mention the fix in the changelog!

isoos (approved):

PR #1453, opened 2025-03-17
Filter SDK package 'flutter' from the outdated report.

isoos:


sigurdm (commented):
It would be better to filter all sdk packages out.

I think we need to either read pubspec.lock or the output of `dart pub deps` to know the source of the dependencies (the source of direct dependencies can be read from the pubspec.yaml, but for transitive dependencies we need one of those)

isoos:
Would it be enough to filter the `source: sdk` in `pubspec.lock`?

sigurdm:
> Would it be enough to filter the source: sdk in pubspec.lock?

Yes, I think that is the best way to do it!

sigurdm (approved):
Nice!

Issue #1487, opened 2025-09-29
Implement/use token-based diff algorithm for license diffs


isoos:
#1485 introduces a token-based longest common subsequence calculation. With that, we will have two diffs algorithms, and we should replace the current text-based with the token-based algorithm.

PR #1490, opened 2025-09-30
Calculate token-based edit operations.

isoos:
- #1487
- group tokens under operations
- backtracking the distance table no longer requires repeating the string comparison

### DIFF
--- a/lib/src/license_detection/lcs.dart
+++ b/lib/src/license_detection/lcs.dart
@@ -129,9 +129,20 @@ List<TokenOp> _dynamicLcs({
   var j = n;
 
   while (i > 0 || j > 0) {
+    // shortcut: only deletes left
+    if (i == 0) {
+      opsBackwards.add(DeleteOp(known.take(j).toList()));
+      break;
+    }
+    // shortcut: only inserts left
+    if (j == 0) {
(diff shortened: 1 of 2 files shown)


isoos:
Note: this does not replace the string-based diff yet, only the coverage calculation uses it.

lib/src/license_detection/lcs.dart
```diff
@@ -91,22 +123,63 @@ Iterable<TokenPair> _dynamicLcs({
...
+      } while (isDelete() && !isMatch());
+      opsBackwards.add(DeleteOp(deleteTokens.reverseIfNeeded()));
+    }
+  }
+
+  return opsBackwards.reverseIfNeeded();
+}
+
+extension _ListExt<T> on List<T> {
+  List<T> reverseIfNeeded() {
```
sigurdm, line 179: Consider just always reversing.
It doesn't seem meaningful to save the overhead here...

sigurdm (approved):

lib/src/license_detection/lcs.dart
isoos: I did also consider to do in-place replacing the items in the list, but wanted to start with a simpler code. I think it is not worth it for the coverage calculation, but may be worth it when we replace the text-diff with the token-diff.

PR #1492, opened 2025-09-30
Fix bug in removeOverLappingMatches

### LINKED ISSUE #1488: License detection overlap removal fails
When the `SSH-OpenSSH` license is matched against other licenses, its smaller version is recognized inside the full range. The current overlap removal logic does not remove the smaller version, but we think it should.

isoos:
Fixes #1488.


isoos:
Apparently there is a test which somewhat contradicts with the fix, since the removal algorithm was ambiguous:
https://github.com/dart-lang/pana/blob/master/test/license_detector_test.dart#L132-L145

The test says that while we have the enclosed region with less density, we shouldn't remove it, as it has better confidence score. But I think that makes the algorithm ambiguous (as we've seen with the SSH example), and it is better to change the test specification.

sigurdm (approved):

PR #1520, opened 2025-11-07
Adapt golden testing framework from pub client

sigurdm:
(adapted from https://github.com/dart-lang/pub/blob/master/test/golden_file.dart)

With this:
* golden files gets their name from the running test automatically
* there is a concept of multiple sections in a golden file
* you can update the golden files by running the tests with `_WRITE_GOLDENS=TRUE`.
* You can run a command, and have the command, stdout, stderr and exit code all included in the golden file.
  * and have specific expectation machers on those at the same time
* missing golden files cause an error on CI.


isoos:
While the filenames with spaces are probably a non-issue, shouldn't we rather have `-` or `_` instead of spaces?

PR #1539, opened 2026-01-12
Fix exposing the SDK root path in the environment.

isoos:


isoos:
> Remind me what the SDK root path is used for?

This will be used as a parameter in a future sandbox-wrapper script.

PR #1540, opened 2026-01-12
Always use 'flutter pub pub' when invoking 'pub' with Flutter SDK.

isoos:
- Unrolls `_runPub` which makes the actual `pub` call a bit more obscure, and we didn't use it for `pub outdated` anyway.
- Updates `flutter pub` and `flutter packages pub` calls, uses `flutter pub pub` (double `pub`) instead.


sigurdm (approved):
Please update description

LGTM

PR #1552, opened 2026-01-21
Add --license-data option to support AOT compilation

jonasfj:


isoos (changes requested):
Also add a note in changelog.

PR #1588, opened 2026-06-16
Fix: git file list parsing bug.

isoos:
The previous version did drop the first entry silently.

### DIFF
--- a/lib/src/tool/git_tool.dart
+++ b/lib/src/tool/git_tool.dart
@@ -190,10 +190,14 @@ class GitTool {
           GitToolException('Could not list `$ref`.', pr.asTrimmedOutput),
     );
     return pr.stdout.asBytes
-        .splitAfter((b) => b == 0)
-        .map((chunk) => chunk.where((b) => b != 0).toList())
+        .splitBefore((b) => b == 0)
         .where((chunk) => chunk.isNotEmpty)
-        .map(utf8.decode)
+        .map(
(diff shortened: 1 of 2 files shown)


lib/src/tool/git_tool.dart
```diff
@@ -173,10 +190,10 @@ class GitTool {
...
     );
     return pr.stdout.asBytes
-        .splitBefore((b) => b == 0)
+        .splitAfter((b) => b == 0)
+        .map((chunk) => chunk.where((b) => b != 0).toList())
         .where((chunk) => chunk.isNotEmpty)
-        .map((chunk) => chunk.first == 0 ? utf8.decode(chunk.sublist(1)) : '')
-        .where((item) => item.isNotEmpty)
+        .map(utf8.decode)
         .toList();
```
sigurdm: Maybe this would also work (and create fewer sublists)
```suggestion
        .splitBefore((b) => b == 0)
        .where((chunk) => chunk.isNotEmpty)
        .map((chunk) => chunk.first == 0 ? utf8.decode(chunk.sublist(1)) : utf8.decode(chunk.sublist(0)))
        .where((item) => item.isNotEmpty)
        .toList();
```

sigurdm (approved):

lib/src/tool/git_tool.dart
isoos: Done. (github didn't merge this directly, updated as a separate commit)

PR #1592, opened 2026-06-19
Detect legacy Kotlin configuration in Android plugins

### LINKED ISSUE #1589: Warn of Future Penalization When Not Migrated to Built-in Kotlin
From the Built-in Kotlin migration docs for plugin authors:
https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

Please add support to warn plugins of future penalization if the plugin's newest release has not migrated to built-in kotlin and continues to apply Kotlin Gradle Plugin (KGP). 

To audit whether or not a plugin has been completely migrated to built-in kotlin, we must evaluate a plugin module's `build.gradle(.kts)` file located at `<plugin_name>/android/build.gradle(.kts)`.

Plugin **should be penalized** if in `<plugin_name>/android/build.gradle(.kts)`:
1. KGP is applied. See examples in migration doc [here](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#:~:text=legacy%20apply-,Before%3A,-%3Capp%2Dsrc%3E/android).
2. OR `android.kotlinOptions{}` block still exists.  see example in migration doc [here](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#:~:text=legacy%20apply-,Before%3A,-%3Capp%2Dsrc%3E/android).

Plugin **should not be penalized** if in `<plugin_name>/android/build.gradle(.kts)`:
**pre-req**
1. KGP was previously applied in a prior release version.

> **Note:** If KGP was NEVER applied, it means the Flutter plugin never needed/used KGP to compile Kotlin source code. No built-in kotlin migration necessary. Good pattern and should not be penalized. 

1. KGP is not applied
2. AND `kotlin.compilerOptions{}` exists

This regex to detect if KGP was declared may be useful: https://github.com/flutter/flutter/blob/d135f1e569d2638c88a20054e147a0bcd9b4c6ac/packages/flutter_tools/gradle/src/main/kotlin/FlutterPluginUtils.kt#L86-L93

sigurdm:
Adds detection for legacy Kotlin Gradle Plugin (KGP) application and legacy `android.kotlinOptions` blocks in Android plugins.
Formulates the tag in the positive (`is:built-in-kotlin`), which is added to Android plugins that do not use legacy Kotlin configuration.
Updates the report to show a positive note when they are ready, or a warning when they are using legacy configuration. Still no scoring.

Also updates CHANGELOG and sets version to `0.23.12`.

Fixes: https://github.com/dart-lang/pana/issues/1589


sigurdm:
cc @jesswrd - does this work for you?

lib/src/tag/tagger.dart
```diff
@@ -640,4 +707,30 @@ class Tagger {
       );
     }
   }
+
+  bool _hasLegacyKotlinGroovy(String content) {
+    final kgpRegexGroovy = RegExp(
```
isoos, line 747: The regexpes here are a bit hard to digest. Please add some explanation and/or segment it.

sigurdm: Good idea!

isoos (approved):

lib/src/tag/tagger.dart
```diff
@@ -474,6 +475,73 @@ class Tagger {
...
+        hasLegacyKotlin = true;
+        buildGradlePath = 'android/build.gradle.kts';
+      }
+    }
+
+    if (hasLegacyKotlin) {
+      explanations.add(
+        Explanation(
+          'Legacy Kotlin plugin DSL detected in `$buildGradlePath`.',
+          'This plugin applies Kotlin Gradle Plugin (KGP) or uses `android.kotlinOptions`.',
```
jesswrd:
```suggestion
          'Legacy Kotlin configuration detected in `$buildGradlePath`.',
          'This plugin applies the Kotlin Gradle Plugin (KGP) or uses the `android.kotlinOptions{}` block.',
```

It is more accurate to replace "DSL" with "configuration" here because "DSL" refers to the scripting language so `build.gradle.kts` is in Kotlin DSL and `build.gradle` is in Groovy DSL.

jesswrd:
> cc @jesswrd - does this work for you?

Yes LGTM. Just left a minor comment.

sigurdm:
> > cc @jesswrd - does this work for you?
> 
> Yes LGTM. Just left a minor comment.

Hmm, cannot see that comment. Did you forget to publish it?

jesswrd:
> Hmm, cannot see that comment. Did you forget to publish it?

Yeah it was not published. Can you see it now?

jesswrd (approved):

lib/src/tag/tagger.dart
sigurdm: Done

PR #1593, opened 2026-06-22
Scope static analysis to lib/, bin/ and pubspec.yaml

xvrh:
pana runs `dart analyze .`, which also scans test/. Dev dependencies are stripped before analysis, so the test files can't resolve their imports and thus report errors. A package with a large test suite produces enough of these errors to exceed the analyzer's output limit, and that fails static analysis completely. Limiting analysis to those paths avoids the problem.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Detect legacy Kotlin Gradle Plugin application and `kotlinOptions` in Android plugins.
 - Upgraded dependencies to the latest.
+- Run static analysis on `lib/` and `bin/` instead of the whole package, so a large `test/` 
+  directory can no longer exceed the analyzer output limit and fail analysis.
 
 ## 0.23.12
 
--- a/lib/src/package_context.dart
+++ b/lib/src/package_context.dart
@@ -32,7 +32,7 @@ import 'tag/pana_tags.dart';
 import 'tag/tagger.dart';
 import 'tool/run_constrained.dart';
 import 'tool/webp_tool.dart';
-import 'utils.dart' show isAnalysisTarget, listFiles;
+import 'utils.dart' show analysisTargetPaths, isAnalysisTarget, listFiles;
 
 /// Shared (intermediate) results between different packages or versions.
 /// External systems that may be independent of the archive content may be
@@ -324,7 +324,7 @@ class PackageContext {
   }) async {
     final output = await toolEnvironment.runAnalyzer(
       packageDir,
-      '.',
+      analysisTargetPaths(packageDir),
       usesFlutter,
       inspectOptions: options,
     );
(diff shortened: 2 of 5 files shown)


sigurdm (approved):
Surprised we didn't already do this.

@isoos are you happy?

isoos:
We have changed this in https://github.com/dart-lang/pana/pull/1533/changes

Before that we had only `bin/` and `lib/`, but we wanted to also include further checks. I believe we should filter out test errors though, isn't it the case? Is there any specific example which shows problem with the analysis - @xvrh ?

xvrh:
@isoos when I added many generated tests in `package:aws_client` the pub.dev analysis failed (and I got 70/160 points). 
Turned out, they were so many errors reported by the analyzer (because package:test was missing) that it exceeded the 10MB limit from here: https://github.com/dart-lang/pana/blob/master/lib/src/tool/run_constrained.dart#L15

I fixed it simply by excluding the tests with a `.pubignore`. But the experience was so frustrating that I also opened this PR just in case.

isoos:
> Turned out, they were so many errors reported by the analyzer (because package:test was missing) that it exceeded the 10MB limit from here

Ah, okay, makes sense. Let me play with it a little, maybe revisiting the otherwise arbitrary 10 MB limit could be a better path here (or making an alternative method that streams the lines so they can be filtered efficiently). If possible, I'd rather keep the analysis to be more broad but use better filtering.

isoos:
Interestingly, the new pana (released a few days after the latest aws_client) is now passing both locally and on pub.dev. I'm wondering what have changed. Tbh. I did not expect that it will pass flawlessly (given its size).

xvrh:
@isoos It's probably because I've added a `.pubignore` to ignore all tests: https://github.com/agilord/aws_client/blob/master/aws_client/.pubignore

sigurdm:
I think we decided on https://github.com/dart-lang/pana/pull/1611 instead of this

PR #1594, opened 2026-06-25
Fix conditional export configuration keys

leehack:
## Summary
- Preserve conditional import/export configuration names with token lexemes instead of token values.
- Add a WASM tag regression for `dart.library.js_interop` conditional exports that select a web-safe implementation over an IO fallback.

## Root cause
With analyzer 13, `Token.value()` canonicalizes the `library` token in `dart.library.js_interop` to `LIBRARY`, so pana looks up `dart.LIBRARY.js_interop` instead of the declared runtime key `dart.library.js_interop`. That makes the conditional export resolver miss the web branch and follow the default IO export path.

Using `Token.lexeme` preserves the source spelling and matches the runtime declared variables.

## Validation
- `dart format --output=none --set-exit-if-changed .`
- `dart analyze --fatal-infos`
- `dart test test/tag/tag_end2end_test.dart --name "dart.library.js_interop conditional export"`
- `dart test --platform vm -j 1 test/tag/tag_end2end_test.dart`
- `dart test --platform vm test/tag/tag_detection_test.dart`
- `dart test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.23.14
+
+- Fix: use `lexeme` in `package:analyzer` to properly detect conditional imports.
+
 ## 0.23.13
 
 - Detect legacy Kotlin Gradle Plugin application and `kotlinOptions` in Android plugins.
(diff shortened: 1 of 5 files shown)


isoos:
@leehack: is there any specific package where this causes an analysis issue?

isoos:
This seems to fix the issue reported at https://github.com/dart-lang/pana/issues/1595

sigurdm (approved):

leehack:
> @leehack: is there any specific package where this causes an analysis issue?

Sorry for the late response. I noticed the issue when I saw the score drop for `llamadart` and `mcp_dart`. 

It seems it has already been merged. Thanks!

PR #1596, opened 2026-06-30
Stream processing of 'dart analyze' stdout for better and more resilient parsing.

isoos:
Alternative to #1593.


sigurdm (approved):

sigurdm:
We should probably still do this, even if we limit the directories we process.

PR #1605, opened 2026-08-04
Protect against YAML alias expansion bombs in yamlToJson

sigurdm:
Replaces `json.decode(json.encode(yamlMap))` with a bounded recursive converter in `yamlToJson`.

The converter guards against entity expansion bombs (Billion Laughs), cyclic references, and extreme nesting depth by enforcing `maxNodes` (default 10,000) and `maxDepth` (default 64) limits while sorting map keys directly without intermediate JSON stringification.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 import 'dart:collection';
-import 'dart:convert';
 import 'dart:io' hide BytesBuilder;
 
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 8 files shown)


lib/src/utils.dart
```diff
@@ -87,9 +103,115 @@ Map<String, Object?>? yamlToJson(String? yamlContent) {
...
-  return sortedJson(json.decode(json.encode(yamlMap))) as Map<String, Object?>;
+  final converter = _YamlConverter(maxDepth: maxDepth, maxNodes: maxNodes);
+  final converted = converter.convert(yamlMap, 0);
+  if (converted is! Map<String, Object?>) {
+    return null;
+  }
+  return converted;
+}
+
+final class _YamlConverter {
```
isoos, line 114: I'm wondering if this should be part of `package:yaml`? some kind of API that filters/sanitizes it...

isoos (approved):

lib/src/utils.dart
sigurdm: I think we can make something for package yaml to serialize in a bounded way, but let's land this here first.

PR #1612, opened 2026-08-05
Parallelize end-to-end golden tests

sigurdm:
Speed up running `test/end2end_test.dart` by parallelizing package inspections across a shared `TestEnv` and bounding concurrency with a pool sized by `Platform.numberOfProcessors`.

### Changes
- Shares a single `TestEnv` (temporary directory, local pub proxy, tool environment, and shared pub cache) across all package inspections in `setUpAll`.
- Runs package inspections concurrently using a resource pool bounded by `Platform.numberOfProcessors.clamp(1, _packages.length)`.
- Isolates errors per package so assertion and golden failures map directly to their respective `test()` case.
- Reduces execution time from ~138s down to ~15s on multi-core workstations (~9x speedup).

### DIFF
--- a/lib/src/sdk_env.dart
+++ b/lib/src/sdk_env.dart
@@ -559,7 +559,10 @@ class ToolEnvironment {
       return await _sandboxRunner.runSandboxed(
         [...command, 'global', 'run', 'dartdoc', ...args],
         workingDirectory: packageDir,
-        environment: _dartSdk.environment,
+        environment: {
+          ..._dartSdk.environment,
+          'PUB_HOSTED_URL': 'https://pub.dev',
+        },
         timeout: timeout,
         outputFolder: outputDir,
         writableConfigHome: true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.23.17-wip
 
+- Parallelize end-to-end golden tests with a shared test environment and concurrency pool.
 - Delete symlinks before writing `dartdoc_options.yaml` and `analysis_options.yaml` to prevent out-of-tree file writes.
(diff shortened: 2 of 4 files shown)


lib/src/sdk_env.dart
```diff
@@ -559,7 +559,10 @@ class ToolEnvironment {
       return await _sandboxRunner.runSandboxed(
         [...command, 'global', 'run', 'dartdoc', ...args],
         workingDirectory: packageDir,
-        environment: _dartSdk.environment,
+        environment: {
+          ..._dartSdk.environment,
+          'PUB_HOSTED_URL': 'https://pub.dev',
```
isoos, line 564: Not sure: why is this needed now?

sigurdm: `dart pub global activate dartdoc` explicitly sets `PUB_HOSTED_URL: 'https://pub.dev'` during activation (`lib/src/sdk_env.dart:550`).
When `dart pub global run dartdoc` (`lib/src/sdk_env.dart:560`) ran with `_dartSdk.environment`, `PUB_HOSTED_URL` was set to the test proxy (`http://localhost:<port>`). In sequential tests with an isolated pub cache per test, `pub global run` executed the precompiled snapshot directly without checking dependencies. But when multiple `pub` commands run concurrently against a shared pub cache, `pub global run` validates its lockfile against the active `PUB_HOSTED_URL`. Because the test proxy filters packages by a cutoff date, re-resolving `dartdoc`'s dependencies against the proxy fails (`analyzer 13.x` is missing from the proxy).
Setting `PUB_HOSTED_URL: 'https://pub.dev'` on `global run` ensures running `dartdoc` uses the same pub host it was activated from.

isoos (approved):

PR #1615, opened 2026-08-06
Keep sdk dev_dependencies when updating pubspec.yaml

isoos:


lib/src/sdk_env.dart
```diff
@@ -680,15 +680,43 @@ class ToolEnvironment {
...
+        // do not keep path or git dependencies
         if (value is Map &&
             (value.containsKey('path') || value.containsKey('git'))) {
           continue;
         }
-        keptDevDependencies[name] = value;
+
+        // do not keep hosted dependencies outside of pub.dev
+        if (value is Map &&
+            value.containsKey('url') &&
```
sigurdm: Are we sure this is right? I think a hosted dependency has one of the forms:

```
dependencies:
  foo:
  bar:
    hosted: https://my-repo.com
  baz:
    hosted:
      url: https://my-repo.com
      name: baz
```

see: https://dart.dev/tools/pub/dependencies#hosted-packages

lib/src/sdk_env.dart
```diff
@@ -680,15 +680,43 @@ class ToolEnvironment {
...
+        // keep SDK dependencies
+        if (value is Map && value.containsKey('sdk')) {
+          // sanity check on keys
+          final keys = value.keys.toSet();
+          if (keys.length != 2 || !keys.containsAll(['name', 'sdk'])) {
+            continue;
+          }
+          // sanity check that sdk is filled
+          final sdkValue = value['sdk'];
+          if (sdkValue is! String || sdkValue.trim().isEmpty) {
```
sigurdm: I don't really think we need this. The solver will complain.
We could restrict the sdk to be one of `dart`, `flutter`, but again, I think the pub solver will do that for us - in the future more sdks might exist.

lib/src/sdk_env.dart
isoos: good catch, updated, PTAL

lib/src/sdk_env.dart
```diff
@@ -680,15 +680,50 @@ class ToolEnvironment {
...
+
+          // hosted outside of pub.dev, do not keep
+          continue;
+        }
+
+        // keep SDK dependencies
+        if (value is Map && value.containsKey('sdk')) {
+          // sanity check on keys
+          final keys = value.keys.toSet();
+          if (keys.length != 2 || !keys.containsAll(['name', 'sdk'])) {
```
sigurdm: Will an sdk dependecy have a name? usually I think they look like:

```
dev_dependencies:
  flutter_test:
    sdk: flutter
```

on the other hand I think they in theory might have a version constraint (though it currently doesn't make sense)

isoos: yeah, true, assuming the solver will complain on conflicting items, I will just remove this check too.

sigurdm (approved):

PR #1616, opened 2026-08-06
Passthrough 'analyzer: exclude:' in 'analysis_options.yaml'

isoos:


lib/src/analysis_options.dart
```diff
@@ -133,6 +134,24 @@ String updatePassthroughOptions({
       }
     }
 
+    if (origAnalyzer case {'exclude': List origExclude}) {
+      for (var pattern in origExclude) {
+        if (pattern is! String) continue;
+        // sanity check to prevent too wide excludes
+        final glob = Glob(pattern);
+        if (glob.matches(
```
sigurdm: I find this too arbitrary. How would we document this? Rather I think we should just pass everything throught

sigurdm (approved):

PR #1619, opened 2026-08-06
Explicit CLI option for specifying the project root (git detection not used).

isoos:


bin/pana.dart
```diff
@@ -110,6 +110,12 @@ final _parser = ArgParser()
     help: 'The binary to use for sandboxed subprocess executions.',
     hide: true,
   )
+  ..addOption(
+    'workspace-root',
+    help:
+        'The root directory path of the workspace. '
+        'When specified, pana copies the entire tree for analysis.',
```
sigurdm: This is a bit hard to understand.
What is a workspace here? pub workspace?
Maybe also hint when this option should be used.

isoos: Renamed and updated to make it more explicit, that this is not necessarily a pub workspace, and specified the use case. PTAL.

test/goldens/help.txt
```diff
@@ -12,4 +12,5 @@ Options:
       --[no-]dartdoc           Run dartdoc and score the package on documentation coverage.
                                (defaults to on)
       --dartdoc-version        The dartdoc version to use: `sdk`, `latest` (default) or `<version constraint>`.
-      --license-data           The path to the directory containing the SPDX license data.
\ No newline at end of file
+      --license-data           The path to the directory containing the SPDX license data.
+      --project-root           The root directory path of the project (pub workspace or otherwise). Use this when the package is located in a subdirectory so pana can copy and analyze the entire project tree.
```
sigurdm, line 16: Perhaps state the default also

sigurdm (approved):

test/goldens/help.txt
isoos: I find it hard to do it in a compact way. Let's keep it as-is, and if people find it confusing or lacking, we can update it.

PR #1620, opened 2026-08-11
fix(tagger): check web plugin implementations for is:wasm-ready

kevmoo:
This fixes false positive `is:wasm-ready` tags for Flutter web plugins that import `dart:html` or `package:js` in their web implementations (e.g. `ua_client_hints`) by explicitly evaluating web plugin `fileName` definitions during web runtime verification.

### DIFF
--- a/lib/src/tag/tagger.dart
+++ b/lib/src/tag/tagger.dart
@@ -358,6 +358,9 @@ class Tagger {
     return TaggingResult(innerTags, innerExplanations);
   }
 
+  /// Returns the declared web plugin platform implementation library URIs
+  /// from `pubspec.yaml` (under `flutter.plugin.platforms.web.fileName`,
+  /// defaulting to `<package_name>_web.dart`), if the file exists.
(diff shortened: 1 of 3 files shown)


lib/src/tag/tagger.dart
```diff
@@ -352,6 +358,32 @@ class Tagger {
     return TaggingResult(innerTags, innerExplanations);
   }
 
+  List<Uri> _getWebPluginLibraries() {
```
kevmoo, line 364: Document this function. explain what it does

kevmoo: Added doc comment explaining web plugin platform entry point resolution and fallback logic, and addressed the omit_local_variable_types lint.

isoos (approved):

kevmoo:
Hey! Whatcha doing to my PR?!?

isoos:
> Hey! Whatcha doing to my PR?!?

rebased, removed extra commit because git is crap sometimes, extended changelog and rebased again... sorry for the noise

PR #1622, opened 2026-08-11
Allow more pass-through in analysis_options.yaml: analyzer error and linter ignore overrides.

isoos:


lib/src/analysis_options.dart
sigurdm: This needs an update

test/analysis_options_test.dart
```diff
@@ -62,63 +62,6 @@ formatter:
     });
   });
 
-  test('passthrough analysis errors except one', () {
```
sigurdm, line 65: Instead of deleting these tests, could we update them to confirm that you can now override the options?

sigurdm (approved):
Could the PR title be made more specifc?

lib/src/analysis_options.dart
isoos: Updated.

test/analysis_options_test.dart
isoos: Updated.

jakemac53:
Why do we want to allow arbitrary diagnostic overrides? I think that packages should have their score dinged for triggering these kinds of diagnostics, with the exception possibly of specific ones that we think are legitimate.

But for instance I think that using deprecated APIs should affect your pub score, and you should not be able to hide that from your users via your analysis options.

jakemac53:
As I understand it, this would actually allow you to hide any kind of error you want. So now when you see you are being dinged for some analysis error on pub, a valid way to resolve that is to just silence the errors? That doesn't seem right at all to me.

PR #1621, opened 2026-08-11
Updated dev_dependency handling: keep for analysis, remove for upgradability.

isoos:
Note: I've added a fallback logic for `pub downgrade` - it tries regular downgrade, then it tries it without dev-dependencies, to prevent a regression which was revealed by the end2end tests.

PR #1624, opened 2026-08-13
Fix dev-dependency resolve before running analysis.

isoos:
In #1621 we've updated the `dev_dependencies` resolution logic to remove them for `pub outdated` but keep them for other analysis. However, right after the `pub outdated` command, we never run a separate resolve with the `dev_dependencies` re-added.

The fix will try to re-add them, and if that fails, strips them again, so analysis may continue in cases where dev-dependencies are causing conflicts.


sigurdm:
I think the overall intent—trying to keep SDK/hosted `dev_dependencies` first and falling back to stripping all `dev_dependencies` if resolution fails makes sense.

Some issues:
 
* `runPub` doesn't throw on failure
In `lib/src/package_context.dart`:

```dart
      try {
        await toolEnvironment.runPub(
          packageDir,
          usesFlutter: usesFlutter,
          command: 'upgrade', // Should this also be 'get' instead of 'upgrade'?
        );
      } catch (_) {
        ...
      }
```

By default, `ToolEnvironment.runPub` sets `throwOnError: false` and returns a `PanaProcessResult` where `wasError == true`. When dependency resolution fails, it won't throw an exception, so we'll never enter the `catch (_)` block to trigger the fallback.
We should either pass `throwOnError: true` or inspect `pr.wasError` explicitly:

```dart
      var pr = await toolEnvironment.runPub(
        packageDir,
        usesFlutter: usesFlutter,
        command: 'get',
      );
      if (pr.wasError) {
        pr = await toolEnvironment.runPub(
          packageDir,
          usesFlutter: usesFlutter,
          command: 'get',
          stripAllDevDependencies: true,
        );
      }
```

* `.dart_tool` is deleted when `stripAllDevDependencies: true`

In `ToolEnvironment._withStripAndAugmentPubspecYaml` (`lib/src/sdk_env.dart`), the cleanup logic in the `finally` block unconditionally deletes `.dart_tool` and `pubspec.lock` whenever `stripAllDevDependencies == true`:

```dart
      if (stripAllDevDependencies) {
        if (await dartToolDir.exists()) {
          await dartToolDir.delete(recursive: true);
        }
        if (await lockFile.exists()) {
          await lockFile.delete();
        }
        ...
      }
```

Because `stripAllDevDependencies` was originally built for `pub outdated` (where we didn't want side effects left on disk), calling `runPub(..., stripAllDevDependencies: true)` here deletes `.dart_tool/package_config.json` right as the method finishes, leaving the package unresolved for subsequent `dart analyze` steps.
We should decouple the deletion of `.dart_tool` from `stripAllDevDependencies` (e.g. by adding a `preserveDartToolAndLock: true` parameter or moving the deletion specifically into `runPubOutdated`).

* Test coverage

Could we also add a unit test in `test/report/static_analysis_test.dart` for a package with an unresolvable hosted/SDK `dev_dependency` to verify that the fallback kicks in and allows `staticAnalysis` to succeed?

isoos:
I think #1625 has made this one obsolete.

PR #1625, opened 2026-08-19
Include hosted and SDK dev_dependencies during static analysis

sigurdm:
### Summary

When analyzing packages, standard hosted and SDK `dev_dependencies` should be available for static analysis and documentation generation (for example, to resolve `@docImport`s referencing dev-dependencies such as in `cupertino_ui`, and to load package-configured lint rules).

At the same time, `pub outdated` must only consider runtime/consumer-facing dependencies and ignore all `dev_dependencies`.

### Changes

1. **`_stripAndAugmentPubspecYaml`**:
   - Strips non-hosted (`path:`, `git:`, or custom `hosted:`) `dev_dependencies` while keeping standard hosted and SDK `dev_dependencies`.
   - When `stripAllDevDependencies: true` is passed, strips all `dev_dependencies`.
2. **`PackageContext`**:
   - In `staticAnalysis`, runs `toolEnvironment.runPub(packageDir, command: 'get')` to ensure hosted/SDK dev-dependencies are installed and resolved in `.dart_tool/package_config.json` before running the analyzer and dartdoc.
   - In `downgradeAnalysisErrorMessage`, attempts `pub downgrade` with standard dev-dependencies first, falling back to stripping dev-dependencies if downgrade fails (e.g. when legacy dev-dependencies have lower bounds incompatible with the current Dart SDK, such as non-null-safe constraints).
3. **`runPubOutdated`**:
   - Runs with `stripAllDevDependencies: true` so that `pub outdated` calculates dependency bounds for consumer-facing dependencies only.
4. **End-to-End Goldens**:
   - Updated golden test expectations (`audio_service` and `gg`) to reflect the linter rules loaded from their dev-dependencies during static analysis.

### Tests

- Unit tests in `test/sdk_env_test.dart` for `_stripAndAugmentPubspecYaml`.
- Integration tests in `test/report/static_analysis_test.dart` and `test/report/dependencies_test.dart` verifying that hosted dev-dependencies are present for analysis but ignored by `pub outdated`.


jonasfj (approved):
> In downgradeAnalysisErrorMessage, attempts pub downgrade with standard dev-dependencies first, and retries with all dev-dependencies stripped if the initial downgrade fails.

This makes no sense.

If a resolution is possible a downgrade cannot fail.

But should downgradeAnalysisError only look at `lib/`? and should it not avoid all lints? Only look for compile errors? Maybe even ignore warnings? Wouldn't that be better?

So we get to downgrade as much as possible, and see if there are static errors?

I guess the problem is that docImports and other things that may reference dev_dependencies could still fail, but my suggestion would be that we then only consider static compile errors as issues, not warnings. etc.

But I can also be okay with this behavior, it makes downgrade analysis less strict, but there is no harm in that.

sigurdm:
Thanks @jonasfj!

1. **Why `pub downgrade` can fail when `pub get` succeeds:**
   `pub get` resolves to the latest compatible versions, whereas `pub downgrade` resolves to constraint lower bounds. If a `dev_dependency`'s lower bound has transitive conflicts with the lower bounds of regular `dependencies`, `pub downgrade` will fail solving even if `pub get` worked. Since users of the package don't inherit `dev_dependencies`, falling back to stripping them allows us to still verify if runtime dependencies downgrade properly.

2. **Errors only / `lib/` scoping:**
   Downgrade analysis already works exactly as you suggested:
   * It only looks at compile errors: `problems.where((e) => e.isError)` (warnings and lints are ignored).
   * It only looks at `lib/` and `bin/`: all issues are filtered through `isAnalysisTarget`.
   * Keeping dev-dependencies in the first downgrade attempt avoids false-positive compile errors from `@docImport`s in `lib/` that reference dev-dependencies.

sigurdm:
Good point! You're completely right — because PubGrub is a complete backtracking solver over the same constraint space, if a solution exists with dev-dependencies included, `pub downgrade` will explore and find a valid resolution without needing any manual retry.

I've cleaned up `downgradeAnalysisErrorMessage` to remove the redundant retry logic.

sigurdm:
Looking at CI test results on existing packages (e.g. `dnd 2.0.1`), there is a concrete real-world case where `pub downgrade` fails even though `pub get` succeeds:

`dnd 2.0.1` has `dev_dependencies: { build_web_compilers: '>=0.2.0 <3.0.0' }` and no direct runtime dependencies.
* During `pub get` (upgrade), Pub resolves to the latest matching `build_web_compilers 2.16.x` (null-safe), so solving succeeds on Dart 3.
* During `pub downgrade`, Pub tries to select older versions towards `0.2.0`. Because early versions of `build_web_compilers` do not support null safety and Dart 3 requires null safety for all selected packages, Pub fails version solving with:
  `Because dnd depends on build_web_compilers >=0.2.0 <3.0.0 which doesn't support null safety, version solving failed.`

Because `dnd` has no runtime dependencies (or its runtime dependencies downgrade fine), falling back to stripping dev-dependencies allows `pub downgrade` to evaluate the package's actual runtime dependencies without failing due to legacy dev-dependency lower bounds.

